Issue #79, opened 2017-11-22
Add methods to change parameters of mcmc method


MichaelClerx:
And use them in an example

PR #81, opened 2017-11-22
Changeable mcmc parameters Issue #79

chonlei:
Re #79


pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
...
+
+    def set_iterations(self, iterations):
+        """
+        Set the total number of iterations.
+        """
+        self._iterations = iterations
+        self._adaptation = int(self._iterations * self._adaptation_rate)
+        self._burn_in = int(self._iterations * self._burn_in_rate)
+        if self._thinning > self._adaptation:
+            print('WARNING: Thinning larger than adaptation')
```
MichaelClerx: This is an error, if this happens a ValueError() should be raised.
The message inside the error should not use variable names
Not sure this is the right place for this check either. What if you want to change both? That shouldn't give us a warning or error!

pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
...
+            raise Exception('Thinning larger than total number of iterations')
+
+    def set_adaptation_rate(self, rate):
+        """
+        Set the first fraction of the total number of iterations before 
+        using adaptation.
+        """
+        self._adaptation_rate = rate
+        self._adaptation = int(self._iterations * self._adaptation_rate)
+        if self._adaptation > self._burn_in:
```
MichaelClerx: See above

pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
...
+        """
+        self._iterations = iterations
+        self._adaptation = int(self._iterations * self._adaptation_rate)
+        self._burn_in = int(self._iterations * self._burn_in_rate)
+        if self._thinning > self._adaptation:
+            print('WARNING: Thinning larger than adaptation')
+        if self._thinning > self._burn_in:
+            print('WARNING: Thinning larger than burn-in')
+        if self._thinning > self._iterations:
+            raise Exception('Thinning larger than total number of iterations')
```
MichaelClerx: A bare `Exception` is almost never appropriate. You can look up good error types here: https://docs.python.org/2/library/exceptions.html. In this case, a `ValueError` would be appropriate (but see above!)

pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
...
+
+    def set_adaptation_rate(self, rate):
+        """
+        Set the first fraction of the total number of iterations before
```
MichaelClerx: the first fraction?

pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
...
+        """
+        self._thinning = thinning
+        if self._thinning > self._adaptation:
+            print('WARNING: Thinning larger than adaptation')
+        if self._thinning > self._burn_in:
+            print('WARNING: Thinning larger than burn-in')
+        if self._thinning > self._iterations:
+            raise Exception('Thinning larger than total number of iterations')
+
+    def print_setup(self):
```
MichaelClerx: This is something the user can do her/himself!

Remember if we add these methods, and people use them, then we won't be able to remove them anymore without breaking their code, and then we're stuck with a whole bunch of methods that were convenient for 1 specific project for the rest of Pints' lifetime!

pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
...
-        burn_in = int(iterations * 0.5)
+        burn_in = self._burn_in
 
         # Thinning: Store only one sample per X
-        #TODO Allow changing before run() with method call
-        thinning = 4
+        thinning = self._thinning
+
+        # Print out what we are using once
+        self.print_setup()
```
MichaelClerx: This is ok, but maybe just copy-paste the code you have for print_setup into this method.

MichaelClerx: This is also the place where you should check the argument. Raise a ValueError if they're wrong.

pints/_mcmc/_adaptive.py
```diff
@@ -117,6 +280,13 @@ def run(self):
         # Return generated chain
         return chain
 
+    def chain(self):
```
MichaelClerx: I don't see the added value of this method! It will also hold the chain in memory in a way the user can't do anything about, best remove this!

pints/_mcmc/_adaptive.py
```diff
@@ -26,27 +28,190 @@ class AdaptiveCovarianceMCMC(pints.MCMC):
     Johnstone, Chang, Bardenet, de Boer, Gavaghan, Pathmanathan, Clayton,
     Mirams (2015) Journal of Molecular and Cellular Cardiology
     """
+    def __init__(self, log_likelihood, x0, sigma0=None, verbose=True):
+        super(AdaptiveCovarianceMCMC, self).__init__(log_likelihood, x0, sigma0, verbose)
+        # Set default parameters
+        self.set_default()
```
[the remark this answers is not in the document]
chonlei: But good to have an option to "reset" default if they were using an interactive one, no?

chonlei:
@MichaelClerx Changed suggested. Merge it if you think it is okay now.

Issue #66, opened 2017-11-20
Add examples/recipes of plots for posterior model-predicted time series


chonlei:
Plots so far only concern about the parameters, but it would be good to compare the posterior model-predicted time series with the observed data. 

E.g. add posterior mean (80% prediction interval) model-predicted time series with the 10th and 90th percentiles of the prediction intervals, as well as dots as observed data (i.e. pints.SignleSeriesProblem( . ).values() ).

chonlei:
And/or use each sampled parameters (from the MCMC chain) from the posterior to compare the posterior model-predicted time series with the observed data.

ben18785:
I wholeheartedly agree with this sentiment -- this is a must for model validation in PINTS.

chonlei:
Working on it, see: https://github.com/pints-team/pints/blob/mcmc-plot-examples/examples/plot-mcmc-fitting-result.ipynb

chonlei:
Done in #88

Issue #63, opened 2017-11-20
Set up general framework for plotting


MichaelClerx:
Requirements:

- Cleanly separated from the optimisation / inference code
- No graphical packages should be imported unless explicitly specified. So that everything can run without an X server (e.g. on servers), and on different platforms. In addition, loading graphical packages can be a state-changing operation (for example, you can't set the matplotlib backend if someone else has already touched matplotlib in the same process)
- Minimal dependencies (e.g. seaborn etc. only if they provide some killer feature)

MichaelClerx:
From Slack:

> **Michael:** Having a good discussion with martinjrobins and chon here about plotting
> The main question is: Should we have plotting in Pints?
> One goal of Pints (as chon) mentions, was to re-use and share code throughout the group. So from this point of view it makes a lot of sense
> **Martin:** My opinion is that there should be minimal or no plotting functionality, since we are not a plotting library.
> **Michael:** And since it creates a bunch of issues, I'm coming round to martin's side!
> *But* we could definitely add a bunch of examples that show all the different plots, so that people can copy/paste and then customise
> (Because you always want to customise right?)
> **Gary:** Can’t we just plot with simple matplotlib scripts? A handful of them for common output like Ross’ nice MCMC matrix plot, makes a lot of sense?
> **Michael:** Yeah absolutely, but it's hard to stick them in a library without (1) adding dependencies or (2) making accidental errors (causing Pints not to run on headless systems / servers etc.)
> **Gary:** a folder of example plot scripts that are updated to work with output from various pints things makes sense to me. Don’t need to be ‘in’ the library.
> (Jonathan, Martin, and Michael agree)
> **Martin:** yea, agree with "example plot scripts" you can cut and paste

MichaelClerx:
Unless anyone has any examples this ticket has then become:

- [ ] Finish #67
- [ ] Add some example plot scripts

MichaelClerx:
@ben18785 @martinjrobins @mirams @sanmitraghosh Have a look at the four examples @chonlei placed in #88 

Is that the best way to do an autocorrelation plot?
And these look quite bulky! Are we sure we shouldn't add them as convenience functions?

MichaelClerx:
I'm still undecided

martinjrobins:
[...]

PR #88, opened 2017-11-28
MCMC plot examples

chonlei:
4 new MCMC plots with a clean history.

see e.g. #63 #66


MichaelClerx:
Autocorrelation: https://github.com/pints-team/pints/blob/clean-mcmc-plot-examples/examples/plot-mcmc-autocorrelation-plot.ipynb

MichaelClerx:
Pairwise scatter: https://github.com/pints-team/pints/blob/clean-mcmc-plot-examples/examples/plot-mcmc-pairwise-scatterplots.ipynb

MichaelClerx:
Predictions: https://github.com/pints-team/pints/blob/clean-mcmc-plot-examples/examples/plot-mcmc-predicted-time-series.ipynb

MichaelClerx:
Trace plots: https://github.com/pints-team/pints/blob/clean-mcmc-plot-examples/examples/plot-mcmc-trace-plots.ipynb

ben18785:
Nice work Chon. Looking good to me. A few things:

- Pairwise plots: I think I like the KDE matrix scatterplots versions. I'm not on my computer now and so can't run them now. Do they take long to run?
- Autocorrelation: worth having, although only plot the first 10 or so lags (not 350 lags).
- Predictions: I think the figure legend needs to have the 'inferred series' line without opacity (the plot should still plot with opacity).
- Trace plots: look fine. I personally don't find trace plots that useful, although I know other people do.

@MichaelClerx An aside, but can we perhaps rename the 'BayesianLogLikelihood' function in PINTS? I think it is confusing as people don't really use this terminology -- a log likelihood is not inherently Bayesian although I know you mean this as the likelihood * prior. How about 'LogPosterior'?

@MichaelClerx Another aside, sorry. Do the optimisation examples work off of data with noise added? If not, I would suggest this is worth adding. Also, can we have an actual versus fitted at the end of these examples? Think it would make the case better than just printing the numbers...(thinking https://github.com/pints-team/pints/blob/master/examples/optimisation-xnes.ipynb, https://github.com/pints-team/pints/blob/master/examples/optimisation-snes.ipynb, https://github.com/pints-team/pints/blob/master/examples/optimisation-cmaes.ipynb).

sanmitraghosh:
I think "LogPosterior" is the correct name for the Bayesian inference

martinjrobins:
These plots look great, thanks @chonlei. I would say that some of the explanation and plot labels are a bit sparse in some of them. For example, in the pairwise scatter plot, what is a kde plot? and some plots don't have x-axis labels, or legends indicating, for example, that the green lines indicate the true value (I know it might be obvious, but its nice to be able to skim down an example and get all the information you need from the figures)

And also the print statements seem a bit out of place for what is a static notebook, I'd probably get rid of them. Even for a dynamic notebook, I guess you already get a visual indication that a cell has finished running?

MichaelClerx:
It would be great if @ben18785 and @sanmitraghosh could update the explanatory text in these notebook!

MichaelClerx:
I might split the pairwise example into two notebooks: one with and one without KDE. Thoughts?

chonlei:
Thank you all!

@ben18785 

* Pairwise plots: I think I like the KDE matrix scatterplots versions. I'm not on my computer now and so can't run them now. Do they take long to run?
   - Surprisingly fast using scipy KDE to do the plot.

@martinjrobins 

Good point! Yes, you are right, I kind of stopped quite abruptly the other day, so this has the main plots in it, but not yet "polished" for the script. I think @MichaelClerx was doing that(?), if not, I can add them in later.

---
Also, shall we add all those plots as part of the Pints and change those example by just using our new functions? 

(I am very tempted to do that - especially for the matrix plot, if I were the user, I definitely would not want to stick that "long-and-annoying two for-loops" everywhere in my script.) 

Any thoughts? I think it wouldn't be too much extra work?

ben18785:
@Chon -- is surprisingly fast still fast for 100,000 samples? Also, I agree
with you about functions -- I thought this was the intention?

chonlei:
Just did a quick test on how fast is the KDE from Scipy.

For 100,000 samples (excluded all thinning and burn-in), remains very fast for 1 variate, but it is no more surprising for multivariate plots. For a single multivariate plot: 
KDE: 38.082449913 s
histogram: 0.0163419246674 s
(Can't do much, coming from Scipy, I doubt it would be easy to beat it.) @ben18785

examples/plot-mcmc-autocorrelation.ipynb
```diff
@@ -108,7 +117,9 @@
    "source": [
     "pl.figure()\n",
     "pl.acorr(chain[:,0] - np.mean(chain[:,0]))\n",
-    "pl.xlim(0, 10)\n",
+    "pl.xlim(-0.5, 10.5)\n",
+    "pl.xlabel('Lag')\n",
+    "pl.ylabel('Autocorrelation')\n",
```
MichaelClerx: Thanks Chon!

MichaelClerx:
This is merged now (So everyone can delete their local branch).

@ben18785 If you feel like elaborating on the descriptions of the plots (the bit of text near the bottom) that's still very welcome :-)

Nice work @chon!

PR #98, opened 2017-11-29
Custom model example

### LINKED ISSUE #92: Add example of a custom model
This will be really useful!

Ideally do an ODE example (using scipy?)

MichaelClerx:
Can be seen at: https://github.com/pints-team/pints/blob/custom-model-example/examples/writing-a-model.ipynb

Fixes #92


chonlei:
I have not finished running the optimisation and terminated it. But didn't change anything there - should not be any error there!

ben18785:
@MichaelClerx Looking good. Not sure if you saw, but I wrote a few comments [here.](https://github.com/pints-team/pints/issues/92)

chonlei:
@MichaelClerx I agree, this is very important for Pints users and it looks really nice! 

I think the wrapper class section should really be stand-alone (I changed that already). Also made some slight changes.

MichaelClerx:
- I shortened it a bit!
- Replaced SNES explanation with link to the relevant pints.readthedocs.io page
- Removed the links at the bottom (I'm a bit scared to add too many hardcoded links in case we need to move/rename the project and then end up with lots of dead links everywhere!)

PR #103, opened 2017-11-30
Mcmc plot helper func

chonlei:
Put all 4 MCMC plot examples into functions as discussed in #63 

NOT customisable!


chonlei:
Also added an example of showing how to use them as "how to diagnose MCMC".

https://github.com/pints-team/pints/blob/mcmc-plot-helper-func/examples/inference-diagnose-results.ipynb

At the end emphasises if people want to do their own version, then have a look at the examples in #88 and build their own version.

chonlei:
Error message: `ImportError: No module named pylab` 

@martinjrobins  @MichaelClerx  I thought it just does not have `matplotlib.pyplot`, so I changed it to `pylab`. Is there anything we can add to make this compatible?

pints/__init__.py
```diff
@@ -80,3 +80,9 @@ def version(formatted=False):
 from _mcmc import MCMC
 from _mcmc._adaptive import AdaptiveCovarianceMCMC, adaptive_covariance_mcmc
 
+#
+# Basic plots
+#
+import _plot as plot
```
MichaelClerx: Hi @chonlei ! This makes matplotlib importing part of loading Pints. We don't want that as servers etc. might not be able to `import matplotlib`. Could you change this to make `_plot.py` into `plot.py` (so an official submodule)? Users will then have to do things like
```
import pints
import pints.plot
...
pints.plot.trace(chain)
```

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+#  For licensing information, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+import pints
+import numpy as np
+from scipy import stats
+#import matplotlib.pyplot as plt
+import pylab as plt
+
+def trace(chain):
```
MichaelClerx: Could you change the signature of this method to `trace(chain, *args)` and make it plot one or multiple chains? (The first argument is to ensure there's at least one chain)

https://stackoverflow.com/questions/3394835/args-and-kwargs#3394898

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+#
+# This file is part of PINTS.
+#  Copyright (c) 2017, University of Oxford.
+#  For licensing information, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+import pints
+import numpy as np
+from scipy import stats
+#import matplotlib.pyplot as plt
```
MichaelClerx: Please use pyplot instead of pylab!

It'd also be good not to import matplotlib in the module, but individually in each method (this doesn't have an impact on performance, the module is still only loaded once).
This way, if people import the plot module Matplotlib isn't automatically imported too (which can help prevent errors on servers etc.)

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+import pylab as plt
+
+def trace(chain):
+    """
+    Trace plot.
```
MichaelClerx: Please write a better description! Typically a one line paragraph ("This method creates and returns..."). Followed by a second paragraph with any extra description needed, and finally an argument list (like you already started) and a line at the bottom saying what exactly the method returns.

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+        plt.xlabel('Iteration')
+        plt.ylabel('Parameter ' + str(i + 1))
+        plt.plot(chain[:,i], color='green', alpha=0.5)
+
+    plt.tight_layout()
+    return fig
+
+def autocorrelation(chain):
+    """
+    Autocorrelation plot for MCMC.
```
MichaelClerx: See above

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    fig, axes = plt.subplots(n_param, 1, sharex=True, figsize=(6, 2*n_param))
+    for i in xrange(n_param):
+        axes[i].acorr(chain[:,i] - np.mean(chain[:,i]), label='p'+str(i+1))
+        axes[i].set_xlim(-0.5, 20.5)
+    axes[i].set_xlabel('Lag')
+    fig.text(0.04, 0.5, 'Autocorrelation', va='center', rotation='vertical')
+
+    return fig
+
+def prediction(chain, problem):
```
MichaelClerx: See above

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+        # Add trace subplot
+        plt.subplot(n_param, 2, 2 + 2 * i)
+        plt.xlabel('Iteration')
+        plt.ylabel('Parameter ' + str(i + 1))
+        plt.plot(chain[:,i], color='green', alpha=0.5)
+
+    plt.tight_layout()
+    return fig
+
+def autocorrelation(chain):
```
MichaelClerx: I know I said we shouldn't but... can we have an argument for the maximum lag here?

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    `chain`
+        MCMC routine returned chain
+
+    """
+    n_sample, n_param = chain.shape
+
+    fig, axes = plt.subplots(n_param, 1, sharex=True, figsize=(6, 2*n_param))
+    for i in xrange(n_param):
+        axes[i].acorr(chain[:,i] - np.mean(chain[:,i]), label='p'+str(i+1))
+        axes[i].set_xlim(-0.5, 20.5)
```
MichaelClerx: Please make max lag a param (at least internally, but maybe also as a user option, see above!)

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    axes[i].set_xlabel('Lag')
+    fig.text(0.04, 0.5, 'Autocorrelation', va='center', rotation='vertical')
```
MichaelClerx: Why not just a y label?

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+
+    return fig
+
+def prediction(chain, problem):
+    """
+    Predicted time series plot.
+
+    `chain`
```
MichaelClerx: Please stick a line `Arguments:` above this list

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+        MCMC routine returned chain
+    `problem`
+        pints.SingleSeriesProblem object
+
```
MichaelClerx: Returns...

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    if not isinstance(problem, pints.SingleSeriesProblem):
```
MichaelClerx: You don't really have to check for types in Python, but I don't mind if it makes for better error messages like here :-)

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+        raise TypeError('pints.SingleSeriesProblem type is expected')
```
MichaelClerx: Please make a little more helpful (it's 99% of the way there). "First argument to prediction() must be " or something. Error messages are the main place a library interacts with its users...

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    plt.ylabel('Value')
+    plt.plot(times, predicted_values[0], color='#1f77b4', label='inferred series')
+    for v in predicted_values[1:]:
+        plt.plot(times, v, color='#1f77b4', alpha=0.05)
+    plt.plot(times, mean_values, color='black', lw=2, label='mean inferred')
+    plt.plot(times, problem.values(), 'o', color='#7f7f7f', ms=6.5, label='data points')
+
+    return fig
+
+def forceAspect(ax, aspect=1):
```
MichaelClerx: Please stick to `lower_case_names_with_underscores` for methods and functions! (ClassesAreAllCamelCased)

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    Set aspect.
+
```
MichaelClerx: Add attribution to stackoverflow

pints/_plot.py
MichaelClerx: Consider removing aspect argument (we never use it)

MichaelClerx: Actually, I would make this a hidden method `_force_aspect` and maybe give it a better name `_force_equal_aspect_ratio` for example

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+        aspect to be set
+
+    """
+    im = ax.get_images()
+    ex = im[0].get_extent()
+    ax.set_aspect(abs((ex[1]-ex[0])/(ex[3]-ex[2]))/aspect)
+
+def hist2d(X, Y, ax):
+    """
+    2D histogram plot.
```
MichaelClerx: See previous comments

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    Xmin, Xmax = np.min(X), np.max(X)
+    Ymin, Ymax = np.min(Y), np.max(Y)
+    x2 = np.linspace(Xmin, Xmax, 25)
+    y2 = np.linspace(Ymin, Ymax, 25)
+    histplot = ax.hist2d(X, Y, bins=[x2,y2], normed=True, cmap=plt.cm.Blues)
+    forceAspect(ax)
+
+def kde2d(X, Y, ax):
+    """
+    2D kernel density estimation plot.
```
MichaelClerx: See previous comments

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    Xmax = np.max(X)
+    x1 = np.linspace(Xmin, Xmax, 100)
+    x2 = np.linspace(Xmin, Xmax, 50)
+    kernel = stats.gaussian_kde(X)
+    Z = kernel(x1)
+    ax.hist(X, bins=x2, normed=True)
+    ax.plot(x1, Z)
+    #forceAspect(ax)
+
+def pairwise_hist(chain):
```
MichaelClerx: I wouldn't use all these abbreviations in the method names! Just make it `pairwise_histogram` full out (or `pairwise_scatterplot`)

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+                    tl.set_rotation(45)
+            if j > 0:
+                # Only show y tick labels for the first column
+                axes[i,j].set_yticklabels([])
+        axes[i,0].set_ylabel('parameter %d'%(i+1))
+        axes[-1,i].set_xlabel('parameter %d'%(i+1))
+
+    return fig
+
+def pairwise_kde(chain):
```
MichaelClerx: Can we maybe turn these methods into a single method `scatterplot(chain, kde=True)` ? Then do the same for the 1d and 2d methods (in fact, decide if you want them to be exposed to the user or not, if not, then make them nested methods of this one)

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    X1, Y1 = np.mgrid[Xmin:Xmax:100j, Ymin:Ymax:100j]
+    positions = np.vstack([X1.ravel(), Y1.ravel()])
+    values = np.vstack([X, Y])
+    kernel = stats.gaussian_kde(values)
+    Z = np.reshape(kernel(positions).T, X1.shape)
+    ax.imshow(np.rot90(Z), cmap=plt.cm.Blues, extent=[Xmin, Xmax, Ymin, Ymax])
+    ax.plot(X, Y, 'k.', markersize=2)
+    forceAspect(ax)
+
+def kde1d(X, ax):
```
MichaelClerx: Can we make `ax` the first argument? And just change it to `axes`, no (non-standard) abbreviations in method and class names (unless it's a numerical method, like CMAES).

You could also consider merging all these methods into `graph(ax, x, y=None, kde=True)` that does 1d or 2d plots depending on `y is None`. Then you could make `force_aspect` a nested function of this one so it's hidden from users. Although, maybe keep 1d and 2d separate and call them `histogram` and `scatter` ?

pints/_plot.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    1D kernel density estimation plot.
+
+    `X`
+        An array of the variable
+    `ax`
+        matplotlib axes handler
```
MichaelClerx: handle. But please write full sentences here!

pints/__init__.py
```diff
@@ -80,3 +80,9 @@ def version(formatted=False):
...
+
```
MichaelClerx: Once you've changed it to `plot.py` you can also add documentation to `pints/docs/source`. Then copy paste another module's `rst` file (for example `error_measures.rst` --> `plot.rst`) and update it. Also make sure to add your new file to the main `index.rst` (where there's an alphabetically ordered list). You can test locally by running `make html` from `pints/docs` and then checking the stuff in `pints/docs/build` in your browser.

MichaelClerx:
The example notebook is very pretty, but maybe we could just use these methods in MCMC example notebooks instead of having a separate file? It would reduce the number of things people need to read (and make the MCMC examples look extra spectacular)!

pints/__init__.py
chonlei: Done

pints/_plot.py
chonlei: Do you mean having `import matplotlib.pyplot as plt` inside, e.g., `trace()`?

Something like: 
```
def trace(chain):
    """
    ...
    """
    ...
    return fig
```

And do it for each method?

MichaelClerx: Yeah!

pints/_plot.py
chonlei: Done

pints/_plot.py
chonlei: Added a few lines.

pints/_plot.py
chonlei: To avoid every subplot have an "autocorrelation" label hanging around.

pints/_plot.py
chonlei: What is see above for this one?

pints/_plot.py
MichaelClerx: Cool!

pints/_plot.py
MichaelClerx: Comment and possibly also method name. Not sure what a better name would be, but the explanation can definitely be improved!

pints/_plot.py
chonlei: I always thought `ax` was a standard abbrev. of an item of `axes` and `axes` should be _iterable_!?

MichaelClerx: Hmmmm. Is it a system of axes (plural of axis) that it represents though? Not sure I understand why `axes` would be iterable

chonlei:
Removed `matplotlib` from requirements.

pints/_plot.py
chonlei: Yes you are right it is a set of axes, so axes. But I usually do something like
```
fig, axes = plt.subplots(3, 4)
for ax in axes:
    ...
```
In this case, `axes` is a set of "axes", but they usually (the way I use it) are iterable, and I will iterate it through to get one actual plot and usually call it `ax`. 

Anyway, could be my convention, not sure, but now I call them `axes` in the method. :-)

Issue #70, opened 2017-11-20
Add notebook showing how to set stopping criteria


MichaelClerx:
Gary says:
> I'd like to see how to change tolerances within the optimisers etc. Even if the methods in the examples/tutorials leave them at defaults it is nice to be shown how to use them with comments saying this is optional.

~We should probably add this at the end of the "First example" for optimisation and MCMC (so add a bit where we use the full API, not just the simple function)~

MichaelClerx:
Currently done this for:
- [ ] CMA-ES
- [ ] PSO
- [ ] xNES
- [ ] SNES

MichaelClerx:
We should probably add a new example showing 'advanced' optimiser use (once #132 is done)

MichaelClerx:
Ideally this would be demonstrated in a way that applies to opt. as well as mcmc. Of course they don't use the same criteria, but keeping it in mind when writing the text will be good :-)

PR #134, opened 2017-12-07
Added methods to set PSO stopping criteria, including tests.

MichaelClerx:
See also #70

### DIFF
--- a/pints/__init__.py
+++ b/pints/__init__.py
@@ -37,6 +37,7 @@ from _core import ForwardModel, SingleSeriesProblem
 # Utility classes and methods
 #
 from _util import strfloat, vector
+from _util import _StdOutCapture
 
 #
 # Boundaries and prior distributions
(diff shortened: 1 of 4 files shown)


chonlei:
Not quite related to the main point of this PR! But why do we have the `_StdOutCapture` in `__init__`? 

Is that just for testing? Or you want to use it as error log as well? But probably use [`logging`](https://docs.python.org/2/library/logging.html) is better for latter case, no?

If not, is it better to just import in the tests? Something like `from pints._util import _StdOutCapture` in the tests. As you said, if we keep these things, it might slow down importing pints. Unless it is a common practice! (I don't know much about it!)

MichaelClerx:
@chonlei I think it'll be useful when wrapping around external optimisers/mcmc methods as well, that's why I stuck it somewhere more public. I could change it to something not imported by default I suppose... 

About loading times: Importing all of matplotlib can affect performance, creating a single class is hardly any work at all!

MichaelClerx:
I definitely wouldn't use it to create an error log, btw. It's just a tool to capture output generated by some code you run (but that you don't have control over)

Issue #42, opened 2017-08-31
Add simple method that plots 1d log-likelihoods near points in parameter space


MichaelClerx:

MichaelClerx:
This would be a good diagnostic plot too. We could make it accept `ErrorMeasure` and `Likelihood` objects, and then for each parameter just do + or - 1%. Possibly something smarter like taking tiny steps and trying to get a certain percentage change in the output

MichaelClerx:
If implemented, we could also change the `inference-first-example` to use this to plot the loglikelihood (instead of the current manual approach)

mirams:
A very useful feature is this 1D line between two particular points. 

We do this a lot from the point in parameter space you end up at (p1 max posterior or max likelihood) to another particular point in parameter space (p2 e.g. the right answer that simulated synthetic data). With a bit of overshoot either side. e.g. a vector from p1-0.5(p1->p2) to p1+1.5(p1-p2). This is what Ross' plots of numerical tolerance effect are doing.

MichaelClerx:
Yeah I'm thinking anything that has a `dimension()` method and is callable

chonlei:
@mirams  I am not sure if it is that easy to combine what Micheal is suggesting and the 1D line between two particular points. So I separated them and made two methods instead (because both are useful!). The 1D line between two points is `log_likelihood_two_points(...)` and an example of it in [here](https://github.com/pints-team/pints/blob/4329f14b84d233985037dd0f690768c240d518c7/examples/optimisation-first-example.ipynb) -- I think it makes more sense to do it in the optimisation example than the inference one.

@MichaelClerx I think probably it is easier (and make more sense) to just ask the users to specify the range (boundaries) to look at! It is in the `log_likelihood_1d(...)`. 

Also, I added `histogram` to do the manual plot in the inference example -- I think having a shorter script there can make people focus on the good things about MCMC and Pints (because if they want to know how to plot, just go to the plotting examples; whereas in that inference-first-example, just focus on inference).

mirams:
Looks nice, I guess there should be an optional argument that says how many samples to take along the line?

chonlei:
Yup, I have that called `n_eva` stands for number of evaluations -- maybe call it something else (?).

MichaelClerx:
I was hoping you'd work out some sort of algorithm to plot a loglikelihood without having the user say what the limits on each parameter are! That's quite a bit of work for a high dimensional problem...

mirams:
[...]

PR #139, opened 2017-12-08
(1dLogLikelihood) Add 1d log likelihood plots

chonlei:
See #42 

Added 3 methods to module `pints.plot` and their examples to `inference-first-example.ipynb` and `optimisation-first-example.ipynb`.


MichaelClerx:
I the example, you're plotting an `ErrorMeasure`, not a log likelihood

pints/plot.py
```diff
@@ -8,6 +8,201 @@
 #
 from __future__ import division
 
+def log_likelihood_1d(log_likelihood, parameters, boundaries=None, n_eva=20, chain=None):
```
MichaelClerx: Instead of `n_eva` I'd just call this `evaluations`

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    """
+    Takes a set of parameters as input and creates and returns a plot showing 
+    the log-likelihood function of each parameter with all the other 
+    parameters held at the given set of parameters.
+    
+    Arguments:
+
+    `log_likelihood`
+        A class (or that extends from) :class:`pints.LogLikelihood` of the
```
MichaelClerx: - The word "class" means "type". The argument `log_likelihood` is an object of a certain type. So you're not passing in a class, you're passing in an object of that class.*
- The general terminology here is "A extends B", no "from" is required! (It actually changes the meaning a bit: a branch extends from a tree, but a mathematical method can extend a previous method (no from!) meaning it encompasses the previous method _and_ adds something new. The second meaning is the one used when you say one class extends another)
- It's part of the rules of classes that if something accepts X, it accepts all subclasses Y that extend X. So you don't need to spell this out here!

\* This is extra important because in Python you can actually pass in a class, for example:
```
def creator(some_class):
    return some_class()
class Test(object):
    def __str__(self):
        return 'I am a test object'
x = creator(Test)
print(x)
I am a test object
```

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+        A class (or that extends from) :class:`pints.LogLikelihood` of the 
+        problem of interest.
+    `parameters`
+        A set of parameters, which the `log_likelihood` will be evaluated at, 
+        with the same dimension as the `log_likelihood` will accept.
+    `boundaries`
+        A class (or that extends from) :class:`pints.Boundaries` which defines 
+        the boundaries of each parameters.  
+    `n_eva`
+        (Optional) The number of evaluation of each parameter.
```
MichaelClerx: evaluations (plural). But also the parameter won't be evaluated...

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    `chain`
```
MichaelClerx: Cool option!

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+        provide either `chain` or `boundaries`, but not both.
+
+    Returns a `matplotlib` figure object and axes handle.
+    """
+    import pints
+    import numpy as np
+    try:
+        import matplotlib.pyplot as plt
+    except:
+        raise ImportError("The method pints.plot.trace requires matplotlib")
```
MichaelClerx: Are we doing this in the other methods? It seems a bit redundant, since you're catching an ImportError saying matplotlib can't be imported, and then raising a new ImportError to say matplotlib can't be imported

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+
+    n_param = len(parameters)
+    # Check dimension
+    if n_param != log_likelihood.dimension():
+        raise ValueError('Input parameters dimension must match the dimension'
+                        ' of the log_likelihood accept')
```
MichaelClerx: Please fix this sentence (and add punctuation)

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    # Check input
+    if not isinstance(boundaries, pints.Boundaries) and boundaries is not None:
+        raise TypeError('The argument boundaries must be a type of or extended'
```
MichaelClerx: - Don't mention the extending bit
- It's more python-like not to have this check at all, btw, and just let the method accept anything that has similar methods to a boundaries object

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+                ' from pints.Boundaries')
+    if chain is None and boundaries is not None:
+        def_bound = False
+    elif chain is not None and boundaries is None:
+        def_bound = True
+    else:
+        raise TypeError('One of the input chain or boundaries (and only one)'
```
MichaelClerx: Please formulate a bit more clear (and punctuate). Maybe "The optional argument `chain` can not be used when `boundaries` are given." ?

MichaelClerx: or "...when the argument `boundaries` is set."

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+                        ' is expected')
+
+    fig, axes = plt.subplots(n_param, 1, figsize=(6, 2*n_param))
```
MichaelClerx: I'm worried about all the fixed plot sizes in `plot.py`. We're choosing them just because they look good in our examples

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    
+    Arguments:
+
+    `log_likelihood`
+        A class (or that extends from) :class:`pints.LogLikelihood` of the 
+        problem of interest.
+    `parameters`
+        A set of parameters, which the `log_likelihood` will be evaluated at, 
+        with the same dimension as the `log_likelihood` will accept.
+    `boundaries`
```
MichaelClerx: You need to indicate this is optional (i.e. a chain can be given instead). Also add a check that one or the other is given. (Or combine them into a single argument)

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    else:
+        raise TypeError('One of the input chain or boundaries (and only one)'
+                        ' is expected')
+
+    fig, axes = plt.subplots(n_param, 1, figsize=(6, 2*n_param))
+    for i, p in enumerate(parameters):
+        axes[i].set_xlabel('Parameter ' + str(1 + i))
+
+        # Generate some x-values near the given parameters
+        if def_bound:
```
MichaelClerx: I don't understand what `def_bound` means. Why not just write `if chain is not None:` ... `else`

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+            sigma = np.std(chain[:,i])
+            xmin = mu - 3*sigma
+            xmax = mu + 3*sigma
+        else:
+            xmin = boundaries.lower()[i]
+            xmax = boundaries.upper()[i]
+
+        x = np.linspace(xmin, xmax, n_eva)
+
+        # Calculate log-likelihood with other parameters fixed
```
MichaelClerx: http://pints.readthedocs.io/en/latest/function_evaluation.html#pints.evaluate

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+            + list(parameters[1+i:])) for j in x]
+
+        # Plot
+        axes[i].plot(x, y, c='green', label='Log-likelihood')
+        axes[i].axvline(p, c='blue', label='Input value')
+        axes[i].legend()
+
+    # Add vertical y-label to middle plot
+    #fig.text(0.04, 0.5, 'log Likelihood', va='center', rotation='vertical')
+    axes[int(i/2)].set_ylabel('log Likelihood')
```
MichaelClerx: You've set this up (and used it in an example) where the function is not a log likelihood... I can see this being useful for logposteriors etc. as well!

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+
+    plt.tight_layout()
```
MichaelClerx: Is it always ok to use this? I get errors / ugly plots some times. Not sure why!

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    return fig, axes
+
+
+def log_likelihood_two_points(log_likelihood, p1, p2, n_eva=20):
+    """
+    Creates and returns a plot showing the log-likelihood function between two
```
MichaelClerx: Always a log_likelihood?

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+    """
+    Creates and returns a plot showing the log-likelihood function between two 
+    points in the parameter space. 
+    
+    Arguments:
+
+    `log_likelihood`
+        A class (or that extends from) :class:`pints.LogLikelihood` of the 
+        problem of interest.
+    `p1`, `p2`
```
MichaelClerx: Split into separate lines! And don't say `p1` and `p2` are "a set of parameters", that implies there is a single second argument to this method that should be `set([p1, p2])`

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+        A set of parameters, which the `log_likelihood` will be evaluated at, 
+        with the same dimension as the `log_likelihood` will accept.
+    `n_eva`
```
MichaelClerx: `evaluations`

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+        (Optional) The number of evaluation along the line in parameter space.
+
+    Returns a `matplotlib` figure object and axes handle.
+    """
+    import pints
+    import numpy as np
+    try:
+        import matplotlib.pyplot as plt
+    except:
+        raise ImportError("The method pints.plot.trace requires matplotlib")
```
MichaelClerx: See previous note

pints/plot.py
```diff
@@ -8,6 +8,201 @@
...
+
+    n_param = len(p1)
+    p1 = np.array(p1)
+    p2 = np.array(p2)
+
+    # Check dimension
+    if n_param != log_likelihood.dimension() or n_param != len(p2):
+        raise ValueError('Input parameters dimension must match the dimension'
+                        ' of the log_likelihood accept')
```
MichaelClerx: Please fix

pints/plot.py
MichaelClerx: Can you split it into two checks (one check for both, one check for neither) and give the appropriate error message? Will make it a bit easier for people to debug :-)

Issue #80, opened 2017-11-22
Set up an interface for SMC via PyMC3 and Theano


chonlei:
Extended from #56  

And add an example.

PR #146, opened 2017-12-19
SMC via Pymc3

MichaelClerx:
Just making a ticket for this so we can discuss the implementation!

See #80


MichaelClerx:
Couple of issues that need to be dealt with:

- [ ] All imports of theano and pymc3 need to inside methods, so that the module can be loaded without having them installed (this is why the tests are currently failing)
- [ ] Tests should be added!
- [ ] We should create a new list of requirements (`full`, or `all`, or `extras` or something) that should be loaded by travis as well (this means changing setup.py and adding a `requirements-full.txt` or something)

In addition to that the code needs reviewing

pints/_mcmc/_smc.py
```diff
@@ -0,0 +1,266 @@
...
+import pints
+import numpy as np
+import theano
+import theano.tensor as tt
+from theano.compile.ops import as_op
+import pymc3 as pm
+# from pymc3.step_methods import smc
+
+
+# Set some flags for theano
```
MichaelClerx: Why? Which flags? Why here? What are they for?

pints/_mcmc/_smc.py
```diff
@@ -0,0 +1,266 @@
...
+THEANO_FLAGS = 'floatX=float32,optimizer=fast_compile'
+theano.config.floatX = 'float64'
+theano.config.compute_test_value = 'raise'
+theano.config.exception_verbosity = 'high'
+
+
+#TODO: issue due to @as_op cannot be done inside class
+global evaluate, times
```
MichaelClerx: Will this work for repeated runs etc.? Usually a bad idea!

pints/_mcmc/_smc.py
```diff
@@ -0,0 +1,266 @@
...
+                'Currently only support pints.BayesianLogLikelihood')
+
+        # Marginal Likelihood
+        self._marginal_likelihood = None
+
+    def _get_prior_as_pm(self, prior):
+        """
+        Return PyMC3 type prior.
+        """
+        if isinstance(prior, pints.UniformPrior):
```
MichaelClerx: Is this the only way to solve this? Would rather not test what type of prior is given! (or even _if_ it's a prior that's given, as long as it works it works (duck typing!))

pints/_mcmc/_smc.py
```diff
@@ -0,0 +1,266 @@
...
+                sd=cov[-1, -1])
+            return prior_pm
+        elif isinstance(prior, pints.Prior):
+            raise NotImplementedError('Provided Prior not yet implemented.')
+        else:
+            raise NotImplementedError('pints.Prior object is expected.')
+
+    def tt_evaluate(self, x):
+        """
+        Return Theano tensor object.
```
MichaelClerx: That's not a helpful comment!

pints/_mcmc/_smc.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        return self._tt_evaluate(x)
+
+    def run(self):
+        from tempfile import mkdtemp
```
MichaelClerx: This can be a top-level import

pints/_mcmc/_smc.py
```diff
@@ -0,0 +1,266 @@
...
+        Return the computed marginal likelihood through the SMC routine.
+
+        Return None if SMC.run() not yet run.
+        """
+        return self._marginal_likelihood
+
+
+@as_op(itypes=[tt.dvector], otypes=[tt.dmatrix])
+def _tt_evaluate_single_series(x):
+    sol = evaluate(x)
```
MichaelClerx: What's going on here? Should the evaluation be inside the try/catch? What kind of exception are we hoping to catch?

pints/_mcmc/_smc.py
chonlei: Yes! This is a really bad idea indeed. But I tried putting the `@as_op` on line 244 inside the class, it does not like it. If we find a better way of dealing the `@as_op` from `theano`, then we should definitely remove this.

pints/_mcmc/_smc.py
chonlei: Since we had the model returning `inf` if we cannot solve it with the given parameters, then I put this `try` to temporary get away from that. But it should be something else really... (We should either raise the exception as `ValueError` and ask the user (myself) to change the return to something else when a parameter is not solvable -- but what should the something else be(?). Or we explicitly say return `inf` if the user-defined model becomes not-solvable for the given parameters, and we deal with it better here.)

chonlei: Having the `evaluate(x)` inside try/catch does not help I think, because the `evaluate(.)` does not return any error, but yes, we can put it inside (it probably looks better).

MichaelClerx: I still don't understand what the try is supposed to catch

MichaelClerx:
What's the status on this @sanmitraghosh ? Should we abandon this PR?

sanmitraghosh:
Yes. But also ask chon.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

MichaelClerx:
@chonlei ?

MichaelClerx:
Closing this PR!

MichaelClerx:
Since we've closed the issue long ago, and are doing SMC ourselves, I guess we can delete this branch?

ben18785:
@MichaelClerx Happy for this to go, yep.

PR #151, opened 2017-12-26
Issue#148 - emcee hammer mcmc

ben18785:
Need to import emcee in Travis build to stop it failing.


MichaelClerx:
Have fixed the import error and added a test: At the moment this method seems very slow, is that normal or am I using it wrong? @ben18785 ?

MichaelClerx:
And are these walkers perhaps something we can parallelise?

chonlei:
I am not sure if the emcee package has an inbuilt version of parallelisation of workers or not. But when I was trying it, I used the parallelisation code that we had in Pints to make things parallelised. So, yes, should be doable.

MichaelClerx:
Thanks Chon! Ben just showed me that yes, it does have built-in parallelisation, so that's very good! We've made a ticket for it in fact :-)

chonlei:
@MichaelClerx On that note, although they have builtin parallelisation, I remember I had some problems with their parallelisation -- so I adapted the one in Pints, maybe just something to watch out later.

PR #154, opened 2017-12-27
Class for MCMC results written

ben18785:
Along with a notebook that demonstrates its use. We need to make all MCMC methods return these objects as default, rather than the samples themselves.

Also, this method should be almost applicable (without rhat and ess calcs) to the result of other samplers, so not sure whether we want to abstract this class up a level?


chonlei:
Should we not have the MCMC `run()` returns the `McmcResultObject`? (For example in PyMC3 their MCMC run returns their class object instead of an array. Though I am actually not sure if this is really the best thing to do...)

ben18785:
Hi Chon,

Yep, I was thinking that all MCMC routines would return these objects. I wanted to make sure we got the format of the object right before implementing it across all MCMC samplers, though.

Best,

Ben

pints/_mcmc/_result.py
```diff
@@ -0,0 +1,130 @@
...
+        # If there is more than 1 chain calculate rhat
+        # otherwise return NA
+        if self._num_chains > 1:
+            self._rhat = diagnostics.rhat_all_params(chains)
+        else:
+            self._rhat = np.repeat("NA", self._num_params)
+
+        self._summary_list = None
+        self.make_summary()
+
```
chonlei: Hi Ben,

Maybe we should also add function to return the chain itself (as a numpy array object)! I think it will be especially useful if later we return this class object in the MCMC `run()`.

Best,
Chon

ben18785:
Good shout Chon -- definitely useful for people wanting the raw chains.

pints/_mcmc/_result.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, chains, time):
+        self._chains = chains
+        stacked = np.vstack(chains)
+        self._time = time
+        self._mean = np.mean(stacked, axis=0)
+        self._std = np.std(stacked, axis=0)
+        self._quantiles = np.percentile(stacked, [2.5, 25, 50,
+                                                        75, 97.5], axis=0)
+        self._ess = diagnostics.effective_sample_size(stacked)
```
martinjrobins: would it make sense to move this calculation to the make_summary() method? So if the user doesn't want the ess it isn't calculated.

pints/_mcmc/_result.py
```diff
@@ -0,0 +1,157 @@
...
+        self._ess_per_second = self._ess / time
+        self._num_params = stacked.shape[1]
+        self._num_chains = len(chains)
+
+        # If there is more than 1 chain calculate rhat
+        # otherwise return NA
+        if self._num_chains > 1:
+            self._rhat = diagnostics.rhat_all_params(chains)
```
martinjrobins: Perhaps move this to make_summary() as well?

ben18785: Good points Martin and Chon! Happy to move these calculations to make_summary; agreed that this should only be done if requested by the user. I will make these changes and make MCMC routines return these objects by default.

@martinjrobins @MichaelClerx @chonlei -- how do you feel about the extra dependency that this functionality requires (it needs a Python package called "tabulate")? It's a lightweight package that does the formatting of the table when we print the object. I'm tempted for now to include it as a core dependency but, later on, someone/me can copy its action using core Python code. Just don't want to get too caught up on dependency stuff now!

MichaelClerx:
Please don't merge things before all tests pass :-)

PR #158, opened 2017-12-28
Issue#149 differential evolution mcmc

ben18785:


examples/Differential evolution.ipynb
```diff
@@ -0,0 +1,276 @@
...
+    "print('Done!')\n",
+    "\n",
+    "# Plot output\n",
+    "import pints.plot # Use Pints' diagnostic tool\n",
+    "\n",
+    "stacked = np.vstack(chains)\n",
+    "pints.plot.pairwise(stacked, kde=False)\n",
+    "\n",
+    "plt.show()"
+
```
chonlei: We probably want the examples to be more consistent. We can either split the two examples or we can also do one that use the same data and compare the performance of different MCMC algorithms?

I will split the two first.

pints/_mcmc/_differentialEvolution.py
```diff
@@ -0,0 +1,578 @@
...
+        """
+        Sets the number of chains to evolve
+        """
+        if num_chains < 10:
+            raise ValueError('This method works best with many chains (>>10,'
+                             + 'typically).')
+        self._num_chains = num_chains
+
+
+def R_draw(i, num_chains):
```
chonlei: Should we make this to a hidden method? e.g. `_R_draw` and add some comments?

pints/_mcmc/_differentialEvolution.py
```diff
@@ -0,0 +1,578 @@
...
+    Uses differential evolution adaptive Metropolis (DREAM)
+    MCMC as described in [1] to do posterior sampling
+    from the posterior.
+
+    In each step of the algorithm N chains are evolved
+    using the following steps:
+
+    1. Select proposal:
+
+    x_proposed = x[i,r] + (1 + e) * gamma(delta, d, p_g) *
```
chonlei: Can we possibly use LaTex format? Probably hard to read these equations in the documentation. (see above -- line 217)

pints/_mcmc/_differentialEvolution.py
```diff
@@ -0,0 +1,578 @@
...
+
+def differential_evolution_mcmc(log_likelihood, x0, sigma0=None):
+    """
+    Runs an differential evolution MCMC routine with the default parameters.
+    """
+    return DifferentialEvolutionMCMC(log_likelihood, x0, sigma0).run()
+
+
+class DreamMCMC(pints.MCMC):
+    """
```
chonlei: Can we do something like `r""" ... $equation$ ... """` to make things in LaTex format? @MichaelClerx probably have more ideas on this.

Issue #156, opened 2017-12-27
Add twisted Gaussian and bimodal as toy pdfs


ben18785:
Described here: https://www.researchgate.net/publication/41035177_Accelerating_Markov_Chain_Monte_Carlo_Simulation_by_Differential_Evolution_with_Self-Adaptive_Randomized_Subspce_Sampling

ben18785:
See also:  #123

Issue #123, opened 2017-12-05
Add Banana problem


MichaelClerx:
Please edit this issue until it makes sense :D

sanmitraghosh:
I am assigning this to me as well.

PR #162, opened 2017-12-29
Issue#156 toy pdfs

ben18785:
See: #156

Added a new class of object which samplers can run on which represent (often un-normalised) pdfs. Currently, there are four toy pdfs:

- Twisted Gaussian.
- Multivariate Gaussian.
- Bimodal multivariate Gaussian.
- Rosenbrock function (which, I think, is the banana function mentioned in #123).

Sorry @sanmitraghosh, I realise that the banana problem was assigned to you, but I couldn't help myself!


sanmitraghosh:
Dear Ben,

Its great that you have implemented these toy distributions. It would be great if you can add a sample() method to each of these.

Regards,
Sanmitra


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

ben18785:
Dear Sanmitra,

Apart from the multivariate normal, the best way to sample from these distributions (I fear) is MCMC - the very methods we're testing. Please let me know if you think otherwise...

An exception would be the multivariate normal...

ben18785:
Also, not sure I follow you in terms of the twisted Gaussian. As I read it from the paper, I believe it is as I have implemented it...

sanmitraghosh:
Since we know the analytic form of these distributions we can sample from them. Sampling from twisted can be done as follows:
V=100 
b=0.01
    def sample(V, b, dimension,n=1):
        X = np.random.randn(n, 2)
        X[:, 0] =np.sqrt(V) * X[:, 0]
        X[:, 1] = X[:, 1] + b* (X[:, 0] ** 2 - V)
        if dimension > 2:
            X = np.hstack((X, np.random.randn(n, dimension - 2)))

Sampling from mixture is a bit tricky as one need to sample from a categorical distribution. You have to do a few steps.

sanmitraghosh:
However, I don't think these sampling functions are that important so don't spend time on it. I would implement if I need them.

ben18785:
Still not sure I follow to be honest (knowing the analytic form doesn't
mean we can necessarily sample from them :).

If you want to implement sampling though, please feel free. As you say, I
don't think it's that crucial though.

sanmitraghosh:
By analytic I mean that we know the analytic form of the transformation. We can sample from a mvn distribution and then we just apply the transformation (bananification) to get the sample from banana. If you run my code snippet you would get samples as banana distributed. But as I said this is more useful for visualisations so not that important. By the way we run MCMC on the banana problem while pretending that we don’t  know the analytic form of the transformation. We pretend this way for benchmarking purposes only as we can evaluate exactly how good the MCMC is doing. And for this benchmarking we need to calculate empirical quantiles of the banana distribution (using the true analytical form of the banana density) using the samples from MCMC. This is how we can test whether the sampler is adequately capturing the non-linear contours. However, calculating the empirical quantile is a bit tricky and I am working on it. I will add the empirical quantile method once I get that working.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

sanmitraghosh:
Lets say we have the normal density f(x) and a sample x~f(x). Now we get a sample y from the banana distribution g(y) by transforming x as y=banana(x). Now when we are sampling with MCMC we are building a chain of y. So we need the pdf of y given by

g(y)=f(x(y))|dx/dy|,   for the banana |dx(y)/dy| =1.

Thus to evaluate g(y) we need the inverse_banana function x=inverse_banana(y).

In your code you, I believe, you are evaluating g(y)=f(y(x))|dx(y)/dy|. I think you need to use f(iverse_banana(y)) instead of f(banana(x)).

Subtracting will give you the inverse_banana.


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

pints/__init__.py
```diff
@@ -103,6 +103,11 @@ def version(formatted=False):
 #
 import pints.noise
 
+#
+# Toy pdfs
+#
+from .toy._logpdfs import LogPDF
```
chonlei: Why we want to separate `LogPDF` and `LogLikelihood`?

I think they are really the same thing when we use them. We can have all toy PDFs inherit the `LogLikelihood` class or `LogPosterior` (if we want to have it looks more "meaningful" -- as effectively, or eventually _we pretend_, they are the log posteriors, right?).

pints/_mcmc/__init__.py
```diff
@@ -32,7 +32,8 @@ def __init__(self, log_likelihood, x0, sigma0=None):
 
         # Store function
         if not isinstance(log_likelihood, pints.LogLikelihood):
-            raise ValueError('Given function must extend pints.LogLikelihood')
+            if not isinstance(log_likelihood, pints.LogPDF):
```
chonlei: Can we not have something like `if not isinstance(log_likelihood, pints.LogLikelihood) and not isinstance(log_likelihood, pints.LogPDF):` here? (If we still want to keep `LogPDF` and `LogLikelihood` as two separate things, see comment above.)

pints/toy/_logpdfs.py
```diff
@@ -0,0 +1,189 @@
...
+#  software package.
+#
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import numpy as np
+from scipy.stats import multivariate_normal
+
+
+class LogPDF(object):
```
chonlei: Maybe we don't need this and replace everything by `LogPosterior` or `LogLikelihood`? See comment above.

pints/toy/_logpdfs.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    Class for log pdfs. These are typically
+    toy problems where the target density
+    is known.
+    """
+    def __init__(self, dimension):
+        self._dimension = dimension
+
```
chonlei: If we want to keep this, maybe add `def __call__(self, x)` as `raise NotImplementedError`?

ben18785:
Hi Chon,

Whilst mathematically LogPDF are the same as LogPosteriors (currently
LogLikelihood), the way the latter are implemented means that in Pints they
are not. In particular, LogLikelihoods have attributes like time series
which the toy pdfs do not. This way I found it simpler to implement them as
a separate class. Let's discuss this at the next meeting, though.

Cheers,

Ben

MichaelClerx:
I'm glad we have these!
For the examples pages, can we have a section on examples.md called "Toy problems" and then list these under there? (We should also add an example of the logistic model there, I'll create a ticket for that!). And then one notebook per PDF, I'm thinking

pints/__init__.py
MichaelClerx: I think this is resolved now, with Ben's comments below.

However, I'd like to add that I don't see why this is imported here! All other toy stuff isn't imported as part of pints proper, which is good, because we should only use stuff from the toy module when benchmarking / demoing / testing!

pints/_mcmc/__init__.py
MichaelClerx: Should LogPDF perhaps be a main pints thing? So _not_ specific to the toy model?
In that case our LogPosterior could extend it, so that this could be a single check

chonlei: If we are moving this to a main pints thing, we might want to make this very clear -- the similarity and/or the difference between the two. Actually we should make it clear either we move it or not, because they very similar in some sense, and that might confuse users.

MichaelClerx: Yeah! This is making it more clear to me why I don't like calling the thing a LogPosterior. It's the log of a PDF, why should we care what it's meaning is in the context of a particular equation?
@ben18785 ?

pints/__init__.py
sanmitraghosh: @chonlei I just had a chat with Michael about this. let me explain.
In Bayesian inference we have:
logPosterior =LogLikelihood + LogPrior

Now for time series problems we calculate LogLikelihood, LogPrior and add them up to evaluate LogPosterior. But for some toy densities such as the Banana one we have a direct analytical form of the logPDF (not to confuse with posterior as in Bayesian sense). Thus
logPDF= f(Theta), f some function that we know in closed form. In this case the notion of likelihood and prior does not exist. MCMC is a general algorithm to sample from any unnormalised density that we can evaluate. Bayesian inference is one example problem that can be solved using MCMC, but is not limited to Bayesian only. When @ben18785  wants to sample from these toy distributions then he is not doing Bayesian inference, he is just sampling some known distributions that are hard to sample from.

MichaelClerx: Yes! Or go even further and just let it be any kind of distribution? Or function?

pints/_mcmc/__init__.py
MichaelClerx: Nevermind, see above

pints/__init__.py
chonlei: Good to keep its requirement for `dimension(self)` and `__call__(self, x)` though...

chonlei: And it makes sense to force the sampler to do sampling from an unnormalised PDF (as @sanmitraghosh said). Then it probably also make sense to have classes extend from the `logPDF (unnormalised)` as above, and indeed logPosterior, logPrior etc. belong to this category.

pints/_mcmc/__init__.py
MichaelClerx: Is LogPDF an ok name @ben18785 and @sanmitraghosh ? Or are we losing generality (could it be a pmf?)

ben18785:
Good point, although don't think we'll be dealing with discrete
distributions and so think LogPDF is fine.

pints/_mcmc/__init__.py
MichaelClerx: Ok, will have a look at this now as part of #166

MichaelClerx:
A few points about the notebook:

- [ ] Are we using enough points for the banana function? 
- [ ] Why do we use four chains everywhere? Would it not be better to just run one until it converges (since we know the base truth here)?

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+#  software package.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+from scipy.stats import multivariate_normal
+
+
+class TwistedGaussianLogPDF(pints.LogPDF):
```
MichaelClerx: Needs:

- [ ] An example in a notebook
- [ ] A test
- [ ] Sphinx documentation

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+    [1] "Accelerating Markov Chain Monte Carlo Simulation by Differential
+    Evolution with Self-Adaptive Randomized Subspace Sampling",
+    2009, Vrugt et al.,
+    International Journal of Nonlinear Sciences and Numerical Simulation.
+    """
+    def __init__(self, b=0.1):
+        self._dimension = 10
+        self._b = float(b)
+        if self._b < 0:
+            raise ValueError('b cannot be negative.')
```
MichaelClerx: Please mention in the docstring above

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __call__(self, x):
+        if len(x) != 10:
```
MichaelClerx: This check can go! The default error is fine if people call with the wrong input...

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+            raise ValueError('Dimensions must equal 10')
+        cov = np.diag(np.repeat(1, 10))
+        cov[0, 0] = 100
+        mu = np.repeat(0, 10)
+        phi = np.array([x[0], x[1] + self._b * x[0]**2 - 100 * self._b,
```
MichaelClerx: Could this not be something simpler? Like `phi = np.array(mu, copy=True)`, `phi[1] += ...`

Maybe even twist x[0] and make the dimension a parameter to `__init__`

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class MultivariateGaussianLogPDF(pints.LogPDF):
+    """
+    *Extends:* :class:`LogPDF`.
+
+    Multivariate Gaussian pdf. Default is a covariance matrix with
+    ``cor(i,j) = 0.5``, and ``var(i) = i``.
+    """
+    def __init__(self, dimension=100, mu=None, cov=None):
```
MichaelClerx: Could we use more descriptive names like `mean` and `covariance` ?

MichaelClerx: Would it not be better to remove the `dimension` parameter and infer this from the mean? (Not sure when the `mean=None` version would be useful!)

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+                       x[2], x[3], x[4], x[5], x[6], x[7], x[8], x[9]])
+        log_pdf = multivariate_normal.logpdf(phi, mean=mu, cov=cov)
+        return log_pdf
+
+    def dimension(self):
+        """ See: :meth:`LogPDF.dimension()`. """
+        return self._dimension
+
+
+class MultivariateGaussianLogPDF(pints.LogPDF):
```
MichaelClerx: Do we need this class? Or is it a copy of the version in `_log_priors.py` ?

If we do keep it, it needs a test, notebook example & docs

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+            raise ValueError('Dimensions of x must equal pdf dimensions.')
+        log_pdf = multivariate_normal.logpdf(x, mean=self._mu, cov=self._cov)
+        return log_pdf
+
+    def dimension(self):
+        """ See: :meth:`LogPDF.dimension()`. """
+        return self._dimension
+
+
+class BimodalMultivariateGaussianLogPDF(pints.LogPDF):
```
MichaelClerx: Needs:

- [x] An example in a notebook
- [ ] A test
- [ ] Sphinx documentation

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    *Extends:* :class:`LogPDF`.
+
+    Bimodal multivariate (un-normalised) Gaussian. Default is 2D with modes
+    at ``(0, 0)`` and ``(10, 10)`` with independent unit covariance matrices.
+    """
+    def __init__(self, dimension=2, mu1=[0, 0], mu2=[10, 10],
```
MichaelClerx: Again, we can get the dimension from mu1

pints/toy/_distributions.py
```diff
@@ -0,0 +1,186 @@
...
+            + multivariate_normal.pdf(x, mean=self._mu1, cov=self._cov1)
+            + multivariate_normal.pdf(x, mean=self._mu2, cov=self._cov2))
+        return log_pdf
+
+    def dimension(self):
+        """ See: :meth:`LogPDF.dimension()`. """
+        return self._dimension
+
+
+class RosenbrockLogPDF(pints.LogPDF):
```
MichaelClerx: Needs:

- [x] An example in a notebook
- [ ] A test
- [ ] Sphinx documentation

MichaelClerx:
@ben18785 @sanmitraghosh can you have a look and see if they're better now?

Also added an example to: https://github.com/pints-team/pints/blob/issue%23156-toy-pdfs/examples/toy-distributions.ipynb

PR #163, opened 2017-12-29
Changed pairwise plots to yield different plots and (slightly) more f…

ben18785:
…unctionality.

Please everyone take a look at the Jupyter notebook here and let me know your thoughts. I know we want to keep the plotting functionality simple although I think we need a few more options (for example, opacity, chain thinning and true values options) to make the basic plotting universally useful.

I think that some other packages (Stan's 'pairs' function, for example) is to rigid in its plotting and so users often avoid it. The way I have changed this plotting routine means it now does everything I want.

It's also much much quicker to run than what we had before for the KDE stuff, and I also think the result is more informative.

Look forward to your feedback!


MichaelClerx:
Looking good! But why the new notebook? We should just update the old one :-)

pints/plot.py
```diff
@@ -198,11 +198,19 @@ def pairwise(chain, kde=False):
     """
     import matplotlib.pyplot as plt
     import numpy as np
-    from scipy import stats
+    import scipy.stats as st
+
+    if thinning is not None:
```
MichaelClerx: Do we really need this? It's not hard for users to just do `pairwise(chain[::4])` instead.

MichaelClerx:
If we can drop the thinning parameter, and update the existing plotting example instead of adding a new one, I think this can be merged!

@chonlei what do you think?

ben18785:
I'd prefer to keep the thinning parameter, to be honest. I agree users can do it themselves but it is just neat to do in one function. I use it all the time since it saves me having to remember the numpy command (I imagine other people might be in the same boat). Does it matter if we have an extra argument that has a default of None?

MichaelClerx:
It hides something really simple (`x[::4]` is plain python even, not numpy specific) inside a method. So now users won't know _exactly_ what it does (for example it doesn't do `x[1::4]` or take some random selection of samples) unless they read the documentation or the source code. It also means we've got one more argument to document and test.

We spent a lot of time cutting things _out_ of the plot method signatures because there are lots of things that specific users will find convenient, but if we add a parameter for each we end up with massive method signatures.

ben18785:
It does but this is a functionality that is always (since people run 100,000s of samples) needed. 

It's one extra argument to the plot function - I'm not suggesting we add any more - and one that I personally think is very useful. Since all our sampling methods have a thinning rate, I don't think it's unreasonable that a plot one should too.

Again, I think it's easy to forget what people (like me!) won't know in terms of Python code.

MichaelClerx:
But why would we duplicate the same functionality in lots of places? It just creates lots of scope for errors. Why don't we set up the examples showing people how to do thinning?

To me this feels like we've got a program to plot a graph of `f(x)` versus `x`, and now we're adding methods to plot `1 + f(x)` versus `x`, next we'll want a method for `n + f(x)`, `2 * f(x)` etc.

ben18785:
But there's something I'm not understanding, if people will do the thinning manually before plotting - and from my experience this is almost always required - why not have one simple plot function argument that does this? It'd make people's lives easier (particularly mine).

MichaelClerx:
Is it very common to do the thinning only for plots? (Leaving the 'real' data unaffected)

MichaelClerx:
Alright, I've 

- cleaned up the code (adding bits of documentation)
- updated the example notebooks to use the new code
- resolved conflicts with the master branch
- removed thinning

OK to merge now @chonlei and @ben18785 ?

chonlei:
Hang on a minute...

chonlei:
Spotted a bug with `kde=False`, the scale makes no sense. Now fixed it. Should be good to merge!

MichaelClerx:
Good catch @chonlei ! Thanks!

PR #198, opened 2018-01-29
Issue#132 ask and tell

MichaelClerx:
Initial PR for ask-and-tell optimisers. Will add a similar interface for MCMC before this can be merged. Happy to hear thoughts so far!


mirams:
Ha ha! Brutal!

ask.py
```diff
@@ -0,0 +1,77 @@
...
+# Create an adaptive covariance MCMC routine
+x0 = real_parameters * 1.1
+
+
+mcmc = pints.AdaptiveCovarianceMCMC(x0)
+mcmc.set_acceptance_rate(0.3)
+
+# Start adapting after 1000 iterations
+#TODO
+#mcmc.set_non_adaptive_iterations(1000)
```
martinjrobins: I presume the commented out bits are not in the new interface? They all seem like generic things (burn-in, thinning etc) that can be handled in a wrapper function for convenience

ask.py
```diff
@@ -0,0 +1,77 @@
...
+
+chain = []
+for i in range(4000):
```
martinjrobins: Should the user ask the algorithm to do the covariance adaption somewhere in this loop? Either a function that says `do_adaption`, or a modal switch that says `turn_adaption_on`

pints/_mcmc/__init__.py
```diff
@@ -12,6 +12,108 @@
...
+        """
+        Returns this method's full name.
+        """
+        raise NotImplementedError
+   
+    def tell(self, fx):
+        """
+        Performs an iteration of the MCMC algorithm, using the evaluation
+        ``fx`` of the point ``x`` previously specified by ``ask``. Returns the
+        sample ``x`` accepted by the method.
```
martinjrobins: what does it return if the sample ``x`` is not accepted?

pints/_mcmc/__init__.py
```diff
@@ -12,6 +12,108 @@
...
+        self._adaptation = False
+    
+    def adaptation(self):
+        """
+        Returns ``True`` if (and only if) this sampler is currently in
+        adaptation mode.
+        """
+        return self._adaptation
+    
+    def set_adaptation(self, adaptation):
```
martinjrobins: Yes, something like this! The modal approach seems to involve a lot of code, internal state variable, two separate functions. It might be simpler to have only a `do_adaption_now` type method? Up to you, I'm not sure what is the best way to go

ask.py
MichaelClerx: Yup! Just leaving them in for now so I now which bits to stick in the wrapper!

ask.py
MichaelClerx: I think so yes! But we can maybe take that into account in the wrapper. Not sure how, will think about it!

pints/_mcmc/__init__.py
MichaelClerx: Woops! It returns the next sample in the chain, regardless of whether the proposal was accepted. Will update this!

MichaelClerx:
First version of the adaptive MCMC routine up now. The class `MCMCSampling` performs an MCMC sampling using multiple chains. Stopping criterion is still number of iterations, but should be easy now to add something more complicated @ben18785 

(It would go here: https://github.com/pints-team/pints/blob/issue%23132-ask-and-tell/pints/_mcmc/__init__.py#L320 )

martinjrobins:
@MichaelClerx walked me through these changes this morning and all looks good to merge (once tests have passed)

Issue #113, opened 2017-12-04
writing-a-model.ipynb not working on Windows


chonlei:
Just tested it again. All examples work except `writing-a-model.ipynb`! This is what the output looks like 

![image](https://user-images.githubusercontent.com/26725646/33554144-78c69174-d8f3-11e7-9033-9365009f260a.png)

I guess, this is quite an issue if `toy.LogisticModel()` works but self defined model does not.

chonlei:
Just did a few more test. It seems like our MCMC routine works perfectly fine -- but CMAES and SNES, which use `pickle` and `multiprocessing`, are not compatible with it. 

Any thoughts? If this is users' problem then close this issue, if not, then we probably want to add a note on this potential problem. I guess Windows anaconda users would not be the minority.

MichaelClerx:
The relevant note is here: http://pints.readthedocs.io/en/latest/function_evaluation.html?highlight=evaluator#pints.ParallelEvaluator

This is an issue with parallelisation on Windows, will do some googling to see if there's an easy fix for it in the examples...

chonlei:
More error output if it helps to decide is it users' problem or something we should step into.
```
Run 1: population size 8
(4_w,8)-aCMA-ES (mu_w=2.6,w_1=52%) in dimension 2 (seed=526419, Mon Dec 04 13:16:01 2017)
Traceback (most recent call last):

  File "<ipython-input-9-e03f19ad2d37>", line 1, in <module>
    runfile('C:/Users/cholei/Desktop/test.py', wdir='C:/Users/cholei/Desktop')

  File "C:\ProgramData\Anaconda2\lib\site-packages\spyder\utils\site\sitecustomize.py", line 710, in runfile
    execfile(filename, namespace)

  File "C:\ProgramData\Anaconda2\lib\site-packages\spyder\utils\site\sitecustomize.py", line 86, in execfile
    exec(compile(scripttext, filename, 'exec'), glob, loc)

  File "C:/Users/cholei/Desktop/test.py", line 31, in <module>
    sigma0,

  File "c:\users\cholei\documents\git\pints\pints\_optimisers\_cmaes.py", line 152, in cmaes
    return CMAES(function, boundaries, x0, sigma0).run()

  File "c:\users\cholei\documents\git\pints\pints\_optimisers\_cmaes.py", line 127, in run
    es.tell(candidates, evaluator.evaluate(candidates))

[...]

PR #210, opened 2018-02-06
Writing a model for Windows

chonlei:
An example of showing how to avoid Windows multiprocessing and fork() problem when defining a user model class. See #113

### DIFF
--- a/examples/writing-a-model-for-Windows.py
+++ b/examples/writing-a-model-for-Windows.py
@@ -2,34 +2,34 @@
 # coding: utf-8
 
 # # Writing a model for Windows users
-# 
+#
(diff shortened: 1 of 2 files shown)


MichaelClerx:
I don't think this is a good idea! It drags in all sorts of problems we're not responsible for. Let's just disable parallelism in the example and deal with it later

MichaelClerx:
I don't think having special example scripts for parallel optimisation is a good idea (yet)!
The example is about writing models, not about pickling, multiprocessing, forking on windows, etc.
We should just set parallel=False in this example and if anyone asks we can show them where to look!

PR #233, opened 2018-02-13
Bugfix for error when using logpdf in optimisation

MichaelClerx:
See #232

### DIFF
--- a/pints/_optimisers/__init__.py
+++ b/pints/_optimisers/__init__.py
@@ -186,9 +186,9 @@ class Optimisation(object):
                 ' optimise.')
 
         # Store function
-        if isinstance(function, pints.LogLikelihood):
+        if isinstance(function, pints.LogPDF):
             # Likelihood function given? Then wrap an inverter around it
-            self._function = pints.LogLikelihoodBasedError(function)
+            self._function = pints.ProbabilityBasedError(function)
             # TODO: Do this internally, so that we can return the correct
             # (inverted) value!?
         else:
--- a/pints/_optimisers/__init__.py
+++ b/pints/_optimisers/__init__.py
@@ -153,12 +153,14 @@ class PopulationBasedOptimiser(Optimiser):
 
 class Optimisation(object):
     """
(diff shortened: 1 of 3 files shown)


MichaelClerx:
Closes #232

pints/_optimisers/__init__.py
```diff
@@ -186,9 +186,9 @@ def __init__(
                 ' optimise.')
 
         # Store function
-        if isinstance(function, pints.LogLikelihood):
+        if isinstance(function, pints.LogPDF):
             # Likelihood function given? Then wrap an inverter around it
-            self._function = pints.LogLikelihoodBasedError(function)
+            self._function = pints.ProbabilityBasedError(function)
```
chonlei: Can we have some kind of stdout or print statement saying _explicitly_ that we are now maximising the input `function`? And vise versa (say now minimising the function when it is not pints.LogPDF). Either here or later when it actually starts the optimisation (or least under `verbose` option, if we still have it here).

We have "Optimisation" as the name, when debugging, something people might be not sure if it is their function's problem (i.e. the objective function is doing weird things) or they just made a silly mistake that they are minimising something that they should maximise.

I found this would be helpful particularly when the program is trying to do something smart for me -- because something I might try to be smart that undo what the program nicely did for me... So might happen to other users as well.

test/test_pso.py
```diff
@@ -132,6 +132,26 @@ def test_stopping_no_criterion(self):
         opt.set_max_unchanged_iterations(None)
         self.assertRaises(ValueError, opt.run)
 
+    def test_logpdf(self):
```
chonlei: Just curious, are we going to test this for other optimisation algorithms?

pints/_optimisers/__init__.py
MichaelClerx: Good point! Have added it to the documentation since you used it, but will also add a line to the verbose output!

test/test_pso.py
MichaelClerx: Yeah it should become part of the the other methods too, but as there are other tickets open about this I didn't want to do it here!

Issue #200, opened 2018-01-30
LogPrior.sample() is not implemented for all priors


MichaelClerx:
Not sure we had a discussion about this method either

MichaelClerx:
If it should be there, and we want to be able to sample multiple points, then what should the return type be when sampling a single point? A vector (1d) or a matrix (2d, but with length 1 in 1 dimension)...

MichaelClerx:
Maybe we should have an abstract class that adds/defines this method. I don't think every LogPrior will have one, for example user defined ones or other complex, unnormalised priors...

MichaelClerx:
I'll implement it for our current methods, and add a note saying users don't need it for all methods....

MichaelClerx:
Closed in #235

PR #235, opened 2018-02-13
Added sample() methods and tests to all LogPrior classes

MichaelClerx:
The LogPrior classes needed tests, and not all of them had the new `sample()` method.

See #200

### DIFF
--- a/pints/_log_pdfs.py
+++ b/pints/_log_pdfs.py
@@ -37,18 +37,22 @@ class LogPrior(LogPDF):
     Represents the natural logarithm ``log(f(theta))`` of a known probability
     density function ``f(theta)``.
 
-    Priors are normalised: The integral ``f(theta)`` over all points ``theta``
-    in parameter space sums to 1.
+    Priors are *usually* normalised (i.e. the integral ``f(theta)`` over all
+    points ``theta`` in parameter space sums to 1), but this is not a strict
+    requirement.
     """
     def sample(self, n=1):
         """
         Returns ``n`` random samples from the underlying prior distribution.
+
+        The returned value is a numpy array with shape ``(n, d)`` where ``n``
+        is the requested number of samples, and ``d`` is the dimension of the
+        prior.
+
+        Note: This method is optional, in the sense that only a subsets of
+        inference methods require it.
         """
         raise NotImplementedError
-        #TODO Decide if `n` is needed
-        #TODO For n=1, decide whether to return (n, dim) array or (dim, ) array
-        #TODO Make this clear in the docstring!
-        #TODO Add test to ensure this is present for all priors.
 
 
 class LogLikelihood(LogPDF):
--- a/pints/_log_priors.py
+++ b/pints/_log_priors.py
@@ -11,6 +11,7 @@ from __future__ import print_function, unicode_literals
 import pints
 import numpy as np
 import scipy
+import scipy.stats
 
 
 class ComposedLogPrior(pints.LogPrior):
@@ -38,10 +39,6 @@ class ComposedLogPrior(pints.LogPrior):
         # Store
         self._priors = priors
 
-    def dimension(self):
-        """ See :meth:`LogPrior.dimension()`. """
-        return self._dimension
-
     def __call__(self, x):
         output = 0
         lo = hi = 0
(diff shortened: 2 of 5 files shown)


pints/_log_pdfs.py
```diff
@@ -50,12 +51,15 @@ def __call__(self, x):
     def sample(self, n=1):
         """
         Returns ``n`` random samples from the underlying prior distribution.
+
+        The returned value is a numpy array with shape ``(n, d)`` where ``n``
```
chonlei: So even `n=1` it would be in the shape of (1, d)? And if `d=1`, we are not returning a scalar float number but an array of 1 value?

MichaelClerx: Yup! This way it's always consistent!

pints/_log_priors.py
```diff
@@ -104,22 +121,21 @@ def __init__(self, mean, variance):
...
         """ See :meth:`LogPrior.dimension()`. """
         return 1
 
-    def __call__(self, x):
-        return self._offset - self._factor * (x[0] - self._mean)**2
-
     def sample(self, n=1):
         """ See :meth:`LogPrior.sample()`. """
-        return np.random.normal(loc=self._mean, scale=self._sigma, size=n)
+        return np.random.normal(self._mean, self._sigma, size=(n, 1))
```
chonlei: Just not sure, if we sample from a **log** normal, would that be the same as sampling from a normal distribution? Because it sounds more like "we sample from a **log** normal surface" rather than from a normal distribution.

chonlei: Same thing for the multivariate normal distribution.

pints/_log_priors.py
```diff
@@ -81,15 +92,21 @@ def __init__(self, mean, covariance):
...
+
+    def sample(self, n=1):
+        """ See :meth:`LogPrior.call()`. """
+        # Note: size=n returns shape (n, d)
+        return np.random.multivariate_normal(
+            self._mean, self._covariance, size=n)
+
 
 class NormalLogPrior(pints.LogPrior):
     """
```
chonlei: Typo I think.

>For example: ``p = NormalPrior(0, 1)`` for a mean of ...

I think it should be `NormalLogPrior(0, 1)` instead.

pints/_log_priors.py
```diff
@@ -38,17 +39,27 @@ def __init__(self, *priors):
         # Store
         self._priors = priors
 
+    def __call__(self, x):
```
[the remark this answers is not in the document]
MichaelClerx: I don't see how we can add it to the documentation without making things more confusing!
Why would the users expect the order of the parameters/prior parts doesn't matter here? It matters everywhere else in Pints.

pints/_log_pdfs.py
```diff
@@ -50,12 +51,15 @@ def __call__(self, x):
...
+        is the requested number of samples, and ``d`` is the dimension of the
+        prior.
+
+        Note: This method is optional, in the sense that only a subsets of
+        inference methods require it.
```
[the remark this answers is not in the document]
MichaelClerx: Hope so! But that's not part of this PR :-)

pints/_log_priors.py
```diff
@@ -38,17 +39,27 @@ def __init__(self, *priors):
...
+        output = 0
+        lo = hi = 0
+        for prior in self._priors:
+            lo = hi
+            hi += prior.dimension()
+            output += prior(x[lo:hi])
```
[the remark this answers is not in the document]
MichaelClerx: You can easily do that. Just pass a multivariate normal prior of dimension 5, and then any other priors you like. As long as the individual prior objects are independent. This is what the doc says as well, it doesn't say the priors themselves have to be composed of independent priors per parameter!

pints/_log_priors.py
MichaelClerx: Thanks!

pints/_log_priors.py
MichaelClerx: It would be different! In that case we'd call it a LogNormalLogPrior! This is the LogPrior that you get when you use a Normal distribution, that's why it's a NormalLogPrior!

MichaelClerx:
Uhm, yeah we could something like that! But at some point we'd really just be testing numpy, since we're not responsible for the random number generation

MichaelClerx:
@ben18785 Hi Ben! I've added some really simple tests. Their purpose is to test if we're calling numpy correctly, and returning the formats/types we should return. Testing what the actual distributions are like is the responsibility of NumPy!
Hope this is ok now!

PR #246, opened 2018-02-16
Added Boundaries.sample() method, plus Boundaries tests.

MichaelClerx:


pints/_boundaries.py
```diff
@@ -65,16 +70,27 @@ def lower(self):
...
+    def sample(self, n=1):
+        """
+        Returns ``n`` random samples from the underlying prior distribution.
+
+        The returned value is a numpy array with shape ``(n, d)`` where ``n``
+        is the requested number of samples, and ``d`` is the dimension of the
+        parameter space these boundaries are defined on.
+        """
+        return np.random.uniform(
+            self._lower, self._upper, size=(n, self._dimension))
```
martinjrobins: np.random.uniform excludes the upper boundary, which contradicts the comment earlier that boundaries are inclusive. Since it is traditional that the upper boundary is exclusive, I would follow this and do the same, allowing you to use np.random.uniform as-is

MichaelClerx: Good point!

MichaelClerx:
@martinjrobins Is this better?

PR #248, opened 2018-02-19
Logging numerical (tabular) data to screen and file (Issue#228)

MichaelClerx:
The idea is:

1. The `Optimisation` and `MCMCSampling` create a `Logger` object.
2. They pass it to the underlying method (e.g. `CMAES`, `AdaptiveCovarianceMCMC`), which adds relevant fields (e.g. acceptance rate)
3. Before running, users can enable/disable screen logging, file logging identical to screen output, and csv output to file
4. Run!

Currently have implemented a `Logger` class that handles output to screen/file/csv for ints, positive ints (iteration counts), and floats.
Need to make `Optimisation` and `MCMCSampling` use them next, but curious to hear what you think.


MichaelClerx:
Example:

    import pints
    log = pints.Logger()
    log.set_filename('output.txt', csv=True)
    log.add_counter('iter', width=4)
    log.add_float('fbest')
    log.log(1, 1100.245346)
    log.log(1, 746.674587)
    log.log(1, -2.7)

Output (on screen):

    iter fbest  
    1     1100.2
    1     746.67
    1    -2.7 

Output (in csv file):

    "iter","fbest"
    1,1.10024534599999993e+03
    1,7.46674586999999974e+02
    1,-2.70000000000000018e+00


Some questions:

1. Should we make it possible to add text/status lines in between, for example warnings or statements like running with X particles, halting because of criterion X. These would then show up in screen output, and non-csv file output, but not in csv file output
2. Should we make it possible to log an output file in screen format _and _ in csv format at the same time?

MichaelClerx:
@martinjrobins and @chonlei I think this is ready for merging now!

Maybe easier to just run a few examples instead of reviewing the code itself for this one :-)

MichaelClerx:
Yup! But I don't think it's unreasonable for Pints developers to know about these things, is it?

PR #250, opened 2018-02-20
Issue#186 population mcmc

MichaelClerx:


pints/_mcmc/_population.py
```diff
@@ -0,0 +1,269 @@
...
+    chains):
+
+    1. Mutation: randomly select chain ``i`` and update the chain using a
+    Markov kernel that admits ``pi_i`` as its invariant distribution.
+
+    2. Exchange: Select another chain ``j`` at random from the remaining and
+    swap the parameter vector of ``i`` and ``j`` with probability
+    ``min(1, A)``,
+
+    ``A = pi_i(x_j) * pi_j(x_i) / (pi_i(x_i) * pi_j(x_j))``
```
martinjrobins: change `pi_i` and `pi_j` to `p_i` and `p_j`

pints/_mcmc/_population.py
```diff
@@ -0,0 +1,269 @@
...
+        """ See :meth:`Loggable._log_write`. """
+        logger.log(self._i)
+        logger.log(self._j)
+        logger.log('yes' if self._have_exchanged else 'no')
+
+    def name(self):
+        """ See :meth:`pints.MCMCSampler.name()`. """
+        return 'Population MCMC'
+
+    def schedule(self):
```
martinjrobins: is "schedule" the correct word here, it implies it is changing with time, which I thought it wouldn't?..... "distribution" perhaps?

pints/_mcmc/_population.py
ben18785: I'm easy here. I'd prefer pi, personally as this is the naming convention used in the Holmes paper, but I'm not wed to it.

pints/_mcmc/_population.py
ben18785: Yeah, agreed. How about temperature_schedule?

pints/_mcmc/_population.py
MichaelClerx: Thing is, I read (bits of) the paper too, but looking at the code a few days later still read it as p-i-i, rather than pi-i (sir)

pints/_mcmc/_population.py
martinjrobins: sounds good, that is much more meaningful

pints/_mcmc/_population.py
martinjrobins: I'd be happy with pi as long as you use something like a and b for indexing :)

pints/_mcmc/_population.py
MichaelClerx: Changed it!

pints/_mcmc/_population.py
MichaelClerx: I've gone for p now, which matches with the code!

Issue #231, opened 2018-02-13
Add methods to combine multiple independent loglikelihoods / errors


MichaelClerx:
This is for cases where we have two independent bits of information we want to combine, e.g.

```
P(theta | D, E) = P(D, E | theta) P(theta)
                  ------------------------
                          P(D, E)
                          
                = P(D | theta) P(E | theta) P(theta)
                  ----------------------------------
                               P(D, E)
```

I suggest we add

- [x] `SumOfIndependentLogLikelihoods`
- [x] `SumOfErrors` extends `ErrorMeasure`
- [x] `WeightedSumOfErrors` extends `ErrorMeasure`

the top one just sums `n` log likelihoods but has "independent" in its name as a reminder

MichaelClerx:
Issue #52 is simply writing an example that uses these features

PR #252, opened 2018-02-21
Issue#231 multi trace scores

MichaelClerx:
For issue #231 

This is about combining the results of independent single output problems defined on the same parameter space

### DIFF
--- a/docs/source/error_measures.rst
+++ b/docs/source/error_measures.rst
@@ -25,3 +25,5 @@ Example::
 
 .. autoclass:: SumOfSquaresError
 
+.. autoclass:: WeightedSumOfErrors
+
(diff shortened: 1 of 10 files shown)


MichaelClerx:
The only real difference is that you can create a SumOfErrors as `SumOfErrors(chain1, chain2, chain3)`, whereas in the weighted syntax that would become `WeightedSumOfErrors([chain1, chain2, chain3])`.
Happy to drop `SumOfErrors` though!

martinjrobins:
Yea, I would drop `SumOfErrors`, the syntax is OK enough that I don't want to have to worry about testing each class separately, and remembering to alter `WeightedSumOfErrors` whenever you alter `SumOfErrors`, and documentation and everything else!

MichaelClerx:
Have merged them both into `SumOfErrors` with optional weights!

Issue #68, opened 2017-11-20
Add methods to store / save MCMC chains


MichaelClerx:
~Couple of things that would be nice:~

- ~Somehow making sure we always have the same format output from MCMC methods (this could be as simple as updating the docs to specify it's a numpy array of a certain shape)~
- ~Being able to store chains for off-line analysis (again, could just be numpy)~
- ~Being able to load chains for off-line analysis (again, could just be numpy)~

~So the two main questions are:~
1. ~Should we have some custom datatype for this (pro: more guarantees, con: more room for errors, looks less interoperable even if we just wrap a numpy array)~
2. ~Should we provide methods for loading/saveing~

MichaelClerx:
@martinjrobins says don't implement things that are available elsewhere. Use the numpy stuff! (Or other stuff, whatever you like). Focus on things that will make Pints unique

MichaelClerx:
Had a short discussion on plotting, and consensus was to use examples / recipes instead. I think this also resolves the original question in this ticket!

Instead the issues has become: Create 1 example of storing / reloading a chain using NumPy. This would be good to do as part of the plotting example/recipe scripts #63

MichaelClerx:
Have added some code for doing this that I've been using for another project (work in progress). #254

Issue #224, opened 2018-02-12
Add rhat stopping criterion to MCMCSampling


MichaelClerx:
And show rhat in the first example code, with a tiny (one-liner or wikipedia link) explanation

MichaelClerx:
@ben18785 says default could be rhat all parameters < 1.1

Perhaps with minimum iterations before it starts using this (because if starting point is the same rhat will be 1)

MichaelClerx:
This would make things pretty tricky to implement, and nobody has been that keen on having it, so closing for now!

If we want something like this in the future, it might be better to implement it differently, e.g.:

```
samples = mcmc.run(1000)
check rhat of samples
samples += mcmc.run(1000)
etc.
```

Issue #194, opened 2018-01-18
Set something up so that optimisation/inference can be stopped and restarted


MichaelClerx:
The ask-and-tell interface (#132) takes care of part of this, but it'd also be good to store the internal state of our optimisers/inferers to disk, perhaps using [pickling](https://docs.python.org/3/library/pickle.html) ?

MichaelClerx:
#TODO: Have a quick look at https://pythonhosted.org/joblib/

MichaelClerx:
Meh

MichaelClerx:
@martinjrobins @ben18785 please add Freya to this ticket if she decides to take it up :-)

Rough idea of how to implement it:

- write it as some kind of "deep copy" of every sampler/optimiser's state
- add bits to the ask-and-tell interface that allow the current state to be obtained, and a previous state object to be restored
- add a default implementation using [copy.deepcopy](https://docs.python.org/3/library/copy.html)
- but make sure that copy.deepcopy is not required !
- figure out what to do for wrapper methods i.e. cmaes

martinjrobins:
@MichaelClerx : points 3 and 4 seem contradictory, what do you mean that copy.deepcopy is not required, its part of the standard library isn't it?

MichaelClerx:
Sorry, I meant that the interface shouldn't be written such that implementations require deepcopy. I.E. it should be possible to write a custom implementation instead

MichaelClerx:
Via a different route I've hit on https://joblib.readthedocs.io/en/latest/ again :D

PR #254, opened 2018-02-22
Simple code for loading/saving chains/sample lists

MichaelClerx:
As described in #68

### DIFF
--- a/docs/source/io.rst
+++ b/docs/source/io.rst
@@ -4,6 +4,10 @@ I/O Helper classes
 
 .. module:: pints.io
 
+.. autofunction:: load_samples
+
+.. autofunction:: save_samples
+
 .. autoclass:: StreamCapture
 
 .. autoclass:: TemporaryDirectory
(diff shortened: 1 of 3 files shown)


chonlei:
@MichaelClerx Maybe worth saving the evaluated LopPosterior for each iteration too (together with the parameters)?

MichaelClerx:
@chonlei This code is just to save/load 2d numpy arrays (of samples). No changes to the optimiser/controller code!

We could consider adding logposteriors to the logging output (separate issue!), but that might get annoying if you have 10+ evaluations per iteration (which happens easily!) which is why we usually show best.

If you want to debug what's going on with an optimiser, you can write these things yourself using ask-and-tell, but I think that's outside of scope for the normal Optimisation routine! See: https://github.com/pints-team/pints/blob/master/examples/optimisation-ask-and-tell.ipynb for an example!

MichaelClerx:
Thanks Chon!
We can consider adding more saving options, but... (1) Stopping when converged is already something we're working on #224, and (2) Stopping and restarting takes more than saving the chain, see #194. Also have a look at [tmux](https://askubuntu.com/questions/8653/how-to-keep-processes-running-after-ending-ssh-session)

Issue #247, opened 2018-02-16
CMAES Can have indefinite matrix issues during run, raises ValueError


MichaelClerx:
```
Traceback (most recent call last):
  File "./fit.py", line 187, in <module>
    p, s = opt.run()
  File "/home/michael/dev/pints/pints/_optimisers/__init__.py", line 320, in run
    xs = self._optimiser.ask()
  File "/home/michael/dev/pints/pints/_optimisers/_cmaes.py", line 58, in ask
    self._xs = np.array(self._es.ask())
  File "/home/michael/.local/lib/python2.7/site-packages/cma/evolution_strategy.py", line 1694, in ask
    pop_geno = self.ask_geno(number, xmean, sigma_fac)
  File "/home/michael/.local/lib/python2.7/site-packages/cma/evolution_strategy.py", line 1959, in ask_geno
    ary = self.sigma_vec * np.asarray(self.sm.sample(Niid))
  File "/home/michael/.local/lib/python2.7/site-packages/cma/sampler.py", line 222, in sample
    self.update_now(lazy_update_gap)
  File "/home/michael/.local/lib/python2.7/site-packages/cma/sampler.py", line 293, in update_now
    self._decompose_C()
  File "/home/michael/.local/lib/python2.7/site-packages/cma/sampler.py", line 342, in _decompose_C
    " this must be considered as a bug")
ValueError: covariance matrix was not positive definite, this must be considered as a bug
```

Could be that we're getting a warning about this but not handling it correctly? Will have to check if we're using cma-es's ask-and-tell correctly now that we're avoiding `stop()` !

MichaelClerx:
Guessing the correct behaviour here would be to terminate the optimisation, since the method got stuck somehow? If so, we'd have to update the interfaces slightly so that this is allowed.

mirams:
We aren't letting it see -inf likelihoods are we?

MichaelClerx:
That could also be a thing! Will look into this soon!

MichaelClerx:
(Which will mean setting up some more logging / debug output stuff in Pints first, I think)

MichaelClerx:
Seems @martinjrobins has been seeing this error with the HH IK model too

MichaelClerx:
> We aren't letting it see -inf likelihoods are we?
Checked it: Nope!

PR #255, opened 2018-02-23
ValueError raised by cmaes

MichaelClerx:
Error was due to cmaes getting stuck but Pints ignoring this information. Have added code to let optimiser classes tell the controller to stop! #247

### DIFF
--- a/pints/_optimisers/__init__.py
+++ b/pints/_optimisers/__init__.py
@@ -105,6 +105,14 @@ class Optimiser(pints.Loggable):
         """
         raise NotImplementedError
 
+    def stop(self):
+        """
+        Checks if this method has run into trouble and should terminate.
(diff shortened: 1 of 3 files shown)


MichaelClerx:
It's a good question! I thought about raising an Exception instead but... Users should be aware that the results can be really terrible anyway! There's no guarantee that the results are accurate, unless they do some further investigation (e.g. look at the predicted time series etc.) just like with MCMC.

I'm worried that if we occasionally say: "This result is probably bad!" the users can be fooled into thinking the other results are automatically good!

chonlei:
Hmm... I see your point, but having this `Ill-conditioned covariance matrix.` is not any good unless everyone knows what is potential problem or realise there is actually a problem with the solution. But this could be discussed somewhere else, a more general problem than this PR.

Issue #253, opened 2018-02-22
Add standard Metropolis algorithm


MichaelClerx:
See also: #116 #209

MichaelClerx:
Should be a matter of stripping bits out of the adaptive covariance mcmc method (and checking the resulting implementation against a textbook)

chonlei:
Done via #259

Issue #209, opened 2018-02-01
Hierarchy of methods


sanmitraghosh:
As requested by @MichaelClerx : 

MCMC : 
    i)AdaptiveCovariance, DREAM 
    ii) Netsed sampling 
    iii) Particle based ones--PopulationMCMC, SMC 
    iv) Likelihood free ---ABCMCMC, ABCSMC 
    v) Score function based---MALA, HMC, NUTS 
    vi) DifferentialGeomtric ones---smMALA, RMHMC 

Optimisation :
     i)PSO, GA 
     ii) Info Geometric ones ---CMAES, XNES, SNES.

I have given an exhaustive list, probably all of these wont be implemented (or needed). But you can organise them these way.

ben18785:
Thanks Sanmitra. Although think this is a more apt division:

- Sampling: MCMC, nested, importance sampling, ABC etc.
- Optimisation 

Since, for example, SMC is not MCMC, nor is nested sampling, etc. So think, keep the same two-division thing but call one 'sampling'.

Sound ok?

sanmitraghosh:
Sounds good to me, as there are overlaps anyway.

MichaelClerx:
I remember there was some structure in the information that MCMC methods needed? So an order from likelihood free up to methods needing fisher information?

ben18785:
Yep!

MichaelClerx:
Well what was it? :D

sanmitraghosh:
For MCMC methods i)---iii) only need logPosterior. Methods v) needs derivative of logPosterior and vi) Hessian of logPosterior.
iv) is a bit tricky, ideally the user should define a distance function between data and states. However, least squares is a good distance function for time series problems.

Hope this helps @MichaelClerx @ben18785

MichaelClerx:
@sanmitraghosh @ben18785 can you have a look here: http://pints.readthedocs.io/en/latest/#hierarchy-of-methods and see if you agree?

To edit the file, look in `docs/source/index.rst`

sanmitraghosh:
I think this is okay.

ben18785:
Ok, I think we need to move 'PopulationMCMC' to be under the MCMC heading. We also need to add emcee hammer and simple 'Metropolis' (i.e. the unadapted version) to the MCMC samplers.

[...]

PR #259, opened 2018-02-24
I253 standard metropolis

chonlei:
Add standard Metropolis Hasting, see #253 

This is Metropolis Random Walker ( i.e. just stripping the adaptive covariance from the AdaptiveCovarianceMCMC in #253 ) than generalised Metropolis Hasting as mentioned here #209


MichaelClerx:
Hi Chon! Is it metropolis or metropolis Hastings you've done?

chonlei:
Hi Michael! I removed the adaptive covariance part of the AdaptiveCovarianceMCMC you have mentioned in #253. I am not sure what is Metropolis. I thought it is usually known as Metropolis random walker with normal step.

MichaelClerx:
Have a look at the stuff Ben wrote in the tickets about this then! Plus wiki
etc. :) It's to do with how you choose the next points to test!



On Sat, 24 Feb 2018 04:49:54 -0800 Chon Lok Lei <notifications@github.com>
wrote

chonlei:
@MichaelClerx I had a thought about it, and added a comment about adding asymmetric proposal steps [which have not been done]. But then that requires an input of a function that we can sample from. If that is what we want, then I will try adding that flexibility to the class! (As currently we are only using Normal proposal step -- so "Random walker".) :)

I am not sure about the log transform, how and why we want to do that. @ben18785 I thought we only wanted to have that extra flexibility for users to use any kind of proposal step distribution (either symmetric like Random walker or asymmetric). Please correct me if I am wrong! :)

MichaelClerx:
No need to implement extra things! Just make sure you give it the right name!

n Sat, 24 Feb 2018 06:12:09 -0800 Chon Lok Lei <notifications@github.com> wrote

chonlei:
Ah, I see what you mean now. Okay! Will do!

MichaelClerx:
Thanks @chonlei looks great except for this bit:

    # Choose a covariance matrix for the proposal step
    # This will be the diagonose of the covariance matrix
    sigma0 = np.abs(real_parameters) * 0.000005

Looking at the results, maybe it'd be easier to use a KnownNoiseLogLikelihood in this example? Then we should be able to find a nicer guess for sigma0 perhaps?

PR #268, opened 2018-03-06
Goodwin oscillator toy model

MichaelClerx:
See: #265 

@sanmitraghosh do you know a reference for this particular form of the equations?

Example in https://github.com/pints-team/pints/blob/c278a61d15663e34158864bf70bf0668142b3759/examples/toy-goodwin-oscillator-model.ipynb


MichaelClerx:
Added citations now. Would probably be better to have a multi-series version of this problem!

Issue #298, opened 2018-03-28
implement gradients/sensitivities for toy problems


martinjrobins:
Many other tickets refer to gradient based methods (#22 #283 #293 #104). First step is to implement gradients for the toy problems

- [x] AnnulusLogPDF
- [x] ConeLogPDF
- [x] ConstantModel
- [x] EightSchoolsLogPDF
- [x] FitzhughNagumoModel
- [x] GaussianLogPDF
- [x] GermanCreditLogPDF
- [x] GermanCreditHierarchicalLogPDF
- [x] GoodwinOscillatorModel 
- [x] HighDimensionalGaussianLogPDF
- [x] LogisticModel
- [x] LotkaVolterraModel
- [x] MultimodalGaussianLogPDF
- [x] NealsFunnelLogPDF
- [x] ParabolicError
- [x] RosenbrockLogPDF
- [x] SimpleHarmonicOscillatorModel
- [x] SimpleEggBoxLogPDF
- [x] TwistedGaussianLogPDF

Not doing:
- ActionPotentialModel: odeint
- Hes1Model: odeint
- HodgkinHuxleyIKModel: Separate ticket #650 
- MarkovJumpModel: stochastic simulation algorithm
- RepressilatorModel: stochastic simulation algorithm
- RosenbrockError: Can use pdf instead
- SIRModel: stochastic simulation algorithm

MichaelClerx:
I don't think we need it for all problems, do we? I don't want us to get carried away doing lots of nice gradient-based stuff when in real life we hardly ever have them.

(I know there's ways to get them, but they've been around for decades and still don't see much use, that worries me!)

martinjrobins:
probably start by doing for one easy problem and one hard one to get an idea of how the gradient-based stuff works (presumably after that we can just say "nope, doesn't work").

MichaelClerx:
See also #43 #118

martinjrobins:
I'll wait until @sanmitraghosh verifies the correctness of the fitzhugh nagumo sensitivities before I merge it in...

ben18785:
Good stuff @martinjrobins - thanks for doing this.

sanmitraghosh:
@martinjrobins corrected the FN model in your branch. Have a look.

sanmitraghosh:
I think its pretty easy to test if the sensitivities are correct by using theano gradient test. I am attaching here a theano test file I have written for the FN model.
[...]

PR #299, opened 2018-03-28
#298 added sensitivities for logistic model

martinjrobins:
Does this look like a reasonable way to add sensitivities to the models? I've just done the logistic equation for now


MichaelClerx:
Not sure about this! I think in most cases sensitivities will be calculated at the same time as the ODE solutions, so might be better to have e.g. `simulate(parameters, times, sensitivities=True)`

martinjrobins:
Yea, that sounds better. so return a tuple of `(values,sensitivities)` if `sensitivities=True`?

martinjrobins:
How is this @MichaelClerx?

pints/toy/_logistic_model.py
```diff
@@ -18,10 +18,21 @@ class LogisticModel(pints.ForwardModel):
...
+        f(t) &= \\frac{k}{1+(k/p_0 - 1)*\exp(-r t)} \\\\
+        \\frac{df(t)}{dr} &= \\frac{k t (k / p_0 - 1) \exp(-r t)}
+                              {((k/p_0-1) \exp(-r t) + 1)^2} \\\\
+        \\frac{df(t)}{dk} &= \\frac{k \exp(-r t)}
+                              {p_0 ((k/p_0-1)\exp(-r t) + 1)^2}
+                             + \\frac{1}{(k/p_0 - 1)\exp(-r t) + 1}
+
+    Has two parameters: A growth rate :math:`r` and a carrying capacity
+    :math:`k`. The initial population size :math:`f(0) = p_0` can be set using
+    the (optional) named constructor arg ``initial_population_size``
```
MichaelClerx: Nice!

MichaelClerx:
Looks good! Only thing I'm wondering is if it might be better to have a `class ForwardModelWithSensitivities(ForwardModel)` so that we can check the suitability of the model.

However.... Not sure how this will work out in the loglikelihood / error measure etc., so let's wait and see. Maybe add a note for model implementers that sensitivities aren't required?

martinjrobins:
well, at the moment if you try to run `simulate_with_sensitivies` on a model that doesn't have it you get a not implemented exception, which is pretty spot on ;)

MichaelClerx:
@ben18785 @sanmitraghosh Can you have a look and see if this set-up will be able to satisfy your 1st order sensitivity needs?

PR #314, opened 2018-03-31
Issue#302 student t

ben18785:


ben18785:
In this pull request, I have also put a model called 'constant_model' in the toys, along with a notebook for it. This function is used in the testing for the log_likelihoods for the Student t.

MichaelClerx:
Can we change the filename `Student-t log-likelihood.ipynb` to something more in line with the rest of the examples? (no spaces, some kind of category prefix)

MichaelClerx:
Would also like to see the two constant-model examples merged! (If we keep it, I still really don't like how this violates all the rules about models)

pints/_log_likelihoods.py
```diff
@@ -105,6 +106,61 @@ def __call__(self, x):
...
+        \\frac{x_i - f(\\theta)}{\sigma}^2)
+
+    where B(.,.) is a beta function.
+
+    Arguments:
+
+    ``problem``
+    A :class:`SingleOutputProblem` or :class`MultiOutputProblem`. For a
+    single-output problem two parameters is added (nu, sigma), where nu is
+    the dgrees of freedom and sigma is scale, for a multi-output
```
MichaelClerx: dgrees --> degrees

pints/_log_likelihoods.py
```diff
@@ -105,6 +106,61 @@ def __call__(self, x):
...
+        # Add parameters to problem (two for each output)
+        self._dimension = problem.n_parameters() + 2 * self._no
+
+        # Pre-calculate
+        self._N = len(self._times)
+
+    def __call__(self, x):
+        # For multiparameter problems the parameters are stored as
+        # (nu_1, sigma_1, nu_2, sigma_2,...)
+        params = x[-(2 * self._no):]
```
MichaelClerx: Sorry, know I said it doesn't matter, but just realised it might be better to store model parameters first, distribution parameters second, so that we stay consistent with the UnknownNoiseLogLikelihood version. Sorry!

pints/toy/_constant_model.py
```diff
@@ -0,0 +1,64 @@
...
+    This function is mostly useful for unit testing.
+
+    """
+
+    def __init__(self):
+        super(ConstantModel, self).__init__()
+        self._no = None
+        self._parameters = [0]
+
+    def parameters(self):
```
MichaelClerx: What's this method for?

pints/toy/_constant_model.py
```diff
@@ -0,0 +1,64 @@
...
+#  For licensing information, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import numpy as np
+import pints
+
+
+class ConstantModel(pints.ForwardModel):
```
MichaelClerx: If it's not too cumbersome, could we change this model so that you have to define the number of parameters and outputs when you create it? I'd like this to be constant throughout the model's lifetime, and not change everything you make a call to it. That's going to cause trouble somewhere down the line

MichaelClerx:
Could you explain a bit more about why you'd expect a student t distribution in the first place? Maybe in the notebook, maybe in the docs, maybe both...

For example, wiki says:

_"In probability and statistics, Student's t-distribution (or simply the t-distribution) is any member of a family of continuous probability distributions that arises when estimating the mean of a normally distributed population in situations where the sample size is small and population standard deviation is unknown."_

How do I connect this idea of estimating a mean to the noise in a measured signal?

In what sense is it more "robust"?

pints/_log_likelihoods.py
ben18785: They are currently like that. I admit my explanation was confusing though! I have added a new bit of explanation to the code to (hopefully) explain it more clearly.

pints/_log_likelihoods.py
ben18785: Thanks!

pints/toy/_constant_model.py
ben18785: It returns the parameters of the model. Happy to remove.

pints/toy/_constant_model.py
ben18785: So I did think (and as I said in the commit) actually try and implement this. The problem I was having with this, though, is that if, whenever we instantiate the model, we declare the parameters, this causes issues downstream with simulate. 

Imagine,

params = [0]
model = ConstantModel(params)

Then later, I want to simulate the model over a set of times (using the same set of parameters),

values = model.simulate(times)

That's fine. But what I can't do is use that same model to simulate a new set of data with new parameters because in Python you can't have an optional argument before a obligatory one. So I can't define a 'simulate' function as: simulate(self, parameters=None, times), meaning that I'd have to change the order of parameters and times and make this model different to all other models.

How about I make this model the same as all others by instantiating it with nothing.

def __init__(self):
        super(ConstantModel, self).__init__()

Then self._parameters is only set when simulate is called, like all other functions. Seem ok? That's what the latest commit does.

pints/toy/_constant_model.py
ben18785: Removed!

ben18785:
Yep, good point re: the name of the file. I've now changed it to be in the inference category.

Also good point re: explaining about the Student-t distribution. I have now added two new sections to the Jupyter notebook to, hopefully, elucidate further the differences between this distribution and the normal.

pints/toy/_constant_model.py
MichaelClerx: I think you should still pass it parameters. It should just ignore them (since the output is constant)

MichaelClerx:
Hi @ben18785 !
I've rewritten the `ConstantModel` class as we discussed earlier. Could you have a look at the notebooks and 

1. Check if the output still makes sense
2. Add something about why df=1, scale=1 is comparable to a normal with sigma=3, I would have thought sigma would be equal to the scale parameter?

Issue #181, opened 2018-01-16
Come up with a good way to validate inference methods


MichaelClerx:
Ideas:

- Compare resulting mean: norm(mean - real) should be below some tolerance
- Compare resulting variance: Should be within some range of "true" variance? We can get this for one of our test problems using Pints or STAN or some analytical solution?

- Compare our methods against each other: do they all agree (downside is it's no longer a unit test, so we'd need a test per class for the methods, and a single massive test for all the output? It would also fail when 1 of X methods fails)

~Open questions:~

- Should we seed the random generator first? Reduces the scope of our testing! (And could fail if numpy's generator fails) --> Yes
- Should we test different toy problems or just the one? --> Multiple
- Should we split the tests into a short and a long suite? --> Yes #238

MichaelClerx:
Decided we should have tests of code (do you get errors when you run stuff) separated from this kind of validation stuff

ben18785:
I have been doing some thinking on this and thought we could start with the following models to test all sampling software against:

- Correlated 3/4-D Gaussians (already in our toy models); potentially a number of models: low correlation vs high correlation,  start samplers near mode vs away from it, and so on. The benefit of using a toy model is that we know what the true distribution is and, hence, means that we don't need to do any preliminary runs with a reliable sampler to estimate its properties. 
- Logistic equation. Think have figured out how to determine a numerical deterministic solution to the posterior here -- not as simple as it sounds -- if we use flat priors on all parameters. We can compare this to the sampling one.

Any other ideas? @MichaelClerx @martinjrobins @sanmitraghosh Remember that the models need to be simple enough that even the worst samplers (so long as they are actually coded up correctly!) will converge in a reasonable amount of time to the posteriors. This isn't benchmarking.

sanmitraghosh:
I think since we are not inventing or proposing a new sampler, thus we need simple tests to see that we have coded up correctly. For that the best thing to do is to use the Banana distributions. Pretty much all adaptive samplers should be able to sample it correctly. And this is also a standard test case used in just about every paper. There are a few other non-linear analytically tractable distributions such as the ring and flower. I would get some code for these as well.

For unit test we should compare the norm of sampled vs true mean vector. Some papers also use MMD for comparison which we can incorporate within our tests.

ben18785:
Hi Sanmitra,

Let's chat about this tomorrow if you're in? My feeling is that the banana
may be too complex for some samplers (non-adaptive Metropolis and other
[...]

PR #328, opened 2018-04-04
Methods to validate sampling methods

ben18785:
Related to issue #181. This is just a start but wanted to put it here to discuss. At the moment, I have implemented tests only for adaptive covariance MCMC based on thresholds from 200 runs of Random Walk Metropolis.

I suspect that we will need different thresholds for different classes of algorithm. For example, I suspect that the population MCMC method may fail the thresholds I've used here based on Random Walk Metropolis.

We will also need different test problems for different algorithms. For example, nested sampling requires a prior and a likelihood, so we would have to use Bayesian inference examples. It might be good to build simple examples using a normal prior and likelihood around a constant parameter value to be part of this (as well as tests with, say, the logistic model).

We will also likely want to separate Travis builds into: low burden and high burden testing, since sampler validation will require much more time than we are currently using. Whether we do nightly/ever pull request builds with high burden tests, I don't know, but would be good to get peoples' thoughts.


MichaelClerx:
> I've also realised that since these tests are stochastic, we will need to apply a Bonferroni-esq correction to them to correct for conducting multiple hypothesis tests. For more testing info it may be good to speak with Fergus, as I know he's been involved in this for Chaste.

Would it not be better to fix the random seed?

MichaelClerx:
Before going into detail: This branch was merged with the student-T branch at some point. I don't think that's a great idea!

See: https://github.com/pints-team/pints/network It gets complicated

MichaelClerx:
Keep in mind that whenever you create a branch, it branches off from your current branch, not necessarily from master!

So usually you'd want to do this:

    git checkout i123-very-important
    git commit -m "Some new work on i123"
    git push
    git checkout master                   <-- this!!
    git branch i124-a-new-branch
    git checkout i124
    ...

I do this one wrong all the time :-/

PR #330, opened 2018-04-05
Sampling validation

ben18785:
Related to issue #181. This is just a start but wanted to put it here to discuss. At the moment, I have implemented tests only for adaptive covariance MCMC based on thresholds from 200 runs of Random Walk Metropolis.

I suspect that we will need different thresholds for different classes of algorithm. For example, I suspect that the population MCMC method may fail the thresholds I've used here based on Random Walk Metropolis.

We will also need different test problems for different algorithms. For example, nested sampling requires a prior and a likelihood, so we would have to use Bayesian inference examples. It might be good to build simple examples using a normal prior and likelihood around a constant parameter value to be part of this (as well as tests with, say, the logistic model).

We will also likely want to separate Travis builds into: low burden and high burden testing, since sampler validation will require much more time than we are currently using. Whether we do nightly/ever pull request builds with high burden tests, I don't know, but would be good to get peoples' thoughts.

I've also realised that since these tests are stochastic, we will need to apply a Bonferroni-esq correction to them to correct for conducting multiple hypothesis tests. For more testing info it may be good to speak with Fergus, as I know he's been involved in this for Chaste.

### DIFF
--- a/pints/toy/_high_dimensional_normal.py
+++ b/pints/toy/_high_dimensional_normal.py
@@ -1,5 +1,5 @@
 #
-# Rosenbrock error measure and log-pdf
+# N-dimensional log normal pdf
 #
 # This file is part of PINTS.
 #  Copyright (c) 2017, University of Oxford.
@@ -17,14 +17,20 @@ class HighDimensionalNormalLogPDF(pints.LogPDF):
     """
     *Extends:* :class:`pints.LogPDF`.
 
-    High-dimensional multivariate normal log pdf, with tricky off-diagonal
-    covariances.
+    N-dimensional multivariate normal log pdf with constant
+    correlation rho between
+    all dimensions. By default, rho = 0.5, and N = 100,
+    and the variance of dimension i is i.
     """
-    def __init__(self, dimension=100):
+    def __init__(self, dimension=100, correlation=0.5):
         self._dimension = int(dimension)
         if self._dimension < 1:
             raise ValueError('Dimension must be 1 or greater.')
+        if np.absolute(correlation) > 1:
+            raise ValueError('Correlation between dimensions ' +
+                             'must be less than 1 in magnitude')
 
+        self._correlation = correlation
         # Construct mean array
         self._mean = np.zeros(self._dimension)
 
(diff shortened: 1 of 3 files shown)


MichaelClerx:
In chaste, they're making sure to always seed random generators to get consistent results. I suggest we do that too!

MichaelClerx:
> We will also likely want to separate Travis builds into: low burden and high burden testing, since sampler validation will require much more time than we are currently using. Whether we do nightly/ever pull request builds with high burden tests, I don't know, but would be good to get peoples' thoughts.

We've already set up a travis "cron job" that runs the notebooks daily. It doesn't check the output but just ensures they all run without raising exceptions.

We could add another job for this kind of more stringent testing (but see above). Not sure how much computation time we can use though, as Travis is a free(-for-open-source) service.

studies/RWM_error_3d_normal_high.csv
```diff
@@ -0,0 +1,200 @@
+1.107094319886217981e+00,5.652251302905041408e-01
```
MichaelClerx: If we start doing things like this, we might need to set up a separate testing/development repo for them, so that normal users don't have to download all these files.

ben18785:
Yes, you may be right. Let's chat about this on Tuesday.

studies/RWM_error_3d_normal_high.csv
martinjrobins: I agree with not including file in the repository, its easy enough to have a separate "data" repo for these. Github can do something tricky for large file storage too, but I don't know much about this.

mirams:
For testing, given the choice of lots of random seeds or one with lots of samples, I would have thought it would be more important to put lots of samples into a fixed seed random stream, get something fairly converged, then check the distribution's moments etc.?

With some strong discipline/commenting, this could be done offlline (or nightly/weekly), and then a continuous test could just check the first N matched recorded ones (won't take many to be certain nothing has changed, a bit like a simple hash check). With some Chastey tests we have a commented out one that does something like this thoroughly, and a lightweight continuous one. If the lightweight one fails, then you could run the more complicated one to check converged distribution and record N samples.

MichaelClerx:
Just to remind @martinjrobins and @ben18785 of this PR, which could be a good basis for our new functional testing stuff!
(Which should go in this repo: https://github.com/pints-team/functional-testing )

MichaelClerx:
@ben18785 Looks like there are some things here we could move over to functional testing?

MichaelClerx:
Ok, will delete this PR and branch in a few days if nobody's interested!
@ben18785 @martinjrobins

ben18785:
Can you keep the branch as I am still using bits from it?

MichaelClerx:
Ok!

MichaelClerx:
Just re-opening this: Is there stuff in here we want to save? If so - then which bits? Should we move them somewhere or create a new issue & branch? If not, then lets just delete

ben18785:
This is now way outdated. It was the precursor to our functional testing stuff.

MichaelClerx:
Cool! Do you want to hang onto the branch or can we chuck that out too?

ben18785:
Nah chuck it

Issue #333, opened 2018-04-06
Unsupported sensitivities method shows up in docs of models


MichaelClerx:
Can workaround by adding this to all toy models without sensitivity support:

    def simulate_with_sensitivities(self, parameters, times):
        """ Not implemented. """
        raise NotImplementedError

But I think it would be better to define the `simulate_with_sensitivities` method only in a class `ForwardModelWithSensitivities` that extends `ForwardModel`

That would allow checking model support using `isinstance()` too

martinjrobins:
That is a pain. Ok, I'll go with your suggestion and implement `ForwardModelWithSensitivities`

ben18785:
Just a quick note. Do we want to return sensitivities always alongside with the forward model values? To me, in HMC etc. we only need the sensitivities in the ask side...

martinjrobins:
good question. The majority of the time you will, since for ODE models you need to integrate the forward model to get the sensitivities.  Even for analytical sensitivities, in the logistic model the forward model is a term of the sensitivities, so I calculated it anyway.

ben18785:
Cool. Yep, makes sense. Ignore me!

MichaelClerx:
Ok! Will remove my workaround again then and close this :-)

PR #334, opened 2018-04-06
#333 create new class ForwardModelWithSensitivities

martinjrobins:


martinjrobins:
where should I be adding documentation? I've added a docstring for the class...
Yea, the logistic and fitz tests cover the new class

MichaelClerx:
https://github.com/pints-team/pints/blob/master/docs/source/core_classes_and_methods.rst

You can test with:

    cd docs
    make html

Then check the output in `docs/build/html` in your browser (see also [here](https://github.com/pints-team/pints/blob/master/CONTRIBUTING.md#documentation))

PR #348, opened 2018-04-26
i43 Sensitivities: Alternative implementation

MichaelClerx:
I like this one much better than #347, it keeps everything separated.

Note that in #347 we need to make changes to all the log likelihoods, all the toy problems etc.

-------------------------

Could all interested parties please have a look at this and #347 ? (Probably easiest to look at "Files changed") @martinjrobins @ben18785 @sanmitraghosh @chonlei ?

### DIFF
--- a/docs/source/core_classes_and_methods.rst
+++ b/docs/source/core_classes_and_methods.rst
@@ -9,15 +9,20 @@ Forward model
 
 .. autoclass:: ForwardModel
 
-Forward model with sensitivities
-********************************
-
-.. autoclass:: ForwardModelWithSensitivities
-
 Problems
 ********
 
 .. autoclass:: SingleOutputProblem
 
 .. autoclass:: MultiOutputProblem
+
+
+Models and problems with sensitivities
+**************************************
+
+.. autoclass:: ForwardModelS1
+
+.. autoclass:: SingleOutputProblemS1
+
+.. autoclass:: MultiOutputProblemS1
 
--- a/pints/__init__.py
+++ b/pints/__init__.py
@@ -41,8 +41,13 @@ FLOAT_FORMAT = '{: .17e}'
 #
 # Core classes
 #
-from ._core import ForwardModel, ForwardModelWithSensitivities
-from ._core import SingleOutputProblem, MultiOutputProblem
+from ._core import ForwardModel, ForwardModelS1
+from ._core import (
+    SingleOutputProblem,
+    SingleOutputProblemS1,
+    MultiOutputProblem,
+    MultiOutputProblemS1,
+)
 
 #
 # Utility classes and methods
@@ -58,8 +63,10 @@ from ._log_pdfs import (
     LogPDF,
     LogPrior,
     LogLikelihood,
+    LogLikelihoodS1,
     LogPosterior,
     ProblemLogLikelihood,
+    ProblemLogLikelihoodS1,
 )
 
 #
--- a/pints/_core.py
+++ b/pints/_core.py
@@ -20,15 +20,8 @@ class ForwardModel(object):
     directly in Python or interface with other languages (for example via
     Python wrappers around C code).
     """
-
     def __init__(self):
         super(ForwardModel, self).__init__()
-
-    def n_parameters(self):
-        """
-        Returns the dimension of the parameter space.
-        """
-        raise NotImplementedError
 
     def simulate(self, parameters, times):
         """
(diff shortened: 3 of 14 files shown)


MichaelClerx:
@ben18785 says:

> For me, from a user's perspective, I'd prefer to have the likelihoods all under one banner, rather than defining separate likelihoods for those with/without sensitivities. Then it makes it easier for a user to chop and change sampling algorithms without having to change likelihoods. Also, from a statistical point of view, they're not different likelihoods, which could, I think, be confusing to a user.
>
> Overall, I would quite prefer to have them together rather than having separate classes.

MichaelClerx:
For me they're different things though! An object that returns a vector is different from an object that returns a vector and/or a matrix. A model that can calculate sensitivities is certainly different from one that can't... @martinjrobins also brought up the fact that having a single set of classes would make it easier to switch between methods, but even then you'd probably want to disable the sensitivity output when not needed, so the model the user needs to write gets a lot more complex.
I think the core issue is that all one-set-of-classes designs we've had make it so that everyone always has to think about 1st and 2nd order sensitivities, all the time.

ben18785:
Can we not disable the sensitivity output by default? Then the user
wouldn't have to think about it unless needed. (I suspect I am missing
something here, sorry.)

MichaelClerx:
Well as soon as you say a model can provide either just a solution, OR a solution and sensitivities, then all the classes that build on a model need to start asking which it is. So then even in the cases where the implementation is simple for the user, the documentation still isn't because it contains all this stuff about the optional extra sensitivities they could implement. And that then propagates upwards. Have a look at #347 for example

ben18785:
I personally think the documentation you've provided is pretty clear (and,
hopefully, should be obvious to a user). As Martin says, it's also about
not duplicating workload when we introduce a new likelihood (thinking of
others who will contribute to Pints as well).

MichaelClerx:
I think you've got that exactly the wrong way around!
In this proposal, you can easily add a new loglikelihood or error measure without sensitivities. If you later want to add sensitivities you copy paste (hardly doubling the work load!) and add that functionality then. 
On the other hand, if all these methods need to know about sensitivities it's impossible to write a sensitivity free one: at a minimum, you'd have to add checks and error messages to say they're not supported.

I really don't think ODEs are commonly solved with sensitivities, so it seems a bad idea to build Pints in a way that assumes people do this all the time.

ben18785:
I think that ODEs (just not the ones encountered in electrophysiology, for
example) can often be much more efficiently sampled from using
sensitivities. But, I suppose, we could make the argument that Pints is for
time series problems which are "hard" meaning that sensitivities are too
expensive to calculate to be worthwhile.

Yes, you are right to point out that it isn't doubling the workload! But,
nonetheless, creating separate likelihoods does mean that we need a
separate set of tests etc.

All the aforementioned here is not what I'm really concerned about though.
My concern is for a user, not us as developers. For a user/statistician, I
believe that all should differ between samplers should be the sampling
method. When you start saying "when using adaptive covariance use
ErrorWithoutSensitivity. When using HMC use ErrorWithSensitivity" things
(to me, at least) get complicated. Also, as I said before, it gives the
(false) impression that we are using different likelihoods for different
samplers, when we're not.

Also, imagine that you are in the position (as a new user of Pints) when
you want to do a sweep over lots of different samplers, to see which fits
best. If we go with separate classes for models with/without sensitivities,
then it's not straightforward to do such a sweep.

MichaelClerx:
Thanks Ben, I agree what matters is the users!

But really do believe that having separate classes is simpler for users! For example, say you saw all the sensitivity stuff and wrote an ODE solver that could calculate them, and then settled on CMAES as the best optimiser and started using AdaptiveCovarianceMCMC. At no point in the process would Pints have said: I don't need all those sensitivities!

If you want a model that can do both, you'd have to do a bit of programming to create a wrapper that fits in both interfaces (or 2 wrappers) with this design. But in the other designs _everyone_ needs to do that. That's why I think leaving the sensitivity stuff separate will make life easier for the users

ben18785:
That makes sense. Thanks Michael!

I completely agree that it would be easier if a user can choose a model
with or without sensitivities, and doesn't need to do any wrapping. Is
there no way that this can still happen and a likelihood function can just
check what it is being supplied with though: a vector (then it's a method
without sensitivities), a vector and a matrix (1st order sens.) etc.?

If not, am happy to go with what you suggest.

PR #351, opened 2018-04-26
Goodwin model now exposes 3 states

MichaelClerx:
@sanmitraghosh if you like you can play around with the MCMC a bit! I've got a feeling adaptive covariance might find it with more iterations.


MichaelClerx:
Thanks @martinjrobins ! I've removed the unnecessary code and replaced CMA-ES by XNES in the example.

Unfortunately we can't disable warnings etc. from cmaes entirely (without capturing them). There's a `verbose` argument to pycma but it isn't implemented consistently (see https://github.com/CMA-ES/pycma/blob/master/cma/evolution_strategy.py and look for `verbos` or `print_warning`)

Issue #317, opened 2018-04-02
Move test-only classes to something like `shared.py` in the tests folder


MichaelClerx:
Like @chonlei suggested a while ago

PR #352, opened 2018-04-26
Moved TemporaryDirectory and StreamCapture out of pints, into shared testing module

MichaelClerx:
See #317


MichaelClerx:
Coverage doesn't actually go down, it's just that covered lines have been removed from pints.

PR #353, opened 2018-04-27
Rewrote noise adders as generators.

MichaelClerx:
See #222

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -34,6 +34,7 @@ Contents
     log_priors
     mcmc_samplers/index
     nested_samplers/index
+    noise_generators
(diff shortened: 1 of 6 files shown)


MichaelClerx:
It'd be good if someone could have a look at this!

MichaelClerx:
Fair enough! I wasn't thinking of python generators when I called them that. Seems they're only called that in some of the docs though! Is it ok if I just leave it?

martinjrobins:
yea, happy to leave it as noise generators. I was mainly confused because I was looking forward to seeing a generator implemented in python ;) But "noise generator" is also a perfectly valid descriptive term.

Issue #327, opened 2018-04-04
Add lotke-volterra toy model


MichaelClerx:
See also: http://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html

PR #354, opened 2018-04-27
Added Lotka-Volterra toy model

MichaelClerx:
See #327


pints/toy/_lotka_volterra_model.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _rhs(self, state, time, parameters):
+        """
+        Right-hand side equation of the ode to solve.
+        """
+        x, y = state
+        a, b, c, d = parameters
+        return np.array([a * x - b * x * y, -c * y + d * x * y])
+
+    def set_initial_conditions(self, y0):
```
chonlei: Hmm, if we use this `set_initial_conditions()` type interface, should we add `get_initial_conditions()` too? Or we just put the whole thing into `__init__()`...

MichaelClerx: You can add it if you like, but call it `initial_conditions()` (no `get_`)

pints/toy/_lotka_volterra_model.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Changes the initial conditions for this model.
+        """
+        a, b = y0
+        if a < 0 or b < 0:
+            raise ValueError('Initial populations cannot be negative.')
+        self._y0 = [a, b]
+
+    def simulate(self, parameters, times):
```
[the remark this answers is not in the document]
MichaelClerx: Done!

PR #358, opened 2018-04-28
i43 sensitivities: A third proposal

MichaelClerx:
This one stays much closer to (basically _is_) what @martinjrobins did before, and lets users use the same LogPDF, Problem, etc. for methods with and without sensitivities (provided the used objects all support them), like @ben18785 wanted. It also allows models to implement both the `ForwardModel` and the `ForwardModelS1` class, if they want.

I've also coded up a loglikelihood with sensitivities, which wasn't fun. It did make me think that maybe we should swap the order of the multi-dimensional array indices we're using.
At the moment it's `(n_times, n_outputs)` for a multioutput problem, so `y[10]` gives you both outputs at time 10.
For sensitivities, that then becomes `(n_times, n_outputs, n_parameters)`, so that `dy[10, 2]` gives you the derivatives w.r.t p at time 10, for output 2.
That made (and makes) sense to me, and is probably how you generate the things when running simulations, but if you have a look at [this d/dp loglikelihood calculation](https://github.com/pints-team/pints/blob/i43-sensitivities-alt-2/pints/_log_likelihoods.py#L84-L86) you'll see *5* transpose operations on a single line, which makes me think we should have chosen the opposite ordering.

Would also be very grateful for any numerical examples to test with at this point. I've done some plots and they look fine. But a known example would improve our tests.

Please let me know what you think @martinjrobins @ben18785 !
If you like it, I'll try wrapping a derivative-based optimisation from scipy and test it on the logistic model, just to make sure everything works.

Then there's the long process of adding `evaluateS1` methods to all priors and loglikelihoods we've implemented!

### DIFF
--- a/docs/source/core_classes_and_methods.rst
+++ b/docs/source/core_classes_and_methods.rst
@@ -12,7 +12,7 @@ Forward model
 Forward model with sensitivities
 ********************************
 
-.. autoclass:: ForwardModelWithSensitivities
+.. autoclass:: ForwardModelS1
 
 Problems
 ********
--- a/pints/__init__.py
+++ b/pints/__init__.py
@@ -41,7 +41,7 @@ FLOAT_FORMAT = '{: .17e}'
 #
 # Core classes
 #
-from ._core import ForwardModel, ForwardModelWithSensitivities
+from ._core import ForwardModel, ForwardModelS1
 from ._core import SingleOutputProblem, MultiOutputProblem
 
 #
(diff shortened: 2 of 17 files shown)


MichaelClerx:
> But why did we move away from this in the first place?
Not sure! Think I started from the assumption that e.g. a `LogPDF` would return sensitivities via it's `__call__` function!

> To be consistent with the higher level classes, there should be no extra ForwardModelWithSensitivities class, just use an extra simulateS1 function in a ForwardModel
I left the extra interface in there to make it more obvious this is an optional extra!
Otherwise you'd get the impression that a ForwardModel isn't complete without first & second order sensitivities!

MichaelClerx:
Ben approval in email, informal consent all round, so merging!

PR #370, opened 2018-05-01
Added SIR epidemiology model

MichaelClerx:
An interesting thing here is that we're pretending the data has noise. It doesn't (we can count exactly who has the disease)!

Anyone know how to deal with this? @ben18785 @sanmitraghosh ?


ben18785:
In practice with outbreaks/epidemiological surveys, we never know the true
number of people infected (there is uncertainty in the timing of infection,
the numbers of people screened, whether asymptomatic individuals contribute
to the infectious pool etc.) and so it's fine to assume uncertainty around
the true value.

Issue #373, opened 2018-05-01
Get pints.plot test coverage to 100%


MichaelClerx:
Get pints.plot test coverage to 100%

Don't have to inspect output, just make sure all options run without exceptions.

PR #375, opened 2018-05-02
Update plot series

### LINKED ISSUE #365: Add `ref_parameters` argument to `series()` plot
Just before `n_percentiles` (like in `pairwise`)

### LINKED ISSUE #369: Update color scheme and labels in `series` plot for multiple outputs
As you can see in the repressilator example (PR #368 ), the series plot isn't quite ready for multi-output problems.

1. The label for each output is the same
2. The colour for each output is the same

Two options:

1. Update the labels, removing most of them, and cycle through a colorscheme (make sure this works in matplotlib 1 and 2)
2. (Probably easier), give the figure as many subplots as outputs

chonlei:
This PR will
Close #365 
Close #369 

I have not yet tested it, I will re-run some examples to make sure it works first. Will also consider doing #373, but depends -- might do it as a separate PR if it makes things easier.


MichaelClerx:
> Will also consider doing #373, but depends -- might do it as a separate PR if it makes things easier.

Keep it as a separate PR. But add tests in this PR until you get 100% cover of the code you've added now! We can't accept PRs that make coverage worse :-)

PR #379, opened 2018-05-29
Add `fmin()` and `curve_fit()` convenience methods

MichaelClerx:
To be used for quick & easy minimisation and curve fitting.

Similar to e.g. https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin.html and https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html

These methods _do not_ provide the same level of control that the main interfaces do (that's what the main interfaces are for), but are useful when you're trying to fit some function and want to try out how Pints optimisers do.


test/test_easy_optimisation.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Basic test
+        def f(x):
+            return (x[0] - 3) ** 2 + (x[1] + 5) ** 2
+        xopt, fopt = pints.fmin(f, [1, 1], method=pints.XNES)
+        self.assertAlmostEqual(xopt[0], 3)
+        self.assertAlmostEqual(xopt[1], -5)
+
+        # Function must be callable
+        self.assertRaisesRegexp(ValueError, 'callable', pints.fmin, 3, [1])
```
martinjrobins: curious as to why this is here. Does it really just test if the function is a function (i.e. can be called)?

MichaelClerx: It's a test to see if the correct error is raised if `f` isn't callable! (It should raise a ValueError with some error message containing the word "callable").

MichaelClerx:
Thanks for having a look! Doing `curve_fit` now! (Was a bit early in creating the PR maybe...)

Issue #382, opened 2018-06-13
Replace Boundaries object by Boundaries interface and RectangularBoundaries object


MichaelClerx:
Current `Boundaries` are rectangular, but we might want more complicated things in the future, so I propose to split the current `Boundaries` object into two. This will mean updating a large number of examples.

Any thoughts @martinjrobins @ben18785 @sanmitraghosh @chonlei ? Any reasons I shouldn't?

MichaelClerx:
This could break a load of code, but we're still in alpha version, plus most people are using Pints for MCMC I think??

Issue #380, opened 2018-06-13
Add wrapper around boundaries based on a log-prior


MichaelClerx:
Give it a cut-off, so that everything below or equal to the cut-off (e.g. 0) will count as out of bounds.

My use case is wrapping a boundary around a custom log prior - uniform, but with a funny shape - without having to re-implement the funny shape.

PR #383, opened 2018-06-13
i382 Add more `Boundaries` types

MichaelClerx:
See #380 #382 

This PR:
- Splits the class `Boundaries` into an interface `Boundaries` and an implementation `RectangularBoundaries`
- Updates the optimisers so allow arbitrary boundary shapes (not just rectangular). This is a bit of extra code in each optimiser to keep track of which points we asked the user to evaluate (leaving out the out-of-bounds ones for speed)
- Adds a `LogPDFBoundaries` object that lets you use a LogPrior (with a user specified cut-off value) as boundaries for an optimisation
- Adds a ParabolicError toy error measure, to speed up unit tests for optimisers


MichaelClerx:
Having more complex boundaries creates some issues for optimisers, as we can't use the triangle transform! So this is not ready to merge yet...

MichaelClerx:
Closes #380
Closes #382

MichaelClerx:
Ready for review now @martinjrobins :-)

pints/_boundaries.py
```diff
@@ -13,6 +13,38 @@
...
+
+    def sample(self, n=1):
+        """
+        Returns ``n`` random samples from within the boundaries.
+
+        The returned value is a numpy array with shape ``(n, d)`` where ``n``
+        is the requested number of samples, and ``d`` is the dimension of the
+        parameter space these boundaries are defined on.
+
+        *Note that the :meth:`sample()` method does not always have to be
```
martinjrobins: might want to specify the cases in which the `sample` method *does have to be implemented

PR #396, opened 2018-07-18
Issue-395 ap model

### LINKED ISSUE #395: Add Beeler Reuter AP model to toy models
Discussed with @MichaelClerx, this will replace the HH AP model ticket #126, and the maximum conductance values are the only free parameters here. Both voltage and calcium transient are observable in this toy model. Similarly, solve with scipy.

- [x] Model
- [x] Test
- [x] Example page

chonlei:
- [x] Beeler Reuter AP toy model
- [x] Simple optimisation example notebook.
- [x] Tests for the toy model (100% cover)
- [x] Add to docs.

This PR addresses #395.


MichaelClerx:
Good job!

Looks like a good use case to have weighting on the different outputs. I think it now just ignores the calcium completely. Perhaps we can update the SumOfSquaresError to allow weighting in its constructor. Or maybe we want to go even further and set it automatically as max(experiment) - min(experiment) in every output?

Would need a separate PR for this that we can then merge before this one

chonlei:
@MichaelClerx 
> Good job!
Thanks!

> Looks like a good use case to have weighting on the different outputs. I think it now just ignores the calcium completely.
Yes, agree!

> Perhaps we can update the SumOfSquaresError to allow weighting in its constructor. Or maybe we want to go even further and set it automatically as max(experiment) - min(experiment) in every output?
I think updating the `SumOfSquaresError` to simply allow weighting the output sounds good.

> Would need a separate PR for this that we can then merge before this one
So should I do this `SumOfSquaresError` first and come back to this example/testing?

MichaelClerx:
> So should I do this SumOfSquaresError first and come back to this example/testing?

Yes please!

martinjrobins:
> Perhaps we can update the SumOfSquaresError to allow weighting in its constructor. Or maybe we want to go even further and set it automatically as max(experiment) - min(experiment) in every output?

I think by default the SumOfSquaresError should weight each output vector by its norm. And an option to manually specify weightings is great too!

MichaelClerx:
Rebased this!

MichaelClerx:
As mentioned on slack, please:

- [ ] Change the print-out of the original and new parameters to remove the log-scaling
- [ ] Add a figure of the new VS original conductances, like [figure 1 here]( http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1000914 )

chonlei:
Decided to just remove that last log-likelihood plot first here and opened a separate issue to do that. And we can just put that explanation in the first fitting example but not one of these toy model example -- this is a common issue across all our examples, so a bit out of scope here.

@MichaelClerx Can you have a final look at it?

PR #398, opened 2018-07-19
Issue-397 Add weighting

### LINKED ISSUE #397: Add weighting output option in SumOfSquaresError
As demonstrated in #396, in particular this toy model example: 
https://github.com/pints-team/pints/blob/cb004d40132d83486cf5a956c47653c58afaedfb/examples/toy-model-beeler-reuter-ap.ipynb

Model outputs in `MultiOutputProblem` can be in very different scales/order of magnitudes, which makes the optimisation ignores some of the model output.

@MichaelClerx also suggested to have `SumOfSquaresError` set it automatically as max(experiment) - min(experiment) in every output. 

(Note: not needed in log-likelihood if Gaussian noise is included, as that works as a scaling already.)

chonlei:
Added weighting to `SumOfSquaresError` and `MeanSquaredError`. Added tests.

One question: how should we change
`de = 2 * np.sum((r.T * dy.T), axis=(1, 2))`
in the `evaluateS1()`? Should this be something like:
`de = 2 * np.sum(((r * weights).T * dy.T), axis=(1, 2))`
where weights is the array of the output weights.

This PR closes #397. This should be done and merged before #396.


pints/_error_measures.py
```diff
@@ -214,21 +214,32 @@ class MeanSquaredError(ProblemErrorMeasure):
...
+        elif self._n_outputs != len(weights):
+            raise ValueError(
+                'Number of weights must match number of problem outputs.')
+        # Check weights
+        self._weights = np.asarray([float(w) for w in weights])
+
     def __call__(self, x):
-        return (np.sum((self._problem.evaluate(x) - self._values)**2) *
-                self._ninv)
+        return (np.sum((self._problem.evaluate(x) - self._values)**2 *
```
MichaelClerx: Could you rewrite this with 2 sums to make it a bit faster?

The procedure should become:

1. Evaluate and subtract values (2d)
2. Sum along the time axis
3. Multiply by weights
4. Sum along the output axis
5. Multiply by n_inv and return

MichaelClerx: Actually, could you do some benchmarking and see if this is necessary?
If so, we'll need to do it below as well

chonlei: Just did a little test, it is necessary, so I will apply this change to everywhere whenever possible.

chonlei:
@MichaelClerx It should be ready now.

pints/tests/test_error_measures.py
```diff
@@ -243,6 +261,24 @@ def test_sum_of_squares_error(self):
...
+        # Residuals are: [[0, 0], [-1, -1], [-2, -2]]
+        # Error is (1+4)+(1+4)*2=15
+        self.assertEqual(e(x), 15)
+        y, dy = e.evaluateS1(x)
+        self.assertEqual(y, 15)
+        # derivatives are [[[1, 1], [1, 1]], [[1, 1], [1, 1]], [[1, 1], [1,1]]]
+        # dy1 is: 2 * [(0*1+-1*1+-2*1)*1 + (0*1+-1*1+-2*1)*2]
+        #       = 2 * [-3*1 + -3*2]
+        #       = -18
+        # dy1 is: 2 * [(0*1+-1*1+-2*1)*1 + (0*1+-1*1+-2*1)*2]
```
MichaelClerx: This is dy1 twice. Where have the weights gone?

pints/tests/test_error_measures.py
```diff
@@ -243,6 +261,24 @@ def test_sum_of_squares_error(self):
...
+        #       = 2 * [-3*1 + -3*2]
+        #       = -18
+        self.assertEqual(dy[0], -18)
+        self.assertEqual(dy[1], -18)
```
MichaelClerx: Shouldn't these two be different because of the weight?

pints/tests/test_error_measures.py
```diff
@@ -157,6 +157,24 @@ def test_mean_squared_error(self):
...
+        # Residuals are: [[0, 0], [-1, -1], [-2, -2]]
+        # Error is (1+4)+(1+4)*2=15, divided by 6
+        self.assertEqual(e(x), 15 / 6)
+        y, dy = e.evaluateS1(x)
+        self.assertEqual(y, 15 / 6)
+        # derivatives are [[[1, 1], [1, 1]], [[1, 1], [1, 1]], [[1, 1], [1,1]]]
+        # dy1 is: 2 * [(0*1+-1*1+-2*1)*1 + (0*1+-1*1+-2*1)*2]
+        #       = 2 * [-3*1 + -3*2]
+        #       = -18
+        # dy1 is: 2 * [(0*1+-1*1+-2*1)*1 + (0*1+-1*1+-2*1)*2]
```
MichaelClerx: dy1 twice

pints/tests/test_error_measures.py
```diff
@@ -157,6 +157,24 @@ def test_mean_squared_error(self):
...
+        #       = 2 * [-3*1 + -3*2]
+        #       = -18
+        self.assertEqual(dy[0], -18 / 6)
+        self.assertEqual(dy[1], -18 / 6)
```
MichaelClerx: No difference due to weights?

pints/tests/test_error_measures.py
MichaelClerx: ok got it

pints/tests/test_error_measures.py
chonlei: Just for the record, they are the same because of the constant model. The weights and values will be summed in the same way in the two parameters, and the only thing that matters here is the derivatives of the parameters -- which are the same here (constant model), so no difference.

chonlei:
@MichaelClerx Unfortunately, there is no other toy model that is as easy as this constant model -- all others are lovely and 'difficult' models. Though I will fix the comment typo.

MichaelClerx:
I have updated the ConstantModel so that the derivatives in the different outputs are different! (The first output has derivative 1, the second 2, etc.). Also fixed a big bug in the constant model's derivatives, so should rebase this once #413 has been merged and then update these tests!

MichaelClerx:
I have rebased this!

Issue #107, opened 2017-12-03
Add nested sampling to Pints


ben18785:
Rejection sampling, ellipsoidal sampling and, eventually, multiNest.

MichaelClerx:
It needs tests!

ben18785:
I agree. Adding tests for the various functions and samplers is next on my
agenda.

MichaelClerx:
In general though, no new feature should be merged that

1. Doesn't have tests
2. Doesn't show up in the docs at http://pints.readthedocs.io/

ben18785:
How do I get it to show up in the docs?

MichaelClerx:
There's a directory called `docs` that has a bunch of `.rst` ([reStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickstart.html)) files. It's structured into directories like `optimisation` and `mcmc`. The thing to do would be to find the suitable directory, copy a file describing e.g. adaptive covariance mcmc, and copy and modify it. Then update the `index.rst` in the same directory and it should work!

To test it locally, open a console and navigate to the `docs` directory. Then type `make html` and it should build a website without errors. You can then check in your browser to see if the stuff in `docs/html` makes sense.

After you've pushed and everything has been merged into master, readthedocs.io should automatically pick up on it and rebuild our docs. (This last bit is broken a.t.m., see #175)

MichaelClerx:
Sorry we should really set up some kind of docs for contributing to this project!

MichaelClerx:
Probably won't fit in ask/tell, might be useful to be able to do one step at a time?

MichaelClerx:
General outline:

- Sample from the prior
- While:

  - Calculate likelihood for those points
  - Select points with likelihood greater than some threshold

MichaelClerx:
Here's a list of things we still need to do before we can close this ticket:

- [ ] Add unit tests
- [ ] Implement the `Loggable` interface instead of using `print()`
- [ ] Check if the notebooks are as fast as they can be while staying informative #199 
- [ ] Make sure we 'hide' all functions and methods a user doesn't need to see #174
- [ ] Finish the nested sampling example? #130
- [ ] See if there's any way to make it ask-and-tell-ish (probably not?)
- [ ] See if we can optimise / parallelise anything

MichaelClerx:
Have made a start at this in branch: https://github.com/pints-team/pints/tree/i107-nested

MichaelClerx:
[...]

PR #402, opened 2018-07-23
i107 Nested samplers: tests & refactoring

MichaelClerx:
This PR addresses part of the issues in #107 by refactoring the nested sampler classes and adding (very simple) unit tests, bringing the coverage of these classes from 0 to 100%.


MichaelClerx:
Have successfully re-run the notebook!
It _can_ get stuck sometimes, we should probably think about that :-)

@ben18785 is it ok for me to merge this now?

Also, there's still a few issues to deal with after this regarding nested sampling. See https://github.com/pints-team/pints/issues/107

PR #410, opened 2018-07-24
Is SMC working?

ben18785:
I want to see whether people think (a) I have coded up the algorithm correctly (algorithm 3.1.1. in the attached) and (b) if so, why doesn't it seem to work. As such, I've created a pull request. This isn't to be seen as an attempt to merge with the master. It's just so people can easily access what I am talking about.

There are a couple of choices that one has to make along the way (as SMC offers choices!). I have highlighted these in the Pints code. The first is what form of intermediate distribution to use. The second is whether to resample the particles at the end of a given iteration. Both seem to affect whether the algorithm works, and so feedback would be good.

Within the PR there are two notebooks: one that uses a version of SMC that I have coded up in Pints (SMC-test); another which has a simple univariate example coded up all within a single notebook (SMC--working).


sanmitraghosh:
I had a quick look just now. I think its overall correct especially w_tilde() which is the crucial bit.
I just have a few quick feedback.

1) Take power posteriors intermediate distributions, \pi_n---likelihood^temperature_n \times prior, temperature goes from 0---to---1, e.g. for n=(1,...,N) we can have temperature = np.linspace(0,1,N)**5. This schedule is good for ODEs. You can extract the log likelihood and prior from the log_posterior object in pints. I have done this in the power_posterior branch (see __init__py in mcmc folder). In this case for n=1, the temperature = 0. Thus the first importance distribution is eta_1=prior. So you start by sampling from prior. This is widely used in particle filters and works well for ODE/SDEs in my opinion.

2) I think you will do this anyway. Use adaptiveCovarianceMCMC for k(xn,xn-1). This is theoretically valid. In fact we can use any good adaptive MCMC including gradient ones.

I will review it more carefully in a few days.

sanmitraghosh:
Some other points. We need to add Kitagawa resampling to stabilise ESS. I'll do that in my ABC-SMC and will add that here later, so don't bother. I'll also add an parallel evaluator for new_weight(). @MichaelClerx @ben18785  when you write tests please record the ESS for a given test problem. @ben18785 see if you can add the logZ calculation.

ben18785:
Hi Sanmitra,

That's great, thanks.

1. Yes, I have already been doing this, sort of. I was using a somewhat
arbitrary multivariate normal (both should work). Perhaps the prior is a
more reasonable choice.

2. Yes, that's the idea, eventually. We could even (much much more
eventually) add HMC for the proposals.

I'd really appreciate it if you/everyone else could look through the code,
and try it out on a few problems. As I said in the original message, the
code I have *should* (I think) work. But it seems not to. We need to get to
the bottom of this before we start adding bells and whistles (by that I
mean 2.).

Best,

Ben

sanmitraghosh:
I'll try it on my ion channel problem.

MichaelClerx:
> You can extract the log likelihood and prior from the log_posterior object in pints

Keep in mind that we want most samplers to work with any logpdf, not just bayesian stuff!

ben18785:
Great point Michael. I think what we'll do is something like: for Bayesian
models, we use the prior as a mixing dist; for other cases, we use a
multivariate normal. I am also in favour of allowing simulated annealing as
an option to the user where we just use p(x)^beta, with no mixing component.

MichaelClerx:
What's the state of this? Is it working in the notebook?

Issue #411, opened 2018-07-26
Derivatives in constant model are wrong


MichaelClerx:
For each point in time, it should return an array

```
[[df1/dp1, df1/dp2],
 [df2/dp1, df2/dp2]]
```
i.e.
```
[[df1/dp1, 0],
 [0, df2/dp2]]
```
it currently doesn't have zeros on the non-diagonals

PR #413, opened 2018-07-26
Fixing bug in ConstantModel derivatives

MichaelClerx:
See #411 

This PR does two things:

1. It fixes a bug in the ConstantModel derivatives output, see #411 (and improves testing)
2. It changes the ConstantModel so that the first output is 1*p1, the second output is 2*p2, etc. This way the model is still constant in time, linear in parameters, but now we get a bit more variety in the derivatives of the output. This makes it a better model for testing.


chonlei:
Although we could use `\delta_{i,j}` to replace `\\begin{cases} i, i = j\\\\0, i \\neq j \end{cases}`!

MichaelClerx:
> Although we could use \delta_{i,j} to replace \\begin{cases} i, i = j\\\\0, i \\neq j \end{cases}!

Bit too physics-ey :-)

PR #465, opened 2018-08-28
Sequential Monte Carlo - old PR

ben18785:
This it is now working. Not sure about the way I call the sampler though. At the moment, you can optionally add a log_prior otherwise a multivariate normal (of the same dimension as the problem) is used as a mixing distribution.

The notebook in there has two SMC examples: one with a multivariate normal pdf and another with the logistic model. I can separate these into two different notebooks if you'd like? Think we need both examples though as they illustrate how difficult it is to tune this thing.

To-do:

_Removed old to-do list_


MichaelClerx:
Hi @ben18785 ! This branch has diverged a bit from master, as it's been a while and lots of things have changed. Do you mind if I rebase it so that all the coverage stuff works again?
It will mean deleting your local copy and then checking it out anew, once I'm done

ben18785:
Yep, sounds good. Go for it!

MichaelClerx:
So things left to do for SMC are:

1. Getting the diff coverage up to 100%
2. Splitting up the notebook
3. Adding rst files for documentation

?

MichaelClerx:
Uhm, why does it take a `log_posterior` and a `log_prior` as input arguments? Shouldn't the `log_prior` be the one inside the `log_posterior` ?

ben18785:
@MichaelClerx Yep, I wasn't sure on this implementation. Basically, there are two circumstances:

- Bayesian problem: log_prior is within log_posterior and so no need to pass it (I didn't know how to extract a log_prior from a log_posterior though so didn't do this).
- log_pdf problem when there is no prior (e.g. toy normal problem). Here I use a multivariate normal of same dimension as the problem with (x0, sigma0) as its mean-variance.

Ideally, this would be handled without having to pass an optional log_prior because - as you say - this should be getable from log_posterior.

ben18785:
@MichaelClerx 

I like option 1 to be honest as think it's neater.

Only point on 1. is in mcmc we don't generally (unless populationMCMC -- haven't changed this yet to be same as SMC) need the log_posterior to be split.

ben18785:
Yep, definitely think that option 1 is optimal here. I have just been thinking about populationMCMC and we will definitely want to know if we can decompose the log_pdf into a prior + a likelihood.

ben18785:
Before I add any tests, it'd be good @martinjrobins @MichaelClerx to get your feedback on this reformulation of SMC into the ask/tell framework. It uses adaptive covariance across all the samples (pooling across different temperatures).

At the moment, the algorithm is not taking advantage of parallelism since ask-evaluate-tell is done sequentially across all particles. A way to improve this would be to split all the particles up into a number of groups, and run adaptive covariance on each of these. I may do this as a separate branch though.

MichaelClerx:
Have made a separate issue for splitting logposteriors into logpriors and loglikelihoods #568

MichaelClerx:
Just rebased this!

MichaelClerx:
Hi @ben18785 ,

I've rewritten this as ask/tell now, and the first example seems to work fine, but the second does not.

I've searched for differences, but the only thing I can find is that the old version used a single chain for every point, while the new version uses a chain per particle. In the old code the single chain's current (x, logpdf(x)) pair was updated before every ask and tell, so it acted a bit like multiple chains. However, this version will have adapted n_particles times more often than the current implementation. Do you think that could cause the newly observed behaviour in notebook 2?

MichaelClerx:
Another question: We've got this classified as "particle based" on this page: https://pints.readthedocs.io/en/latest/#sampling

But in the code it's under `sequential`. Should we change that to `particle_based_sampling` or something? Are there other particle-based samples that go in the same class, or just SMC variants?

MichaelClerx:
I wasn't sure where to put a lot of the methods (set_n_particles, set_temperature_schedule) because I'm not sure if they go in (A) the SMCSampler abstract class - so shared functionality among all SMCSamplers, or (B) the individual SMC class - our only implementation of an SMCSampler

This touches on the point above as well; is the abstract class here meant to be an `SMCSampler` or some kind of `ParticleBasedSampler` ?

MichaelClerx:
One issue with using multiple chains is that they only do 9 steps; that means the adaptation doesn't have time to "fade out", i.e. we're still adapting in the final steps. But I guess since we're not storing chains anyway that doesn't make the method invalid?

ben18785:
@MichaelClerx I've started to look at this, sorry for the delay.

Just trying to understand, what does one round of ask and tell do for this method? Do they update positions of all particles? I was wondering whether the conditional resampling was happening too often (it should happen once per temperature) but was failing in my understanding...

Yes, I think we want to use a single chain because of adaptation issues (we may not be doing anywhere near enough adapting as you say using the new method) and so reckon this is likely the cause of the issue. It also makes sense to use one chain so we can learn as much as quickly as possible.

Re: naming, I would keep the current way of naming things as particle samplers can be confused with other things (a later class of things that we want to put into PINTS).

Re: abstract vs base class for stuff, I don't really know to be honest. I think we will make additions to this class but don't know how similar/different they will be from current.

Re: all of this difficultly, this is why the importance-sampler-MCMC paper I proposed is so much better -- it's so much easier to code up and make choices!

ben18785:
@MichaelClerx Thinking about it more this lack of adaptation is almost certainly the cause of the issues since the posterior width for the first parameter of the logistic is so tight relative to the default step size.

MichaelClerx:
Thanks @ben18785 

> Just trying to understand, what does one round of ask and tell do for this method? Do they update positions of all particles? I was wondering whether the conditional resampling was happening too often (it should happen once per temperature) but was failing in my understanding...

Each ask/tell does one round of MCMC sampling (so an MCMC step for every chain).
If this happens to be the last of a series of kernel steps, it also does the other stuff (e.g. resampling).
The crucial bit for that is here: https://github.com/pints-team/pints/blob/i120-SMC/pints/_sequential/_SMC.py#L179-L182

I'll change it back to a single chain then, and we'll see how that goes :-)

MichaelClerx:
Starting a new PR for this

PR #466, opened 2018-08-28
Dream hyperparameters now modifiable

ben18785:
See #466


pints/_mcmc/_dream.py
```diff
@@ -349,11 +342,73 @@ def set_constant_crossover(self, enabled):
         """
         self._constant_crossover = True if enabled else False
 
+    def set_b_start(self, b_star):
```
martinjrobins: should be `set_b_star`?

pints/_mcmc/_dream.py
```diff
@@ -349,11 +342,73 @@ def set_constant_crossover(self, enabled):
...
+        """
+        Sets b* which determines the weight given to
+        other chains' positions in determining new positions
```
martinjrobins: should mention in docstring that b* is required to be positive.

pints/_mcmc/_dream.py
```diff
@@ -349,11 +342,73 @@ def set_constant_crossover(self, enabled):
...
+        """
+        if b_star < 0:
+          raiseValueError('b* must be non-negative.')
+        self._b_star = b_star
+
+    def set_p_g(self, p_g):
+        """
+        Sets p_g which is the probability of
```
martinjrobins: again mention must be non-negative and <1 in docstring

pints/_mcmc/_dream.py
```diff
@@ -349,11 +342,73 @@ def set_constant_crossover(self, enabled):
...
         """
         self.set_b(x[0])
+        self.set_b_star(x[1])
+        self.set_p_g(x[2])
+        self.set_delta_max(x[3])
+        self.set_initial_phase(x[4])
+        self.set_constant_crossover(x[5])
+        self.set_CR(x[6])
+        self.set_nCR(x[7])
+        self.set_initial_phase(x[4])
```
martinjrobins: `set_initial_phase` is done twice here

ben18785: Thanks sorted.

MichaelClerx:
@martinjrobins Some hyper parameters are incompatible: e.g. hyper param X has no effect if Y=0.

martinjrobins:
@MichaelClerx. Having a discrete hyper-param means that I'll have to change some things. And it certainly would make the optimisation problem more difficult, but then that is the point of the automatic hyperparameter tuning for the matrix, i.e. it will penalise methods that are difficult to tune.

Issue #173, opened 2018-01-15
Add tests for diagnostics methods


MichaelClerx:
We should add a few tests for the diagnostics.

They should mostly test if the methods can be called with the different types of arguments we expect, and raise errors etc. as we expect.

ben18785:
Addressed in #467.

PR #467, opened 2018-08-28
Diagnostics tests now implemented

ben18785:
See #173


MichaelClerx:
@ben18785 in general we wouldn't add tests for private methods, (i.e. methods that start with an underscore). The idea is that these should be used by public methods, so testing the public methods should reach every private method (if not, those private methods may be obsolete).

Issue #206, opened 2018-01-31
Diagnostics `rhat`, `rhat-params`, and `effective_sample_size` need docs mention


MichaelClerx:
Need to:

- [x] Add them to the sphinx docs (docs/sources/diagnostics.rst should be created)
- [x] Double-check docstrings
- [x] Do a local build and check if it worked before uploading (`cd docs`, `make clean`, `make html`, check results in `docs/build/html` in browser)

ben18785:
Addressed in #468.

PR #468, opened 2018-08-29
Docs and pruned docstrings for diagnostics

ben18785:
See #206


ben18785:
@MichaelClerx Any idea why this is failing? I have a diagnostics.doctree file after I build it. It works on all versions of Python but 3.6...

MichaelClerx:
Hi @ben18785 ! That particular test is the one that builds the documentation. It failed with a warning that the `diagnostic.rst` file itself wasn't included in any "toctree", i.e. that there was no document linking to it. Have fixed that now!

PR #480, opened 2018-08-31
Multinest implemented in new 'tutorials' folder notebook

ben18785:
Multinest isn't in Pints yet but its bones are sort of there. I've created a new folder called 'tutorials which I do think is something well worth us doing. I did it here because a) this algorithm is pretty complicated and I think it would help users to understand how it works under the hood and b) in case we/I want to develop it further (as I think needs to be done -- see below) this will really help speed up development by helping recall the details.

The reason for this PR is twofold: a) to get your thoughts on these proposed 'tutorial' notebooks (I'm not suggesting these are a priority, merely that I think they are useful for us/users) and b) to encourage others to see if they can improve on my implementation. At the moment, the algorithm seems to do a reasonable job of finding a collection of bounding ellipsoids but I am not convinced that I have done a great job because it doesn't appear as efficient as in the paper.

Even so, the algorithm as it stands could relatively easily be put in Pints as the other parts of it are essentially the same as ellipsoidal sampling.


martinjrobins:
So are you saying that the `tutorials` subdir would contain extended notebooks describing different methods in more detail, i.e. how they work. This seems like a good idea, but I'm don't like about putting an work-in-progress method here until it is completed, fully tested and we are all happy with it.

ben18785:
Yep, totally. I wasn’t suggesting merging this into master. I just put it there for now as thought that - eventually (when the method is done) - this would be a good place to explicate it.

MichaelClerx:
Looks like a lot of work went into this! I don't mind having these notebooks with work in progress, maybe they could just go with `studies`? It seems like a great way to discuss implementations etc. with people while we're building them.

MichaelClerx:
I'll leave commenting on the actual sampling to @sanmitraghosh :D

PR #481, opened 2018-08-31
Ellipsoidal now running >10x faster for the logistic example

ben18785:
(with 4x the number of active points and so really ~ 30x faster)!

The method for calculating the bounding ellipse is, however, a bit more approximate. So what I wonder is -- is it worth having both methods in there that a user can choose from? My reasons for doing so are that, for some problems, the improvement in accuracy of the bounding box may be necessary.


MichaelClerx:
Sounds like a great improvement!

(Implementation wise, If you want to make the new method optional, you could either have an `if` somewhere that decides which method, but you could also have a shared reference to a function that you change, so self._f = self._fast_function or self._f = self._slow_function, and then just call self._f everywhere without needing another if, whatever's easiest!)

MichaelClerx:
Then again, it doubles the amount of testing and maintenance, so only add the option if you know it's necessary!

martinjrobins:
How would selecting between the two methods be for the user?
There seems to me to be two options:
- introducing a hyper-parameter to swap between the two methods
- Having the two methods as two separate classes, sharing a base class for the code common to both.

The latter might be more appropriate if the methods have differing numerical properties.

ben18785:
@MichaelClerx Yep, can we merge this and then create a ticket for me to add an optional fast vs accurate (defaulting to fast, I think) method selection for the user. @martinjrobins I'd prefer to keep these as a single method as it's a bit clunky to have nested and nested_fast etc.

MichaelClerx:
Ok with you @martinjrobins ?

ben18785:
Now addressed in a separate PR.

PR #491, opened 2018-09-01
Substantial change to adaptive MCMC: new class structure

ben18785:
As per #279 I have been investigating other types of adaptive MCMC samplers other than Remi's. Much like the other MCMC algorithms, 'adaptive MCMC' isn't one algorithm but a family of them which have similar structure but substantially different details.

Due to the common structure of these algorithms, I thought it was best to create an abstract AdaptiveCovarianceMCMC class, which implements methods etc. common to each method and then have subclasses for each of the individual algorithms.

I've implemented quite a few different algorithms already in this framework so can testify that it makes development really quite easy. I also think the code itself is much cleaner and more readable this way. At the moment we have working in this branch:

- Remi's AM sampler
- An AM sampler
- An AM local-componentwise sampler
- An AM sampler with global adaptation
- A Rao-Blackwell sampler

Here 'AM' stands for adaptive Metropolis. In development we have:

- An AM global-componentwise sampler. This needs some help from @MichaelClerx @martinjrobins since I can't figure out how to get it into ask/tell.
- A localised AM sampler. This is quite a complex beast but am getting there with it.

At the moment, I haven't flake8'd it or written tests because I wanted to get feedback on the framework I've chosen before I do so. Also, there's definitely scope to shorten names of the classes. I'd also like to make the way we call these methods using something like

pints.MCMCSampling(log_posterior, 3, xs, method=pints.AdaptiveCovarianceMCMC.Global)

rather than the clunkier and verbose current,

pints.MCMCSampling(log_posterior, 3, xs, method=pints.AdaptiveCovarianceAMGlobalAdaptiveMCMC)

Also, I'm aware that some of the other methods like populationMCMC (and potentially SMC) use AdaptiveMCMC as part of their algorithms. So any changes to the framework we make should still mean we can do this easily.


ben18785:
Also, there are a host of adaptive MCMC algorithms that are discussed in [this](https://amstat.tandfonline.com/doi/abs/10.1198/jcgs.2009.06134#.W4ssvtJKhPb) article by Roberts and Rosenthal. These shouldn't be too taxing to code up and include:

- Adaptive Metropolis within Gibbs
- Single Component Adaptive Metropolis (SCAM; although hopefully only by name)
- State-dependent scaling
- Regional adaptive Metropolis

The paper also discusses when it may be beneficial to sample logs of parameters rather than the parameters themselves (turns out to be when using log-tailed priors).

ben18785:
Note to self: need to add set_hyper_parameter methods for the various methods.

ben18785:
Localised adaptive may be working now, but nonetheless not being able to sample from the multimodal posterior. Next steps to consider are (as per the paper):
- use Student-t mixture and let nu->infty as i->infty. The paper has all the updating equations necessary here.
- Add a GMM fitting 1st stage after pre-adaptation iterations. This could be done by keeping track of all the samples within the algorithm.

MichaelClerx:
Hi Ben!
Why are there SMC classes in this branch? Are they duplicates of another branch?

MichaelClerx:
W.r.t. to naming: Shall we call it ACMCMC ? Then we can have names like RemiACMCMC, or perhaps just ACMC and RemiACMC ?

pints/_mcmc/__init__.py
```diff
@@ -423,13 +423,32 @@ def run(self):
...
             # Calculate scores
             fxs = evaluator.evaluate(xs)
-
+            
+            a_none = False
             # Perform iteration(s)
             if self._single_chain:
                 samples = np.array([
                     s.tell(fxs[i]) for i, s in enumerate(self._samplers)])
+                for i, sample in enumerate(samples):
```
MichaelClerx: This breaks parallelism

pints/_mcmc/__init__.py
```diff
@@ -423,13 +423,32 @@ def run(self):
...
+                        a_none = True
+                        while sample is None:
+                            xss = [np.array(self._samplers[i].ask())]
+                            fxss = evaluator.evaluate(xss)
+                            sample = self._samplers[i].tell(fxss[0])
+                    samples[i] = sample
             else:
                 samples = self._samplers[0].tell(fxs)
+            
+            # convert lists to numpy array in case of None
```
MichaelClerx: What's going on here?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -150,15 +143,28 @@ def tell(self, fx):
...
+            self._alpha = 0
+
             # Return first point for chain
             return self._current
 
         # Check if the proposed point can be accepted
-        accepted = 0
+        self._accepted = 0
+        r = fx - self._current_log_pdf
+        self._X = self._current
```
MichaelClerx: What are `self._X` and `self._Y` and where are they used?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -150,15 +143,28 @@ def tell(self, fx):
...
 
+            # Set alpha prob to zero
+            self._alpha = 0
+
             # Return first point for chain
             return self._current
 
         # Check if the proposed point can be accepted
-        accepted = 0
+        self._accepted = 0
```
MichaelClerx: Why is this stored now?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -168,25 +174,42 @@ def tell(self, fx):
...
 
-            # Update mu, log acceptance rate, and covariance matrix
-            self._mu = (1 - gamma) * self._mu + gamma * self._current
-            self._loga += gamma * (accepted - self._target_acceptance)
-            dsigm = np.reshape(self._current - self._mu, (self._dimension, 1))
-            self._sigma = (
-                (1 - gamma) * self._sigma + gamma * np.dot(dsigm, dsigm.T))
+            # For localised AM
+            if self._localised:
+                self._fit_gaussian_mixture()
```
MichaelClerx: This method should be defined in this base class. The base class shouldn't know anything about it's children!

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -168,25 +174,42 @@ def tell(self, fx):
...
+                self._initial_fit = False
```
MichaelClerx: Needs a comment to explain what this is. And should be defined in `__init__` if it's used...

pints/_mcmc/_adaptive_covariance.py
MichaelClerx: Perhaps should have a name like `update_...` to match with the others? In non-local mode we update mu and sigma, but in local mode we fit a gaussian mixture?

pints/_mcmc/_adaptive_covariance_am.py
```diff
@@ -0,0 +1,68 @@
...
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class AdaptiveCovarianceAMMCMC(pints.AdaptiveCovarianceMCMC):
+    """
+    Adaptive Metropolis MCMC, as described by Algorithm 2 in [1],
+    (with gamma = self._adaptations ** -eta which isn't specified
```
MichaelClerx: You can just call this variable `adaptations` in the comments. The fact that it's a hidden variable (`_`) stored in the object (`self.`) isn't relevant to the user reading this

pints/_mcmc/_adaptive_covariance_am.py
```diff
@@ -0,0 +1,68 @@
...
+#  For licensing information, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class AdaptiveCovarianceAMMCMC(pints.AdaptiveCovarianceMCMC):
```
MichaelClerx: Can we call this something like AdaptiveMetropolisACMC ?

ben18785:
I’m happy with that. Good idea.

pints/_mcmc/_adaptive_covariance_am.py
```diff
@@ -0,0 +1,68 @@
...
+        See :meth: `AdaptiveCovarianceMCMC._initialise()`.
+        """
+        super(AdaptiveCovarianceAMMCMC, self)._initialise()
+        
+        # log adaptation
+        self._lambda = (2.38**2) / self._dimension
+
+    def tell(self, fx):
+        """ See :meth:`pints.AdaptiveCovarianceMCMC.tell()`. """
+        super(AdaptiveCovarianceAMMCMC, self).tell(fx)
```
MichaelClerx: Why not just `return super(A...)` ? Now we're discarding the return value and then using knowledge of the parent class. Don't like it!

pints/_mcmc/_adaptive_covariance_am.py
```diff
@@ -0,0 +1,68 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -14,18 +14,11 @@
...
-    MCMC steps converges to a user specified value.
-
-    [1] Uncertainty and variability in models of the cardiac action potential:
-    Can we build trustworthy models?
-    Johnstone, Chang, Bardenet, de Boer, Gavaghan, Pathmanathan, Clayton,
-    Mirams (2015) Journal of Molecular and Cellular Cardiology
-
-    [2] An adaptive Metropolis algorithm
-    Heikki Haario, Eero Saksman, and Johanna Tamminen (2001) Bernoulli
+    In all cases ``self._adaptations ^ -eta`` is used to control decay of
```
MichaelClerx: Don't call things `self._` in the documentation. That's an implementation detail!

MichaelClerx: Could we have a bit more explanation here of what an adaptive covariance mcmc method is?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -92,6 +70,8 @@ def _initialise(self):
         # Adaptation
         self._loga = 0
         self._adaptations = 2
+        self._gamma = 1
```
MichaelClerx: These (and others) should all be defined in `__init__`

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -150,15 +141,28 @@ def tell(self, fx):
             # Clear proposal
             self._proposed = None
 
+            # Set alpha probability to zero
+            self._alpha = 0
```
MichaelClerx: What is this and where is it used?

pints/_mcmc/_emcee_hammer.py
```diff
@@ -0,0 +1,175 @@
...
+        # Clear proposal
+        self._proposed = None
+
+        # Return samples to add to chains
+        self._current.setflags(write=False)
+        return self._current
+
+    def set_a(self, a):
+        """
+        Sets the coefficient ``a`` used in updating the position of each
```
MichaelClerx: Please stick to a line length of 79 characters for word-wrap

pints/_mcmc/_emcee_hammer.py
```diff
@@ -0,0 +1,175 @@
+#
+# Emcee hammer MCMC
```
MichaelClerx: What's this doing here? Shouldn't this have a branch of its own?

pints/_mcmc/_dram.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        Calculates alpha expression necessary in eq. 3 of Haario et al. for
+        determining accept/reject
+        """
+        alpha_log = log_Y[n + 1] - log_Y[0]
+        if n == 0:
+            return alpha_log
+        Y_rev = Y[::-1]
+        log_Y_rev = log_Y[::-1]
+        for i in range(n):
```
MichaelClerx: Can we rewrite this as a (faster) matrix statement?

pints/_mcmc/_dram.py
```diff
@@ -0,0 +1,241 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance_remi.py
```diff
@@ -0,0 +1,69 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance_rao_blackwell.py
```diff
@@ -0,0 +1,83 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance_rao_blackwell.py
```diff
@@ -0,0 +1,83 @@
...
+    def _initialise(self):
+        """
+        See :meth: `AdaptiveCovarianceMCMC._initialise()`.
+        """
+        super(AdaptiveCovarianceRaoBlackWellMCMC, self)._initialise()
+        
+        # log adaptation
+        self._lambda = (2.38**2) / self._dimension
+
+    def tell(self, fx):
```
MichaelClerx: Can we just delete this? It seems like we want the base class version of tell to be called?

pints/_mcmc/_adaptive_covariance_remi.py
```diff
@@ -0,0 +1,69 @@
...
+        
+        # log adaptation
+        self._loga = 0
+
+    def tell(self, fx):
+        """ See :meth:`pints.AdaptiveCovarianceMCMC.tell()`. """
+        super(AdaptiveCovarianceRemiMCMC, self).tell(fx)
+        # Update log acceptance
+        if self._adaptive:
+            self._loga += self._gamma * (self._accepted - self._target_acceptance)
```
MichaelClerx: If this is the only class using `_loga` it should be defined here and not in the base class

MichaelClerx: And where are the implementations of update_mu and update_sigma ?

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
...
+        Updates running acceptance probabilities according to,
+        alpha_t+1^k = alpha_t^k + gamma_t+1 * 1(Z_t+1==k?) * 
+                             (alpha_k(theta_t, Y_t+1) - alpha_t^k)
+        """
+        self._alpha_l[self._Z] += (self._gamma *
+                                     (self._alpha - self._alpha_l[self._Z]))
+
+    def _ratio_q(self):
+        """
+        Yields log q(Y_t+1|Z_t+1) - log q(theta_t|Z_t+1)
```
MichaelClerx: Returns, not yields! (`yield` is a different way of returning in Python, used to create iterators)

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
...
+    def _update_alpha(self):
+        """
+        Updates running acceptance probabilities according to,
+        alpha_t+1^k = alpha_t^k + gamma_t+1 * 1(Z_t+1==k?) * 
+                             (alpha_k(theta_t, Y_t+1) - alpha_t^k)
+        """
+        self._alpha_l[self._Z] += (self._gamma *
+                                     (self._alpha - self._alpha_l[self._Z]))
+
+    def _ratio_q(self):
```
MichaelClerx: Please change to something with a verb in it (e.g. calculate q ratio)

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
...
+                                                         self._mu[i], self._sigma[i],
+                                                         allow_singular=True))
+        self._log_q_l += -logsumexp(self._log_q_l)
+
+    def _update_mu(self):
+        """
+        Updates mu components according to,
+        mu_t+1^k = mu_t^k + gamma_t+1 * q_t^k * (theta_t+1 - mu_t^k)
+        """
+        for i in range(self._mixture_components):
```
MichaelClerx: Can we rewrite all these as matrix statements?

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
...
+        return self._current
+        
+    def _fit_gaussian_mixture(self):
+        """
+        Fits a Gaussian mixture distribution by updating
+        componentwise the means, covariance matrices,
+        weights and lamdas (eq. 36 and 37 in Andrieu &
+        Thoms 2008)
+        """
+        self._update_log_q()
```
MichaelClerx: Do we need these to all be in separate functions? Are they called in more than one place? If not, then shouldn't be functions!

MichaelClerx: Also thinking, perhaps we should merge update_mu and update_sigma into a single function, then perhaps we don't need the split into local/non-local in tell.
Would also be great if we could update the names of all these things to explain what they do in words, `update_arbitrary_symbol` isn't very descriptive (to me - perhaps I don't know the methods well enough?)

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        if mixture_components < 2:
+            raise ValueError('Number of mixture components ' +
+                             'should exceed 1.')
+        if not isinstance(mixture_components, int):
+            raise ValueError('Number of mixture components ' +
+                             'should be an integer.')
+        self._mixture_components = mixture_components
+
+    def tell(self, fx):
```
MichaelClerx: See comments on tell() method below

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
...
+        self._log_q_l = np.log(np.repeat(1.0 / self._mixture_components,
+                                         self._mixture_components))
+
+    def set_mixture_components(self, mixture_components):
+        """
+        Sets the number of Gaussian mixture components
+        to use for proposals
+        """
+        if mixture_components < 2:
+            raise ValueError('Number of mixture components ' +
```
MichaelClerx: You can do these as

```
raise ValueError(
    'Number of mixture components should exceed 1.')
```

pints/_mcmc/_adaptive_covariance_localised.py
```diff
@@ -0,0 +1,245 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance_am_global_component.py
```diff
@@ -0,0 +1,128 @@
...
+        super(AdaptiveCovarianceAMGlobalComponentMCMC, self)._initialise()
+        self._log_lambda_scalar = 0
+        self._log_lambda_vector = np.zeros(self._dimension)
+        self._am_global_rate = 10
+        self._iter_count = 0
+        self._Z = np.zeros(self._dimension)
+    
+    def set_am_global_rate(self, am_global_rate):
+        """
+        Sets number of steps between each global am update
```
MichaelClerx: What's `am` ?

pints/_mcmc/_adaptive_covariance_am_global_component.py
```diff
@@ -0,0 +1,128 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance_am_global_adaptive.py
```diff
@@ -0,0 +1,69 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/_adaptive_covariance_am_component.py
```diff
@@ -0,0 +1,113 @@
...
+        super(AdaptiveCovarianceAMComponentMCMC, self)._initialise()
+        self._log_lambda_scalar = 0
+        self._log_lambda_vector = np.zeros(self._dimension)
+        self._am_global_rate = 10
+        self._iter_count = 0
+        self._k = 0
+    
+    def set_am_global_rate(self, am_global_rate):
+        """
+        Sets number of steps between each global am update
```
MichaelClerx: What's `am`?

pints/_mcmc/_adaptive_covariance_am_component.py
```diff
@@ -0,0 +1,113 @@
+#
+# Adaptive covariance MCMC method
```
MichaelClerx: Please add a one-line description of this file.

pints/_mcmc/__init__.py
ben18785: @MichaelClerx Ok. Or I can just grab that code and use it here?

ben18785: @MichaelClerx I looked at your code and not sure it'll work for this example. Unlike HMC where, inbetween full updates, None is always returned, in DRAM, whether or not None is returned is stochastic. So, I think we have to break parallelism there because, whereas one chain may be None, another may not be. What do you think?

MichaelClerx:
Have fixed the mess I made of this branch with `git`. Please delete your local copy and re-checkout the branch before continuing work on this.

MichaelClerx:
@ben18785 can this be closed in favour or #924 ?

Issue #548, opened 2018-11-06
differential evolution: set_gamma_switch_rate throws exception when given a float


martinjrobins:
float is convertible to an int, so should just do that. This is necessary, for example, if you are using GPyOpt to tune the hyperparameters

after talking with @MichaelClerx, we've decided to do the conversion to an int in the `set_hyper_parameter` function and leave the throw in the `set_gamma_switch_rate` function

martinjrobins:
fixed and merged

PR #549, opened 2018-11-06
#548 allow diff evolution hyper-param interface to take anything conv…

martinjrobins:
…ertable to int for gamma_switch param

### DIFF
--- a/pints/_mcmc/_differential_evolution.py
+++ b/pints/_mcmc/_differential_evolution.py
@@ -270,7 +270,7 @@ class DifferentialEvolutionMCMC(pints.MultiChainMCMC):
         self.set_scale_coefficient(x[1])
         try:
             int_x2 = int(x[2])
-        except ValueError:
+        except (ValueError, TypeError):
             raise ValueError('The interval number of steps between ' +
                              'gamma=1 iterations must be convertable ' +
                              'to an integer.')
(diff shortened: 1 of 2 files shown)


pints/_mcmc/_differential_evolution.py
```diff
@@ -268,7 +268,13 @@ def set_hyper_parameters(self, x):
         """
         self.set_gamma(x[0])
         self.set_scale_coefficient(x[1])
-        self.set_gamma_switch_rate(x[2])
+        try:
+            int_x2 = int(x[2])
+        except ValueError:
```
MichaelClerx: I think it can be a TypeError too, so perhaps `except (ValueError, TypeError)`
(Alternative is not too catch at all, which is generally ok, but this is nicer of course!)

pints/tests/test_mcmc_differential_evolution.py
```diff
@@ -173,9 +173,8 @@ def test_set_hyper_parameters(self):
             ValueError, 'non-negative', mcmc.set_hyper_parameters,
             [1, -0.5, 20, 0, 0])
 
-        self.assertRaisesRegex(
-            ValueError, 'integer', mcmc.set_hyper_parameters,
-            [1, 0.5, 20.5, 0, 0])
+        mcmc.set_hyper_parameters([0.5, 0.6, 20.2, 0, 0])
+        self.assertEqual(mcmc._gamma_switch_rate, 20)
```
MichaelClerx: This should still be checked for the direct method.
In fact, don't really see why we're checking these methods via `set_hyper_parameters` instead of directly!

Issue #542, opened 2018-10-30
Add option to write MCMC chain to disk during run ("live" / dynamically)


MichaelClerx:
This can already be done via ask-and-tell (if you implement the disk-writing bit yourself), but it's a popular request esp. for very long runs that might unexpectedly terminate at some point.

Still wary of 'bloating' the MCMCSampling method, but coming round to the idea of this one...

mirams:
The posterior density values in there too would be nice!

MichaelClerx:
> The posterior density values in there too would be nice!

The LogPDF in Pints terms, we're not necessarily Bayesian ;-)

But yes

mirams:
Was thinking about implementing this myself as it is becoming a roadblock, but there seems to be a method `MCMCSampling.set_log_to_file(filename=None, csv=False)` already? What does this do?!

mirams:
Ah, just prints iteration number, function evaluations, acceptance ratios and not the actual chain locations and likelihoods?

MichaelClerx:
> Ah, just prints iteration number, function evaluations, acceptance ratios and not the actual chain locations and likelihoods?

Yup! The stuff you usually get on screen (possibly with extra info)

MichaelClerx:
What were you thinking for the output format? CSV? I'd like a header row in there. Something like `"iter","logpdf","p1","p2",...`? And multiple files (one per chain) or 1 file with multiple chains?

MichaelClerx:
- Example of creating multiple file names here: https://github.com/pints-team/pints/blob/master/pints/io.py#L78-L86
- Maximum precision float output: `pints.strfloat(number)`
- And I'm guessing you probably want to open and close the file any time you write a line, rather than keeping it open...

Actually... just use another instance of the Logger class!

mirams:
One file per chain will probably be easier to handle after? Not going to have chance to do this for a couple of weeks. But I will note that one of my chains crashed after 3 days with this, leaving me with nothing useful!:

       xs = [sampler.ask() for sampler in self._samplers]
      File "/usr/local/lib/python2.7/dist-packages/pints/_mcmc/_adaptive_covariance.py", line 68, in ask
        self._current, np.exp(self._loga) * self._sigma)
      File "mtrand.pyx", line 4511, in mtrand.RandomState.multivariate_normal (numpy/random/mtrand/mtrand.c:37052)
[...]

PR #552, opened 2018-11-12
Added option to log samples (but not log_pdfs) to disk during MCMCSampling

MichaelClerx:
See #542 

Current solution is easy and is compatible with `io.load_samples()`. Would be good to have `log_pdf` in the files too though, but would probably require changes to MCMCSampling ask-and-tell interface.

@ben18785 what do you think? Is it a common request to have the `log_pdf` values of the samples stored as well? Sounds sensible although I haven't seen any examples.


pints/_mcmc/__init__.py
```diff
@@ -651,6 +728,44 @@ def set_max_iterations(self, iterations=10000):
                     'Maximum number of iterations cannot be negative.')
         self._max_iterations = iterations
 
+    def set_output_files(self, chain_file, evaluation_file):
```
martinjrobins: Might be nicer to split this into two functions, `set_output_chain_file` and `set_output_evaluation_file`.

Reason being is that the name `set_output_files` is not very descriptive, does it refer to the logging perhaps?, and if you come across this in your code

```python
sampler.set_output_files(filename, None)
```

you have no idea what you are outputing....

MichaelClerx: Fair enough. I want to avoid confusion with the other _log_ methods though (which have nothing to do with this!)

martinjrobins: i was thinking that having the words `chain` and `evaluation` might further distinguish it from the logging functions...

MichaelClerx: Ok, have updated them now!

mirams: Not sure that `evaluation` is very clear, I'd be expecting something more like `target` or `target_value`?

MichaelClerx: Well we don't call it target anywhere else, plus the target is surely the distribution itself, not a single point within it? Happy to change it to say `set_logpdf_filename(...)`

MichaelClerx:
@mirams I've changed it to `set_log_pdf_filename` is that better for you?

mirams:
Yes! Thanks!

PR #564, opened 2018-11-16
Added docstring for the Pints module itself.

MichaelClerx:
Hi all,

Just noticed we didn't have a docstring for the `pints` module itself. I've added a short one (typicaly format is one-liner, followed by longer paragraphs with proper explanation).

Please check if you agree

### DIFF
--- a/pints/__init__.py
+++ b/pints/__init__.py
@@ -7,6 +7,15 @@
 #  For licensing information, see the LICENSE file distributed with the PINTS
 #  software package.
 #
+"""
+Pints: A Python Package for Picking Probable Parameters.
+
+This module provides several optimisation and sampling methods that can be
+applied to find the parameters of a model (typically a time series model) that
+are most likely given an experimental data set.
+"""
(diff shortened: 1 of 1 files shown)


pints/__init__.py
```diff
@@ -7,6 +7,15 @@
 #  For licensing information, see the LICENSE file distributed with the PINTS
 #  software package.
 #
+"""
+Pints: A Python Package for Picking Probable Parameters.
```
martinjrobins:
```suggestion
Pints: Probabilistic Inference on Noisy Time Series.
```

pints/__init__.py
```diff
@@ -7,6 +7,15 @@
...
+
+This module provides several optimisation and sampling methods that can be
+applied to find the parameters of a model (typically a time series model) that
+are most likely given an experimental data set.
```
martinjrobins:
```suggestion
are most likely, given an experimental data set.
```

PR #576, opened 2018-11-28
Added note on ask-and-tell.

MichaelClerx:
Hi all, just realised the docs didn't really say what an ask-and-tell interface is, so have added something in for the optimisers. Can you have a quick look and see if the wording makes sense?


mirams:
Maybe wants to state the obvious that `ask` means `ask which points to run a forward model at`, and `tell` means `tell the optimiser what the objective function evaluations were`.

martinjrobins:
Yea, agree with @mirams, otherwise looks good

MichaelClerx:
Just added a note @mirams and @martinjrobins !

PR #592, opened 2019-01-07
Added CITATION file

MichaelClerx:
https://www.software.ac.uk/blog/2016-10-06-encouraging-citation-software-introducing-citation-files


mirams:
Looks good to me

PR #600, opened 2019-01-22
I513 matplotlib 3

### LINKED ISSUE #513: Make Pints work with Matplotlib 3
Seems to cause some issues for our notebooks right now

chonlei:
Tiny fixes to make Pints works with Matplotlib 3.

Fix #513


pints/plot.py
```diff
@@ -698,16 +698,13 @@ def pairwise(samples,
...
                     # Force equal aspect ratio
-                    # See: https://stackoverflow.com/questions/7965743
-                    im = axes[i, j].get_images()
-                    ex = im[0].get_extent()
                     # Matplotlib raises a warning here (on 2.7 at least)
                     # We can't do anything about it, so no other option than
                     # to suppress it at this stage...
                     with warnings.catch_warnings():
                         warnings.simplefilter('ignore', UnicodeWarning)
                         axes[i, j].set_aspect(
```
MichaelClerx: Does this fit on one line now?

chonlei: Nope, it will be 81 col. (Unless we remove the `abs()`... it is probably not necessary.)

chonlei: OK. I will remove the `abs()`. I can't think of a way that it needs the `abs()`.

PR #623, opened 2019-01-29
Add heatmap for pairwise

### LINKED ISSUE #566: Add heatmap as an option for pairwise plot
@mirams suggests to bring back heatmap as an option for `pints.plot.pairwise`.

I cannot remember exactly why we moved away from it and use only scatter/kde plots, because I certainly remember I used to do heatmap instead of scatter/kde plots.

I think if there is no objection of adding it as an option, I will quickly do it with all other pints.plot tickets that I owe. @ben18785 @sanmitraghosh @MichaelClerx?

chonlei:
@mirams requested.

Close #566

### DIFF
--- a/pints/plot.py
+++ b/pints/plot.py
@@ -545,11 +545,12 @@ def series(samples, problem, ref_parameters=None, thinning=None):
 
 def pairwise(samples,
              kde=False,
+             heatmap=False,
              opacity=None,
              ref_parameters=None,
              n_percentiles=None):
     """
-    Takes a markov chain or list of `samples` and creates a set of pairwise
+    Takes a markov chain or list of ``samples`` and creates a set of pairwise
     scatterplots for all parameters (p1 versus p2, p1 versus p3, p2 versus p3,
     etc.).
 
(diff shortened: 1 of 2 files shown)


pints/plot.py
```diff
@@ -565,7 +566,10 @@ def pairwise(samples,
         is the number of parameters.
     ``kde``
         (Optional) Set to ``True`` to use kernel-density estimation for the
-        histograms and scatter plots.
+        histograms and scatter plots. Cannot use together with ``heatmap``.
+    ``heatmap``
+        (Optional) Set to ``True`` to plot heatmap for the pairwise plots.
+        Cannot use together with ``kde``.
     ``opacity``
```
MichaelClerx: Does opacity do anything for the heatmap?

chonlei: No, it should only affect the scatter plot -- neither kde or heatmap uses it.

mirams:
Got an example of it?

chonlei:
@mirams Not in the repo, but if you run `pints.plot.pairwise(samples, heatmap=True)`, you will get something like this: 

image

chonlei:
@MichaelClerx Really not sure why it complained!! Looking at the log: 

image

It says `./pints/_mcmc/_dream.py` flake8 error, but I didn't change it!!

chonlei:
It turned out to be a style error in master!! Fixed and rebased.

chonlei:
Merge if Gary is happy (and pass tests).

mirams:
Looks good.

pints/plot.py
MichaelClerx: Please mention this in the docstring then!

PR #625, opened 2019-01-30
Hamiltonian Monte Carlo

ben18785:
- Now allows variable step sizes for each parameter dimension
- Works well for all examples in notebook
- Added check for divergent iterations (when these occur, the previous parameter value is returned)


pints/_mcmc/_hamiltonian.py
```diff
@@ -186,11 +212,27 @@ def set_leapfrog_step_size(self, step_size):
...
             raise ValueError(
-                'Step size for leapfrog algorithm must be greater than zero.')
+                'Step size for leapfrog algorithm must' +
+                'be greater than zero.'
+            )
+        if len(a) == 1:
+            step_size = np.repeat(step_size, self._dimension)
+        elif not len(step_size) == self._dimension:
+            raise ValueError(
+                'Step size should either be of length 1 or equal to the' +
```
martinjrobins: Allowing for the leapfrog_step_size to be either 1 or of length nparam will also change the hyper-param interface

pints/_mcmc/_hamiltonian.py
```diff
@@ -77,20 +80,47 @@ def __init__(self, x0, sigma0=None):
...
+        """
+        Returns epsilon used in leapfrog algorithm
+        """
+        return self._epsilon
+
+    def scaled_epsilon(self):
+        """
+        Returns scaled epsilon used in leapfrog algorithm
+        """
+        return self._scaled_epsilon
```
MichaelClerx: Please use e.g. `np.array(self._scaled_epsilon, copy=True)` to return a copy of the internal array (otherwise, people can modify the contents of the array using the returned reference)

pints/_mcmc/_hamiltonian.py
```diff
@@ -77,20 +80,47 @@ def __init__(self, x0, sigma0=None):
...
         # Create a vector of divergent iterations
-        #self._divergent = np.asarray([], dtype='int')
+        self._divergent = np.asarray([], dtype='int')
 
         # Default threshold for Hamiltonian divergences
         # (currently set to match Stan)
-        #self._hamiltonian_threshold = 10**3
+        self._hamiltonian_threshold = 10**3
+
+    def set_epsilon(self, epsilon):
```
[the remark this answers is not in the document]
ben18785: @martinjrobins This isn't a hyper-parameter. It's diagnostic test that is specific to HMC which allows you to gauge whether there has been a mistake in solving the Hamiltonian dynamics of the particles. So don't think we want to add it to hyper-parameters as it doesn't affect sampling performance.

pints/_mcmc/_hamiltonian.py
ben18785: Ok. Can we just assume in the hyper-parameter interface that we are optimising over a vector of length == num_parameters?

pints/_mcmc/_hamiltonian.py
ben18785: @MichaelClerx That's fine but we don't seem to do that elsewhere. If we are going to make this PINTS styles perhaps it's better to make this a separate issue?

ben18785:
@MichaelClerx @martinjrobins Can you have a look at the tests here? They seem to be failing...

pints/_mcmc/_hamiltonian.py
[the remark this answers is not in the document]
martinjrobins: ahhh, ok.

pints/_mcmc/_hamiltonian.py
martinjrobins: yea, I think that is fine

martinjrobins:
> @MichaelClerx @martinjrobins Can you have a look at the tests here? They seem to be failing...

The unit tests all look like they are passing. Do you mean the code coverage?

pints/_mcmc/_hamiltonian.py
MichaelClerx: Fair enough!

ben18785:
@martinjrobins Yeah, it seems that code coverage is failing. I think it's because these lines aren't covered:

if np.abs(div) > self._hamiltonian_threshold:
  self._divergent = np.append(self._divergent, self._iterations)
  accept = 1

Any ideas how to write tests for this seeing as it's conditional?

MichaelClerx:
Since it's an error that we're catching I'm not too worried about it, so:

1. If we can set up a situation where it happens a lot, and then seed the number generator so that it happens deterministicly then great
2. If not, we can add some magic comments telling the coverage tool to ignore it

ben18785:
@MichaelClerx @martinjrobins Are we ok to merge this into master then? I want to implement the NUTS sampler which will use bits of HMC and it'd be easier if it were there...

PR #627, opened 2019-01-30
Corrected error for NormalLogPrior and added value tests

ben18785:
Tests added for log pdf value and (if implemented) sensitivities:

- Known noise log-likelihood
- Unknown noise log-likelihood
- Normal log prior


ben18785:
It's failing a test-timer thing. Any ideas @martinjrobins or @MichaelClerx ?

MichaelClerx:
Still needs to inherit from pints.LogPrior!

(Same for Cauchy and Half-Cauchy)

Issue #586, opened 2018-12-12
Replace HH IK model by one that uses analytical simulations


MichaelClerx:
Should speed things up, make @martinjrobins life a bit easier?

PR #644, opened 2019-02-05
Replaced Hodgkin-Huxley ODE model with analytical solution

MichaelClerx:
See #586

For the protocol used inside this toy example, we can actually solve the system analytically instead of using scipy's odeint. This speeds it up by about a factor 60 (!?), which is good for testing


ben18785:
Great! Nice.

MichaelClerx:
@ben18785 thanks for comparing! How small are we talking about? I'd better have a look I suppose :-)

MichaelClerx:
Thanks @ben18785 !

I'd made a mistake that showed up in the first step, where the simulation was accidentally started at t=-1 instead of t=0. This caused a notable (but exponentially declining) difference for the first 200ms or so. Have now added a notebook to `studies` to show that the remaining errors are small; the scale I'd expect from integrator error

Issue #159, opened 2017-12-28
Add an example of using a log_likelihood in an optimisation


ben18785:
At the moment we estimate the parameters of the logistic equation with noise for a squared error (yielding ML estimates of the logistic parameters) but we could actually use it to do full maximum likelihood estimation (i.e. including the noise standard deviation) if we do optimisation on the log_likelihood directly. I feel that this might provide a nicer link between the optimisation and sampling/MCMC parts of Pints.

MichaelClerx:
We can do that, yes, and Gary has done it for several problems!

However, Martin had some issues where the MLE didn't perform well compared to a simple squared error. I also feel it's conceptually much simpler to do optimisation without adding the stats...

MichaelClerx:
I suggest we do this in the form of an example using an ion channel toy model (#124) and then do our thing of running CMA-ES with a log_posterior before running MCMC

PR #652, opened 2019-02-05
Added an example of combining opt. and mcmc, and optimising on a loglikelihood

MichaelClerx:
See #159.


MichaelClerx:
@martinjrobins ?

MichaelClerx:
Fair point. Have added a new example and updated the old one

MichaelClerx:
We should also remember to keep the index in `examples/README` up to date!

MichaelClerx:
@martinjrobins care to have another look?

MichaelClerx:
@ben18785 could you have a quick peek at this?

ben18785:
Could we hold back on this — sorry been meaning to comment on it. Not sure the example here is great, to be honest...

PR #663, opened 2019-02-06
Autoregressive-moving-average(1, 1) log-likelihoods

ben18785:
@martinjrobins Built tests for your log-likelihood class and also added ARMA(1,1) log-likelihood. Also, added example notebook showing how assuming independent errors is a bad thing if your errors have these kind of persistences.


MichaelClerx:
Whenever you add any class (or stand-alone function), please remember to update the documentation in [docs](https://github.com/pints-team/pints/tree/master/docs/source) !

MichaelClerx:
@martinjrobins bit too quick there, these still need to be added to the docs!

MichaelClerx:
(I don't know if there's any way we can have an automated test for that)

martinjrobins:
sorry, forgot about that. I'll add it to the original issue

PR #664, opened 2019-02-06
Added capability to capture log-likelihood during sampling

ben18785:
Only added for MCMC at the moment as am still adapting nested sampling.


pints/_mcmc/__init__.py
```diff
@@ -363,7 +363,30 @@ def __init__(self, log_pdf, chains, x0, sigma0=None, method=None):
...
         self.set_max_iterations()
 
-        # TODO: Add more stopping criteria
+        self.set_log_likelihood_storage(False)
+
+    def set_log_likelihood_storage(self, store=False):
+        """
+        Sets whether or not to store log-likelihood for each posterior sample
+        """
+        self._log_likelihood_storage = bool(store)
```
MichaelClerx: We don't _only_ do Bayesian. Please have a look at the log pdf / log likelihood logging to disk

ben18785: @MichaelClerx Good point. I have now made two methods: log_likelihood_storage (for only Bayesian problems) and log_density_storage (for all problems). Think this is a good thing to do since certain users may want log_pdf whereas others want log_likelihoods. See what you think.

pints/_mcmc/__init__.py
```diff
@@ -377,7 +383,19 @@ def __init__(self, log_pdf, chains, x0, sigma0=None, method=None):
         self._max_iterations = None
         self.set_max_iterations()
 
-        # TODO: Add more stopping criteria
+    def log_density_storage(self):
```
MichaelClerx: @ben18785 and @martinjrobins what do we think of using the term "density" here? It sounds good to me, but we've stuck with "LogPDF" in most of the remainder!

ben18785: @MichaelClerx We use LogPDF when creating a function (as it is a log probability density function) whereas a density is a value (outputted by such a function) that we want to record. So reckon keep as is...

Happy to merge this now?

MichaelClerx: Well no, we're still using `log_pdf` pretty much everywhere else, including in this class e.g. `set_log_pdf_filename` which does the same thing but on disk... I'm also wondering about the nested and smc methods where we have multiple densities (and multiple LogPDFs for that matter) but we have input arguments `log_pdf` and `log_prior` so it would make sense to have names like `save_log_pdf_evaluations` and `save_log_prior_evaluations` rather than `save_density` which could mean either

MichaelClerx:
@ben Any reason not to use the `noise` module in the example?

MichaelClerx:
@ben18785 any reason not to use the `noise` module in the example notebook?

MichaelClerx:
I also don't like that the notebook calls the noise the "model". If I saw this I would think it was about model selection

ben18785:
This has been addressed elsewhere (I couldn't find the specific merge) -- there now exists `set_log_pdf_storage` functions in the `MCMCController` function, so closing.

PR #681, opened 2019-02-09
MALA MCMC now implemented and tested

ben18785:


MichaelClerx:
Looks amazing at first glance, will have a detailed look tomorrow. Thanks Ben!

pints/_mcmc/_mala.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def set_step_size(self, step_size=0.1):
+        """
+        Sets step size used to propose new points. Must exceed 0
+        """
+        step_size = float(step_size)
+        if step_size <= 0:
+            raise ValueError('Step size must exceed 0')
+        self._step_size = step_size
+        self.set_epsilon()
```
MichaelClerx: This will unset a manually specified epsilon!

It might be better to leave epsilon unset (e.g. `None` unless the user sets it explicitly), and then only set it automatically in initialise (if it's still None at that time)

pints/_mcmc/_mala.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        a = np.atleast_1d(scale_vector)
+        if not len(a) == self._n_parameters:
+            raise ValueError('Dimensions of scale vector must be same as ' +
+                             'number of parameters')
+        for element in scale_vector:
+            if element <= 0:
+                raise ValueError('Elements of scale vector must exceed 0')
+        self._scale_vector = np.array(scale_vector)
+        self.set_epsilon()
```
MichaelClerx: See comment above.

pints/_mcmc/_mala.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        step_size = float(step_size)
+        if step_size <= 0:
+            raise ValueError('Step size must exceed 0')
+        self._step_size = step_size
+        self.set_epsilon()
+
+    def set_scale_vector(self, scale_vector):
+        """
+        Sets vector used to scale step size for pr
```
MichaelClerx: We need a better docstring here!

pints/_mcmc/_mala.py
```diff
@@ -0,0 +1,319 @@
...
+        # First point?
+        if self._current is None:
+            if not np.isfinite(fx):
+                raise ValueError(
+                    'Initial point for MCMC must have finite log_pdf.')
+
+            # Accept
+            self._current = self._proposed
+            self._current_log_pdf = fx
+            self._current_gradient = log_gradient
```
MichaelClerx: Think this might need a minus sign, to match with what happens with the proposed_gradient below?

pints/_mcmc/_mala.py
MichaelClerx: (about `set_epsilon()`)

pints/_mcmc/_mala.py
```diff
@@ -0,0 +1,319 @@
...
+        self._proposed = self._x0
+
+        # Update sampler state
+        self._running = True
+
+    def needs_sensitivities(self):
+        """ See :meth:`pints.MCMCSampler.needs_sensitivities()`. """
+        return True
+
+    def set_step_size(self, step_size=0.1):
```
MichaelClerx: For every `set_X` method, please check if it's possible to change this during a run, or if we need to raise a `RuntimeError` if the user tries to change this parameter when `running` is true

pints/_mcmc/_mala.py
ben18785: @MichaelClerx Good spot, I think the old sign was wrong below (which I have now corrected)

pints/_mcmc/_mala.py
ben18785: @MichaelClerx Corrected

pints/_mcmc/_mala.py
ben18785: All of the set methods are fine to be changed during a run, if the user wants it so.

pints/_mcmc/_mala.py
ben18785: @MichaelClerx Thanks for these points. It was far too complicated before. Now I have removed the set_scale and set_step_size functions as we only need epsilon.

ben18785:
@MichaelClerx See what you think now of MALA. It's much cleaner now and better for it (thanks for the reviews!).

pints/_mcmc/_mala.py
ben18785: See changes (and above comment)

PR #700, opened 2019-02-15
Annealed importance sampling

### LINKED ISSUE #160: Add annealed importance sampling
Key paper: https://arxiv.org/abs/physics/9803008

Possibly the most widely used approach to calculating the marginal likelihood.

MichaelClerx:
Fixes #160


ben18785:
@MichaelClerx I'd say this is one for the future. Can we keep the branch but bin the PR? (I think I'd generally prefer having more branches than PRs as I feel like it encourages people -- like me -- to finish 'em.)

MichaelClerx:
Sure.
I made PRs from branches because nobody ever looks at the branches page and forgets what's in there. But if that makes the PR page overwhelming & less useful then sure let's shut them

PR #701, opened 2019-02-15
Tree hierarchy

MichaelClerx:


ben18785:
@MichaelClerx Let me make some changes to this then we can consider merging this. Although not sure where we should merge it into (this is a visualisation and so may be better off in a separate repo...) What do you think?

MichaelClerx:
Moved to comp-bio-paper repo

PR #709, opened 2019-02-15
SIS mcmc

MichaelClerx:
Not sure if there's a ticket for this @ben18785 ?


ben18785:
@MichaelClerx Let me add this MCMC/IS sampler to the samplers we have before we merge...

MichaelClerx:
Replaced by #760

PR #710, opened 2019-02-15
Old functional test code

MichaelClerx:
Think pretty much all of this is now handled elsewhere?


MichaelClerx:
@ben18785 can this one go now? Think it's all in the new toy pdf branch

Issue #399, opened 2018-07-23
Checking if something's a local minimum


MichaelClerx:
I had some code in Myokit to check if something was a local minimum or not. It didn't work very well, but I think that's because of the solver tolerance I was using.

Will stick this code in a Pints branch for now. Can integrate it with main Pints if anyone's interested.

MichaelClerx:
https://github.com/pints-team/pints/tree/i399-quadfit

PR #712, opened 2019-02-15
Quadratic fit around point to test if it's a minimum

MichaelClerx:
See #399 

Not sure if this is a great idea

### DIFF
--- a/quadfit.py
+++ b/quadfit.py
@@ -1,0 +1,202 @@
+def quadfit(x, y):
+    """
+    Calculates the unique quadratic polynomial through a set of points.
+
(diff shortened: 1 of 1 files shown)


ben18785:
@MichaelClerx Not sure I get this idea?

MichaelClerx:
It's something left over from my phd! Did a bit of fitting, found an optimum, got weird results, supervisor asked was it really an optimum, said if so you could fit a quadratic to it (exactly), but that never worked out.
Now thinking it was because of solver tolerances.
But also, if you have a really good optimiser it has a much better idea of what the non-noisy surface looks like like, so a naive approach might not "see" the optimum

PR #713, opened 2019-02-15
Kalman filter log-likelihood

### LINKED ISSUE #433: Kalman filter log-likelihood (ignore for now, just writing down ideas)
For an electrochemistry project I would like to try a kalman filter to estimate a model parameter that varies over time. I would like to incorporate this into pints by writing a log-likelihood class that uses a kalman filter to calculate the marginal likelihood of the other non-varying parameters by using a kalman filter on the varying parameters 

The idea is (I think) that you would have a model defined at time points k =1..N, by the state update eq:

x_{k+1}  =  A_k * x_k + w_k

where x_k is the vector of varying parameters (i.e. states), A_k is a matrix defining how the states evolve over time, and w_k are samples from a multivariate normal distribution. basemodel_k is a normal pints model that is wrapped by this likelihood

you then have an observation process which "measures" your system:

z_k = H_k * x_k + v_k + basemodel_k(p)

Where H_k and basemodel_k(p) is given by the user

You can use a kalman filter to calculate the likelihood given a

MichaelClerx:
Fixes #433


ben18785:
@martinjrobins Interesting to compare this to AR1 model as think they're the same although estimation approach is different (you use the Kalman filter to compute the log-likelihood, I use another approach).

MichaelClerx:
Fixed the `git` mess I made here. Please delete local copy and re-check out before continuing work on this

ben18785:
@martinjrobins this is an oldie but a goodie. Think we might as well try and include it, so long as we have some tests written...

Issue #424, opened 2018-07-31
Add rejection nested sampling example notebook


ben18785:
See #107

MichaelClerx:
Using multinest eggbox toy pdf

MichaelClerx:
@ben18785 is this still something we want? We have a simple egg box too, but I suppose it'd be good to have the multinest one as well?

ben18785:
This was done ages ago, I think. Will be handled in #503 and so am closing.

MichaelClerx:
We usually wait with closing until things have been merged into master, but alright :-)

Issue #503, opened 2018-09-10
Refactor nested sampling algorithms to fit in ask/tell


ben18785:
They can fit into ask/tell as follows,

- Ask: propose a new point using the active set (and/or its segmentation, for example, the ellipses). At current, the ways we have for doing this are a) sampling from the prior (rejection sampling) b) sampling from a surrounding ellipse (elliptic sampling) c) sampling from a series of ellipses (multinest) d) Galilean Monte Carlo velocity process.
- Tell: accept the point if it has a likelihood above the current threshold.

Once convergence in the evidence is reached (this can, however, be controlled by the user on the fly), generate posterior samples. This will mean that we store all the ghost particles and their weights.

This framework will greatly simplify the existing code and make it far simpler to extend it.

MichaelClerx:
See also  #568

PR #714, opened 2019-02-15
 rejection nested notebook

MichaelClerx:
#424


MichaelClerx:
This is being handled in #503 instead

PR #716, opened 2019-02-15
Return posterior pdf for all chains for Bayes Factor calculation

MichaelClerx:
Hi @sanmitraghosh !

Is this issue addressed by the new changes proposed in #664 ?  If so, I'd like to close this and delete the branch

### DIFF
--- a/pints/_mcmc/__init__.py
+++ b/pints/_mcmc/__init__.py
@@ -435,9 +435,19 @@ class MCMCController(object):
         if self._parallel:
             # Use at most n_workers workers
             n_workers = min(self._n_workers, self._chains)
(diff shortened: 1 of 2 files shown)


MichaelClerx:
@sanmitraghosh ?

sanmitraghosh:
@MichaelClerx  I am no longer using this branch, you can safely delete this. I have figured out a more reliable way Bayes factor calculation based on custom likelihood wrapper and saved likelihood values all within pints. Moreover, I don't think, as @ben18785  pointed out, that model selection is a priority right now.

MichaelClerx:
Thanks @sanmitra!

Issue #719, opened 2019-02-19
Value-based tests for log-likelihoods


ben18785:
Log likelihoods:
- [x] AR1LogLikelihood
- [x] ARMA11LogLikelihood
- [x] CauchyLogLikelihood
- [x] GaussianKnownSigmaLogLikelihood
- [x] GaussianLogLikelihood
- [x] ScaledLogLikelihood
- [x] StudentTLogLikelihood

MichaelClerx:
Closed in #728

PR #728, opened 2019-02-19
Value-based tests for log-likelihoods

ben18785:
In this checking of the tests, I (somewhat disconcertingly) found about a dozen errors that would have been picked up if we were doing value-based tests of log-likelihoods. The errors were on the following:

- AR1 values for multiple output problems
- ARMA11 values for multiple output problems
- Gaussian log likelihood values and derivatives for multiple output problems. Given that this is our workhorse distribution for many problems this was particularly concerning
- The ConstantModel which we use for testing previously had incorrect derivatives; these should always be 1 since f = theta -> df / dtheta = 1.

Covers #719

### DIFF
--- a/pints/toy/_constant_model.py
+++ b/pints/toy/_constant_model.py
@@ -30,7 +30,7 @@ class ConstantModel(pints.ForwardModelS1):
     .. math::
 
         \\frac{\partial{f_i(t)}}{dp_j} =
-            \\begin{cases} i, i = j\\\\0, i \\neq j \end{cases}
+            \\begin{cases} 1, i = j\\\\0, i \\neq j \end{cases}
 
     Arguments:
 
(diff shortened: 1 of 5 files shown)


pints/toy/_constant_model.py
```diff
@@ -106,8 +106,6 @@ def simulateS1(self, parameters, times):
             #   [df2/dp1, df2/dp2]]
             # i.e.
             #  [[1, 0],
-            #   [0, 2]]
-            # i.e. np.diag(self._r)
-            dy = np.tile(np.diag(self._r), (len(times), 1, 1))
+            #   [0, 1]]
+            dy = np.tile(np.diag(np.ones(len(self._r))), (len(times), 1, 1))
```
MichaelClerx: Thanks Ben!
Could just be `self._n` instead of `len(self._r)` then I think?

MichaelClerx: Main docstring needs updating too!

ben18785: @MichaelClerx Good idea. Both changed in my latest commit!

ben18785:
@MichaelClerx Can we merge this please if you're happy with my changes?

PR #736, opened 2019-02-21
I725 more log priors

fcooper8472:
#725 adds inverse gamma and log normal log priors.

Also reimplements Cauchy, and orders unit tests alphabetically.

PR #739, opened 2019-02-21
I737 log prior means

### LINKED ISSUE #737: Create mean functions for each log prior
This makes it easier to check what parameterisation is implemented.

fcooper8472:
Fixes #737 

This follows on from #736 which ins't merged yet, so not sure how this works in terms of a PR...

Looks as though we see all changes from master, so probably worth merging the other PR first?


MichaelClerx:
> This follows on from #736 which ins't merged yet, so not sure how this works in terms of a PR...

> Looks as though we see all changes from master, so probably worth merging the other PR first?

Thanks @fcooper8472 . #736 is now merged into master, but the PR didn't change immediately.

I wasn't sure how to get github to notice the difference, so I edited your PR twice, setting the base to <not master> and then to master again, and it seems to have picked up that it only needs 3 commits now!

pints/_log_pdfs.py
```diff
@@ -40,6 +40,13 @@ def evaluateS1(self, x):
         """
         raise NotImplementedError
 
+    def mean(self):
```
MichaelClerx: I think this should be in `LogPrior`, not in `LogPDF`. In general we don't know the mean of the LogPDFs, that's the whole point :-)

pints/_log_priors.py
```diff
@@ -76,6 +76,10 @@ def evaluateS1(self, x):
             return value, np.asarray([np.divide(self._a - 1., _x) - np.divide(
                 self._b - 1., 1. - _x)])
 
+    def mean(self):
+        """ See :meth:`LogPDF.mean()`. """
```
MichaelClerx: See comment above

pints/_log_priors.py
```diff
@@ -255,6 +263,10 @@ def evaluateS1(self, x):
         else:
             return value, np.asarray([-self._rate])
 
+    def mean(self):
+        """ See :meth:`LogPDF.mean()`. """
+        return 1. / self._rate
```
MichaelClerx: We don't need the `1.` here, it can just be `1`

pints/_log_priors.py
```diff
@@ -786,6 +826,13 @@ def evaluateS1(self, x):
         # much...
         return self(x), np.zeros(self._n_parameters)
 
+    def mean(self):
+        """ See :meth:`LogPDF.mean()`. """
+        if isinstance(self._boundaries, pints.RectangularBoundaries):
```
MichaelClerx: nice!

pints/_log_pdfs.py
MichaelClerx: Also, we don't really expect most/many LogPDF classes to implement this method; having it on the interface makes it look too much like a requirement for me

ben18785:
#744 rebases this and so closing

PR #740, opened 2019-02-22
159 opt then mcmc

MichaelClerx:
#159


MichaelClerx:
Cheers @ben18785 ! The original goal for the unidentifiable parameters notebook was to show how optimising a loglikelihood lets you easily do optimisation followed by MCMC. I was searching for an obvious case where that was useful. Could you help me think of a better example maybe?

(By the way, I know it's really artificial but I quite liked that ring :p :p :p )

ben18785:
Yep, will have a think of an example. Reckon — merge what you have minus the circle parameters notebook (sorry, I know the circle is nice, haha!). Have created an issue as definitely think this is an area we should focus on...we need more realistic examples where things fail.

MichaelClerx:
Ok, have split this into separate PRs

PR #749, opened 2019-02-26
Add SMC sampler

### LINKED ISSUE #120: Add sequential MC (SMC)

MichaelClerx:
~Still a work in progress, just need to update it to re-use a single chain and then it should be fine~

~Looks to be working now~

Fixes #120


MichaelClerx:
Had a chat with @ben18785 , decided to do it in chunks, i.e. evaluate `n_particles` in blocks of `n_processes`, then adapting in between

MichaelClerx:
@ben18785 The first example works the same as the old now, but the second will need some tuning I think. Can you have a play around with it?

If all went well - the main difference between your code and my *non-parallelised* code is that your code assumed a very wide uniform log prior if the user didn't specify one, but didn't use this prior for initialisation (for which it used a multivariate gaussian instead, for some reason).

If you have time, please play around with the 2nd example and have a look at the code.

Have also implemented parallelisation by doing batches of samples, as discussed. This reduces the performance somewhat. (On toy problems, it also slows everything down, because the overhead is much larger than the evaluation time of a single loglikelihood)

Looking forward to merging this!

MichaelClerx:
There were some conflicts between this branch and master, so merged master _into this branch_, which is apparently the done thing in these situations!
Part of merging master into this branch was resolving merge conflicts. After that, it should be easy to merge this branch into master, because they're the same!

MichaelClerx:
Tests & docs etc. in too now, have a look if you think the results are right @ben18785  :D :D

MichaelClerx:
@ben18785 can this one go now that SMC is in via another PR?

ben18785:
Closing as this never ended up working!

Issue #668, opened 2019-02-06
Add test for documentation cover


MichaelClerx:
See http://www.sphinx-doc.org/en/master/usage/extensions/coverage.html

MichaelClerx:
Started on this in https://github.com/pints-team/pints/tree/668-doc-cover but doesn't seem to work. Guessing it doesn't like things being in hidden modules (see https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/coverage.py )

May need to use a modified version of this in Pints?

MichaelClerx:
@fergus are you interested in trying to script this?
I think the steps would be:
1. Check what's references in the `rst` files in either an `autoclass` or an `autofunction` call
2. Import `pints` and use e.g. `dir(pints)` to check what is made public, so doesn't start with underscore) 
3. Compare and report

In the next step, we would provide the script with a list of modules ([pints, pints.toy, pints.plot]) to scan. Bonus points for auto-detecting those but definitely not necessary

PR #756, opened 2019-02-27
I668 doc cover rst

fcooper8472:
Addresses #668

### DIFF
--- a/docs/source/toy/neals_funnel.rst
+++ b/docs/source/toy/neals_funnel.rst
@@ -1,0 +1,8 @@
+*****************************
+Twisted Gaussian distribution
+*****************************
+
+.. module:: pints.toy
+
+.. autoclass:: NealsFunnelLogPDF
+
--- a/docs/source/toy/neals_funnel.rst
+++ b/docs/source/toy/neals_funnel.rst
@@ -1,8 +1,7 @@
 *****************************
-Twisted Gaussian distribution
+NEALS FUNNEL DISTRIBUTION
 *****************************
 
 .. module:: pints.toy
 
 .. autoclass:: NealsFunnelLogPDF
-
--- a/docs/source/toy/neals_funnel.rst
+++ b/docs/source/toy/neals_funnel.rst
@@ -1,6 +1,6 @@
-*****************************
-NEALS FUNNEL DISTRIBUTION
-*****************************
+**************************
+Neal's Funnel Distribution
+**************************
 
 .. module:: pints.toy
 
--- a/pints/noise.py
+++ b/pints/noise.py
@@ -7,7 +7,6 @@
 #  For licensing information, see the LICENSE file distributed with the PINTS
 #  software package.
 #
-import numpy as np
 
 
 def independent(sigma, shape):
@@ -29,6 +28,7 @@ def independent(sigma, shape):
         noisy_values = values + noise.independent(5, values.shape)
 
     """
+    import numpy as np
 
     # Don't test sigma/shape: handled by numpy for higher-dimensions etc.!
     return np.random.normal(0, sigma, shape)
@@ -62,6 +62,8 @@ def ar1(rho, sigma, n):
         noisy_values = values + noise.ar1(0.9, 5, len(values))
 
     """
+    import numpy as np
+
     if abs(rho) >= 1:
         raise ValueError(
             'Magnitude of rho must be less than 1 (otherwise the process'
@@ -93,6 +95,8 @@ def arma11(rho, theta, sigma, n):
     and
     ``sigma' = sigma * sqrt((1 - rho^2) / (1 + 2 * theta * rho + theta^2))``.
     """
+    import numpy as np
+
     if abs(rho) >= 1:
         raise ValueError(
             'Magnitude of rho must be less than 1 (otherwise the process'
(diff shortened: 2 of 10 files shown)


pints/noise.py
```diff
@@ -7,7 +7,6 @@
 #  For licensing information, see the LICENSE file distributed with the PINTS
 #  software package.
 #
-import numpy as np
```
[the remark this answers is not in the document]
ben18785: Fergz explained to me in person -- I'm happy now

docs/source/toy/neals_funnel.rst
```diff
@@ -0,0 +1,8 @@
+*****************************
+Twisted Gaussian distribution
```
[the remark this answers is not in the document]
ben18785: I'll change this...

fcooper8472:
Thanks @ben18785 

There's a fair bit of technical stuff so I'll wait for @MichaelClerx comments on that

docs/source/toy/neals_funnel.rst
```diff
@@ -0,0 +1,7 @@
+*****************************
+NEALS FUNNEL DISTRIBUTION
```
MichaelClerx: Don't need all-caps here!
Also *s should have same length as the words below (or sphinx breaks)

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+    check_exposed_symbols(pints.io, [], doc_symbols)
+    check_exposed_symbols(pints.noise, [], doc_symbols)
+    check_exposed_symbols(pints.plot, [], doc_symbols)
+    check_exposed_symbols(pints.toy, [], doc_symbols)
+
+    print('All classes and methods are documented in an RST file, and all '
+          'public interfaces are clean.')
+
+
+def check_exposed_symbols(_module, _submodule_names, _doc_symbols):
```
MichaelClerx: Function arguments should never have initial underscores

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+    """
+    Check _module for any classes and methods not contained in _doc_symbols,
```
MichaelClerx:
```
not contained in ``doc_symbols``
```
(same for remaining)

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+    and check for any modules not contained in _submodule_names.
+
+    :param _module: the module to check
```
MichaelClerx: This is cool, but we're not using it at the moment and not sure if our template supports it. If we start using this it'll have to be a pints-wide change in a separate commit. For now please do:

```
Arguments:

``module``
    The module to check
``submodule_names``
    List of submodules exposed by ``module``
```
etc.

(that creates a definition list)

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+    :param _submodule_names: list of submodules exposed by _module
+    :param _doc_symbols: dictionary containing lists of documented classes and
+     functions
+    """
+
+    import inspect
+    exposed_symbols = [x for x in dir(_module) if not x.startswith('_')]
+    symbols = [getattr(_module, exposed_symbols[x]) for x in
```
MichaelClerx: why not just `[getattr(module, x) for x in exposed_symbols]` ?

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+               range(len(exposed_symbols))]
+
+    classes = [x for x in symbols if inspect.isclass(x)]
```
MichaelClerx: Niiiiiiice

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+        print('FAILED')
+        sys.exit(1)
+
+
+def get_all_documented_symbols():
+    """
+    Recursively traverse docs/source and identify all autoclass and
+    autofunction declarations.
+
+    :return: A dict containing a list of classes and a list of functions
```
MichaelClerx: Same as for `:param:` above!
At the moment we just write "Returns: A dict containing..."

MichaelClerx: Or "Returns a dict..."

run-tests.py
```diff
@@ -148,6 +152,163 @@ def doctest_examples_readme():
...
+    """
+
+    doc_files = []
+    for root, dirs, files in os.walk(os.path.join('docs', 'source')):
+        for file in files:
+            if file.endswith('.rst'):
+                doc_files.append(os.path.join(root, file))
+
+    # Regular expressions to find autoclass and autofunction specifiers
+    regex_module = re.compile(r'\.\.\s*\S*module\:\:\s*(\S+)')
```
MichaelClerx: Maybe move comment down 2 lines, and add a separate comment explaining about the module regex, and how it also captures currentmodule directives

run-tests.py
fcooper8472: Why not?

run-tests.py
fcooper8472: D'oh!

fcooper8472:
@MichaelClerx pretty sure I've hit all your comments now.

run-tests.py
MichaelClerx: They're outward facing. Not hidden. Underscores are for stuff you want to hide

PR #758, opened 2019-02-28
Example of optimisation where MCMC shows unidentifiable parameters

### LINKED ISSUE #754: Think of example of unidentifiable model and include example notebook

MichaelClerx:
Initial version for @ben18785 to improve ;-)

Fixes #754


MichaelClerx:
Hi @ben18785 ! Could you write down what was wrong with this example so that we can have a go at improving it?

MichaelClerx:
Hey @ben18785 !

How about we stick this example in and you can then add a more complex example at some later point? I know this seems silly mathematically, but modellers make these mistakes quite often, because they're not thinking about parametrisation when writing their models.

MCMC is a bit overkill to show this, but it's a very simple example. And I don't think it's fundamentally different from any more complex form we can dream up. Too many parameters is too many parameters, regardless of how well you hide that fact?

MichaelClerx:
Updated this for modern PINTS syntax.

@ben18785 @martinjrobins what do you think?

chonlei:
Thanks Michael! It looks nice!

Can you write down the k = 41 + np.sqrt(k1**2 + k2**2) as an equation right above/before In[1]? So that makes it obvious why/how the model is wrong!

Can you either run a third chain from the other end of the circle (e.g. [-sqrt(18), 0]) or run the MCMC a bit longer so that it covers the ring a bit better? (I thought it was a banana distribution that's poorly sampled at first glance...)

And can you add all missing x, y labels -- the time series plots as well as the last k1 vs k2 parameter plot?

MichaelClerx:
Thanks @chonlei ! Changed as requested!

MichaelClerx:
@chonlei ?

chonlei:
Thanks! Not sure why k1 (y-axis) is still not showing up!

chonlei:
Ah, got a typo there! You've set both being xlabels!

MichaelClerx:
> Ah, got a typo there! You've set both being xlabels!

Thanks @chonlei Fixed it!

PR #760, opened 2019-02-28
SIS MCMC redux

MichaelClerx:
Updated version of SIS MCMC PR, after history changes on master

Old PR: #709


ben18785:
@MichaelClerx I'm trying to think of the best way to implement this method (as it's one of the remaining methods in #917). 

But it's not really a method. More a framework that can be applied to any sampling (or optimisation) routine.

For sampling, all it does is:

  - 'heats up' distribution according to some temperature,
  - samples on the tempered distribution using whatever method; recording the log-posterior of each sample (do we have functionality to do this?,
  - reweights samples based on importance sampling using the log-posterior values as weights.

For optimisation, all it does is heat up the distribution then runs whatever optimiser on it.

So, I was thinking, perhaps MCMCController, NestedController and OptimisationController could just have an optional argument ``temperature`` which, by default, is zero. If it is non-zero, the methods form the heated distribution and (for sampling methods) ensure that the log-posterior values are recorded.

What do you think?

MichaelClerx:
Interesting!

I see what you mean, and there's certainly meta methods for optimisation as well.

But I think there's more than 1 meta-sampler, so sticking it in MCMCController might not be the way to go.

Could we implement this like some of the other samplers, with an option to change whatever underlying sampler you use? Then you'd also get all the log-posteriors, because:

    class SISMCMC(...)
    ...
    def ask(...)
        return self._inner_sampler.ask()
    def tell(self, log_posterior)
        # store log posterior
        return self._inner_sampler.tell(log_posterior)

ben18785:
@MichaelClerx Good idea! I'll do it like that.

MichaelClerx:
Hey @ben18785 , do you remember why we closed this? Can I delete the branch as well?

Issue #757, opened 2019-02-28
Add script to check all "slow notebooks" exist on travis


MichaelClerx:
When notebooks that are listed as "slow" are renamed, the cron job falls over.

It'd be nicer if we have a test on travis PRs that checked this. @fcooper8472 this should be an easier one?

Pretty much a for-loop over the contents of https://github.com/pints-team/pints/blob/master/.slow-books (minus comments)

fcooper8472:
@MichaelClerx  why do some in that list end in `ipynb` and some don't?

MichaelClerx:
Good question! Maybe I was being lazy? Or maybe I made a parser that could handle both and wanted to test it... Feel free to change the rules here, no strong feelings from me!

fcooper8472:
Fair enough.  When I do this one I'll add the extensions to each.  Since this is also essentially documentation (documenting which books are slow), it can go in as another doctest.

MichaelClerx:
>  Since this is also essentially documentation (documenting which books are slow), it can go in as another doctest.

Hmmmmm! OK it can go in there but because the notebooks themselves are part of the documentation :D

PR #762, opened 2019-03-06
I757 slow notebooks

fcooper8472:
Addresses #757

### DIFF
--- a/run-tests.py
+++ b/run-tests.py
@@ -91,6 +91,9 @@ def run_doctests():
     # unintended modules are exposed via a public interface
     doctest_rst_and_public_interface()
 
+    # Check all .slow-books exist (they haven't been, e.g., renamed or removed)
+    doctest_slow_books()
+
 
 def doctest_sphinx():
     """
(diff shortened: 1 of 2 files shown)


run-tests.py
```diff
@@ -316,6 +323,50 @@ def get_all_documented_symbols():
...
+        )]
+
+    if len(other_lines) > 0:
+        print('The following entries are in .slow-books but are not ipynb:')
+        print('  {}'.format('\n  '.join(other_lines)))
+        print('FAILED')
+        sys.exit(1)
+
+    examples = [l for l in os.listdir('examples') if l.endswith('ipynb')]
+    undocumented = [b for b in slow_books if b not in examples]
```
MichaelClerx: Maybe use sets here?

fcooper8472: Maybe next time ;).  I personally find the `difference()` method on sets difficult to read because I always have to check whether it's `A-B` or `B-A`.  I guess here it would be clear from the context, but I've tended to avoid it in the past.

PR #782, opened 2019-04-16
Add an example to the README

MichaelClerx:
See: https://github.com/pints-team/jors-paper/issues/12


MichaelClerx:
I've written a longer example now, but not sure this is the way forward!

Perhaps it'd be better to use a figure (Figure 1 in the paper?) for this, so we can have (fake) syntax highlighting ?

MichaelClerx:
OK Different approach: An image on the front-page (that we can re-use in the paper), and a file with (similar) code in the examples

MichaelClerx:
Viewable here: https://github.com/pints-team/pints/blob/83d0c184419cc749b75c1c7bee99fd99f497e357/README.md

MichaelClerx:
What do you think @martinjrobins ?

PR #791, opened 2019-04-17
Contact section in README

MichaelClerx:
Add some contact info (email address) to README, and link to the issues page.


MichaelClerx:
Is this a good idea? Should we have some kind of maillist that anyone can email instead @fcooper8472 ?
And should we put plain email addresses? I've done it on the myokit website, but guess scrapers will find github a lot more easily?

MichaelClerx:
Okay I'll stick that in!

PR #793, opened 2019-04-17
Updated example in README

MichaelClerx:
Added some whitespace and a title + caption


MichaelClerx:
Thanks for the comments @chonlei . Is this better?

Issue #784, opened 2019-04-16
Find out why adaptive covariance func test results have changed slightly


MichaelClerx:
And emcee hammer

My current guesses:

1. Changes I made to functional testing itself, not changes to the methods
2. Some change in shared code, e.g. picking a default sigma

MichaelClerx:
Investigating this is made a bit harder by the fact that I removed the studies that time, which changed a lot of old commit hashes :-|

MichaelClerx:
So I've added some functionality to `pfunk` to re-investigate old commits (but store the results in a temporary directory), and am now running tests with

```
/funk investigate -c 02be2f50e3abc13485e8ffe9080a5be4ae167a80 mcmc_banana_AdaptiveCovarianceMCMC_1 temp -r 6
```

on scrambler

MichaelClerx:
This is everything since february

![mcmc_banana_AdaptiveCovarianceMCMC_1-2019-04-25-04:28:50](https://user-images.githubusercontent.com/517644/56719400-893fcc80-6738-11e9-89a2-8a221a99affc.png)

MichaelClerx:
Zoomed in on the end:
![mcmc_banana_AdaptiveCovarianceMCMC_1-2019-04-25-08:59:04](https://user-images.githubusercontent.com/517644/56719436-9c529c80-6738-11e9-83ae-c69d859134c1.png)

MichaelClerx:
I think this really means we can't use a threshold for testing. The top graph shows there were several times when exactly this behaviour occurred, and then *presumably* we fixed it

MichaelClerx:
This seems to say the bug is between these two commits:

```
commit c1218dda4def71a3657a33c7ff626fb891eec9e1
Merge: be6ba60a 90746169
Author: Michael Clerx <MichaelClerx@users.noreply.github.com>
Date:   Fri Feb 22 09:45:06 2019 +0000

    Merge pull request #736 from pints-team/i725-more-log-priors
    
    I725 more log priors

commit 907461692f4a226414c594cc080618fab5ca3431
Author: Fergus Cooper <fcooper8472@gmail.com>
Date:   Thu Feb 21 17:19:52 2019 +0000

    #725 Correct support for log normal prior
```

But I can't really see how that would be the case...

MichaelClerx:
[...]

PR #806, opened 2019-04-25
Cosmetic changes to twisted gaussian

MichaelClerx:
Also added value-based test (before I made the changes) to test they don't affect the result

See #784


fcooper8472:
I don't understand how coverage has decreased..?

MichaelClerx:
Me neither. Codecov gets confused sometimes! Seems to have fixed itself now, back at 100%

PR #812, opened 2019-04-30
MCMC.replace() method update, redux

MichaelClerx:
Last one broke master somehow, investigating why


MichaelClerx:
OK, had nothing to do with this PR. Merging again!

Issue #813, opened 2019-04-30
Test for CMA-ES error message fails: Has the wording changed?


MichaelClerx:

MichaelClerx:
Closed in #814

PR #814, opened 2019-04-30
Removed test that no longer triggers error condution with newer cma module

MichaelClerx:
See #813

### DIFF
--- a/pints/tests/test_opt_cmaes.py
+++ b/pints/tests/test_opt_cmaes.py
@@ -80,11 +80,7 @@ class TestCMAES(unittest.TestCase):
         found_parameters, found_solution = opt.run()
         self.assertTrue(found_solution < 1e-3)
 
-    '''
-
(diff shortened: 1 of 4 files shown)


MichaelClerx:
@martinjrobins could you have a look at this one? We used to be able to trigger CMA-ES to give us an ill-conditioned matrix, but that test no longer produces that result in newer CMA versions, so have disabled it. If we see this error in the wild again we should try and convert it to a new test case...

pints/tests/test_opt_cmaes.py
```diff
@@ -79,6 +80,11 @@ def test_bounded_and_sigma(self):
         found_parameters, found_solution = opt.run()
         self.assertTrue(found_solution < 1e-3)
 
+    '''
+
+    # In newer versions of CMA this test doesn't trigger this condition
+    # anymore...
+
```
martinjrobins: use `skip` decorator instead?

PR #841, opened 2019-05-24
Updating minimum Python 3 version to 3.5

### LINKED ISSUE #839: Drop Python 3.4 support
Getting these warnings on build systems:

```
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).
```

So would make sense to remove 3.4 testing (and speed thing up a bit?)

Might start thinking about adding 3.7 testing too (if that's stable for all our dependencies)

MichaelClerx:
Fixes #839

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,7 +6,6 @@ environment:
 
   matrix:
     - PYTHON: "C:\\Python27-x64"
-    - PYTHON: "C:\\Python34-x64"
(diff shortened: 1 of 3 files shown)


MichaelClerx:
Ping!

Issue #864, opened 2019-07-14
Allow individual MCMC branches to return None (when others return samples)


lorcandelaney:
In the pints/_mcmc/init.py, in the run() method, specifically in line 564, there's a check that ensures that the three chains all return a sample or "None" for intermediate steps. However, this check does not work well with slice samplers, since it is not possible to predict the number of intermediate steps required in an iteration of a single chain. As a result, it is inevitable to have some chains returning "None" while others return a sample in the same MCMC iteration. I believe that we might need to change the check in order to use the MCMCController for slice sampling.

MichaelClerx:
Sample implementation by @lorcandelaney here: https://github.com/pints-team/pints/blob/c4fcd6f0f78e3eca2e9d72bc40cdc2f9a70209f1/pints/_mcmc/__init__.py#L550-L575

MichaelClerx:
I'm not averse to having something like this. As far as I can remember it wasn't implemented at the time because we didn't need it.

However, the implementation will be a bit trickier than the code above I think, because we want it to be parallelised more efficiently for single chain samplers.

So the code (for single chain samplers) should probably do something like this :

* set all samplers to "active" state
* ask() each active sampler
* evaluate() the given samples
* tell() each active sampler
* store any new samples in some kind of buffer - it should be possible for some of the samplers to be several steps ahead of the others
* whenever the buffer has a sample for every chain, update the chains in memory, call the disk-logging methods etc.
* whenever any chain has the maximum number of samples, set it to inactivate: stop ask() telling for it

So the different samplers would run out-of-step with each other, but since they're all doing the same problem they should all take roughly the same number of iterations to reach N samples, over time.

MichaelClerx:
Fixed in 2b74a3ce1467d88e715c8ad9fea02a9958692efa (See #882 )

ben18785:
@lorcandelaney @MichaelClerx Can we close this now?

PR #863, opened 2019-07-03
Slice sampling integration: 1. Stepout, 2. Doubling, 3. Overrelaxation, 4. Hyperrectangles, 5. Adaptive hyperrectangles, 6. Rank Shrink, 7. Covariance Match, 8. Generalised Elliptical

ben18785:

Issue #306, opened 2018-03-29
Implement slice sampling


ben18785:
Key [paper](http://www.jstor.org/stable/3448413?casa_token=7mFwj3YoxzoAAAAA:tXuRfVdxzzgze_CyV5kPiAM6LcZGXhJLDouKFITwznZGkR48BCzxZt9nJ8KNTWDUmwMcBbW_UJ3UZi_RpyWEMjY16V4se5HQK3rJYAhTkdhPqPU7&seq=1#page_scan_tab_contents) by Radford Neal. There are also newer approaches here that are probably worth looking into. Slice sampling is actually used in Stan.

ben18785:
[This](https://www.sciencedirect.com/science/article/pii/S1053811915001846) paper suggests that slice sampling can be effective for time series models of similar complexity to those that we consider.

ben18785:
Closing as a duplicate of #772

PR #869, opened 2019-07-16
Allow individual MCMC branches to return None (when others return samples)

MichaelClerx:
See #864

### DIFF
--- a/examples/sampling-hamiltonian-mcmc.ipynb
+++ b/examples/sampling-hamiltonian-mcmc.ipynb
@@ -681,21 +681,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "python2"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
     "name": "ipython",
-    "version": 2
+    "version": 3
    },
    "file_extension": ".py",
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython2",
-   "version": "2.7.14"
+   "pygments_lexer": "ipython3",
+   "version": "3.6.4"
   }
  },
  "nbformat": 4,
--- a/pints/_mcmc/__init__.py
+++ b/pints/_mcmc/__init__.py
@@ -84,6 +84,9 @@ class SingleChainMCMC(MCMCSampler):
         # Get number of parameters
         self._n_parameters = len(self._x0)
 
+        # State of sampler
+        self._active = True
+
         # Check initial standard deviation
         if sigma0 is None:
             # Get representative parameter value for each parameter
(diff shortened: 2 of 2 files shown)


MichaelClerx:
Required for #863 and slice samplers in general (#306)

lorcandelaney:
@MichaelClerx Hey Michael! I hope all is good. I have just uploaded a new version for the code, following your previous suggestions. What do you think?

examples/sampling-hamiltonian-mcmc.ipynb
```diff
@@ -681,21 +681,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
```
MichaelClerx: Hey @lorcandelaney 

Testing is good, but try not to `git add` the notebooks every time you re-run them (it adds a lot of clutter to the commit history)

pints/_mcmc/__init__.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, x0, sigma0=None):
         # Get number of parameters
         self._n_parameters = len(self._x0)
 
+        # State of sampler
+        self._active = True
```
MichaelClerx: I'd prefer not to have this in every optimiser - it's really only something the Controller needs to know about. So better just being a list inside `run()` I think

(Optimisers can also be used without a controller, for example!)

pints/_mcmc/__init__.py
```diff
@@ -543,39 +546,70 @@ def run(self):
 
             # Get points
             if self._single_chain:
-                xs = [sampler.ask() for sampler in self._samplers]
+                # First step: create array for initial samples
+                if evaluations == 0:
```
MichaelClerx: You've got:

```
evaluations = 0
while True:
    if evaluations == 0:
        ...do special thing
```
that can just be 
```
...do special thing
while True:
    ...
```

pints/_mcmc/__init__.py
```diff
@@ -543,39 +546,70 @@ def run(self):
...
+                # samplers
+                else:
+                    for i, sampler in enumerate(self._samplers):
+                        if sampler._active:
+                            xs[i] = sampler.ask()
             else:
                 xs = self._samplers[0].ask()
 
             # Calculate logpdfs
             fxs = evaluator.evaluate(xs)
```
MichaelClerx: You're re-evaluating the same `x` for inactive samplers here. That's potentially very costly so better to avoid that!

I would suggest doing this a different way. For example, you could keep an array of indices of active samplers, then
```
for i, sampler in enumerate(self._sampler):
    if sampler._active:
        ...
```
just becomes
```
for i in i_active_samplers:
    ...
```
and the number of active samplers is `len(i_active_samplers)`

MichaelClerx:
One more thing @lorcandelaney , it seems I can't be listed as a reviewer on this PR, because I created it myself :roll_eyes: :smile: 

Once you're done with the above comments, could you please let me know & I'll close this PR so that  you can create a new one? Thanks

lorcandelaney:
> One more thing @lorcandelaney , it seems I can't be listed as a reviewer on this PR, because I created it myself 🙄 😄
> 
> Once you're done with the above comments, could you please let me know & I'll close this PR so that you can create a new one? Thanks

Hi @MichaelClerx ! Thanks for your comments! I have made the relevant changes to the code. I believe that we can delete this PR and create a new one with you as the reviewer. I will save these comments and include them on the new PR so that we can refer to them if needed.

lorcandelaney:
Hi @MichaelClerx, hope you had a good weekend! Shall we proceed and delete this PR? I will then create a new one with you as a reviewer.

MichaelClerx:
Yes please!

Issue #593, opened 2019-01-08
Add Nelder-Mead method


MichaelClerx:
Looking at scipy: https://github.com/scipy/scipy/blob/master/scipy/optimize/optimize.py#L455-L681
And this: https://github.com/fchollet/nelder-mead/blob/master/nelder_mead.py

it shouldn't be too hard to add an ask-and-tell nelder-mead to Pints

MichaelClerx:
See also #684 and #572

MichaelClerx:
Closed in #880

PR #878, opened 2019-07-22
Nelder-Mead local optimiser

MichaelClerx:
See #593 

Still need to add some docs, notebook, and see about boundaries


MichaelClerx:
Codecov having some issues?

PR #879, opened 2019-07-23
Nelder-Mead optimiser

MichaelClerx:
See #593


MichaelClerx:
Codecov seems broken?

PR #880, opened 2019-07-23
Nelder-Mead Optimiser

MichaelClerx:
See #593


pints/_optimisers/_nelder_mead.py
```diff
@@ -0,0 +1,337 @@
...
+#  For licensing information, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import numpy as np
+import pints
+
+
+# Flow
```
ben18785: Very nice way of illustrating how the algorithm works!

pints/_optimisers/_nelder_mead.py
```diff
@@ -0,0 +1,337 @@
...
+#  else
+#    accept xr, unset xr, return (ask will set xr)
+
+
+class NelderMead(pints.Optimiser):
+    """
+    Nelder-Mead downhill simplex method.
+
+    Implementation of the classical algorithm by [1], following the
+    presentation in Algorithm 8.1 of [2].
```
ben18785: Nicely documented! Do you think it's worth (later on; perhaps as a different optimiser) implementing Algorithm 8.2 since this corrects some of the pathologies of NM? In particular, it tries to guard against the simplex becoming of very small volume...

pints/_optimisers/_nelder_mead.py
```diff
@@ -0,0 +1,337 @@
...
+        # Initialise
+        if self._fs is None:
+            fx = np.array(fx, copy=True)
+            if np.prod(fx.shape) != self._n_parameters + 1:
+                raise ValueError(
+                    'Expecting a vector of length (1 + n_parameters).')
+            self._fs = fx.reshape((1 + self._n_parameters, ))
+
+            # Return: ask will set xr, get f(xr)
+            return
```
ben18785: Due to my Pythonic ignorance, I don't know what return on its own does?

pints/_optimisers/_nelder_mead.py
MichaelClerx: Thanks! It took me a while to fit it into ask-and-tell :-D

pints/_optimisers/_nelder_mead.py
MichaelClerx: It's probably a good idea yeah. Should we add a ticket? And do you think it should be a separate optimiser or an option or even just the default?

pints/_optimisers/_nelder_mead.py
MichaelClerx: Its equivalent to `return None`. I'll change to that explicitly

MichaelClerx: Actually I won't! Got a bit confused there with the MCMC `tell()`, which returns something or doesn't in special cases. Optimiser tell()s never return anything, so will stick to `return` meaning "exit the function now" (and assumes no value needs returning)

pints/_optimisers/_nelder_mead.py
MichaelClerx: Have made a ticket!

PR #882, opened 2019-07-29
Integrate Slice Sampling: Stepout and Doubling

lorcandelaney:
Integration of Slice Sampling: Stepout and Doubling methods. 

**Important:**
1. The Stepout method includes Overrelaxation as an additional option.
2. The ``pints/_mcmc/__init__.py`` file has been modified to allow individual MCMC branches to return None (when others return samples).


pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,397 @@
...
+        chain = []
+        while len(chain) < 100:
+            x = mcmc.ask()
+            fx = log_pdf.evaluateS1(x)[0]
+            sample = mcmc.tell(fx)
+            if sample is not None:
+                chain.append(np.copy(sample))
+
+        # Fit Multivariate Gaussian to chain samples
+        np.mean(chain, axis=0)
```
ben18785: What do these lines do at the end of the test? Shouldn't we be comparing these means to the actual means?

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,397 @@
...
+
+        # Set up modest logging
+        mcmc.set_log_to_screen(True)
+        mcmc.set_log_interval(500)
+
+        # Run!
+        print('Running...')
+        mcmc.run()
+        print('Done!')
+
```
ben18785: I'm not sure I follow what these last lines of the testing procedures are doing? They are checking that the samplers run but not sure what else they do? Do we want to check the length of the samples that are outputted for example? And the means of the outputted samples?

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,397 @@
...
+        chain = []
+        while len(chain) < 100:
+            x = mcmc.ask()
+            fx = log_pdf.evaluateS1(x)[0]
+            sample = mcmc.tell(fx)
+            if sample is not None:
+                chain.append(np.copy(sample))
+
+        # Fit Multivariate Gaussian to chain samples
+        np.mean(chain, axis=0)
```
ben18785: Again, not sure what these lines are doing?

pints/tests/test_slice_stepout.py
lorcandelaney: Hi @ben18785, I used to print these to compare the mean and covariance of the fitted Gaussian with the mean and covariance of the target distribution. It served as a guideline while testing to check whether the sampler was working as expected.

pints/tests/test_slice_stepout.py
lorcandelaney: Hi @ben18785 - As initially I was having some issues running the samplers on the logistic model (I don't remember what the issue was), I implemented the logistic problem as a test to check that everything was running properly. This test doesn't check anything else.

pints/tests/test_slice_stepout.py
lorcandelaney: @ben18785 Please refer to the previous comment.

MichaelClerx:
Cool stuff, will have a detailed look tomorrow!

(Not important, but just to get the terminology right: We write (1) code, and (2) unit tests, and then (3) use coverage testing to see if our unit tests "cover" our code. So you might change the unit tests to improve their coverage, but you don't actually change the coverage tests themselves - that's just a program we run!)

pints/tests/test_slice_stepout.py
MichaelClerx: I'm with Ben on this one! That line just calculates a mean and then doesn't assign the result to anything...

MichaelClerx:
Good stuff! I'd still like to have a quick look tonight or tomorrow before we merge, is that OK?

ben18785:
Cool. That'd be great -- thanks Michael!

Best,

Ben

pints/_mcmc/__init__.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, x0, sigma0=None):
         # Get number of parameters
         self._n_parameters = len(self._x0)
 
+        # State of sampler
+        self._active = True
```
MichaelClerx: I think this can go now?

pints/_mcmc/__init__.py
```diff
@@ -543,39 +550,63 @@ def run(self):
...
+                    xs.append(self._samplers[i].ask())
             else:
                 xs = self._samplers[0].ask()
 
-            # Calculate logpdfs
+            # Calculate logpdfs of points in active samplers
             fxs = evaluator.evaluate(xs)
 
             # Update evaluation count
-            evaluations += len(fxs)
```
MichaelClerx: Is this chain necessary? The number of evaluations should equal the number of active samplers, no?

pints/_mcmc/__init__.py
```diff
@@ -543,39 +550,63 @@ def run(self):
...
+                # First step: create array of samples
+                if evaluations == self._chains:
+                    samples = np.array([
+                        s.tell(fxs[i]) for i, s in enumerate(self._samplers)])
 
-                none_found = [x is None for x in samples]
+                # Subsequent steps: update values for active samplers
+                # until all chains return a sample
+                else:
+                    temp = i_active_samplers[:]
```
MichaelClerx: What does `temp` do? Is it not equal to `i_active_samplers` ?

pints/_mcmc/__init__.py
```diff
@@ -543,39 +550,63 @@ def run(self):
...
-            # be checked
+            # If at least one chain hasn't returned a sample, the iteration
+            # count shouldn't be updated, logging shouldn't be triggered,
+            # and stopping criteria shouldn't be checked
             if intermediate_step:
                 continue
 
             # Add new samples to the chains
-            chains.append(samples)
+            chains.append(np.array(samples, copy=True))
```
MichaelClerx: Why `copy=True` here? I don't think we re-use the samples object do we?

pints/tests/test_slice_stepout.py
MichaelClerx: Also just noticed the filenames here should be something like `test_mcmc_slice_sampling.py` (all the mcmc method tests start with test_mcmc)

MichaelClerx: Also, the unit tests shouldn't print anything please :-)
So that means no `print(...)`, but also `mcmc.set_log_to_screen(False)`

MichaelClerx: It'd also be great if you could reduce the time it takes for these tests to run!

MichaelClerx: For example, you don't need to test on different problems unless those problems trigger very different parts of the code. Testing how the methods behaves on different problems is done in the functional testing - which is a separate problem you don't need to worry about :D

pints/tests/test_slice_stepout.py
MichaelClerx: The ideal way to write unit tests is to have them test very small bits of code (ideally in different situations), and then check that the output is what you expect. The idea being that if all the little bits function well individually, the whole should work well too. So if you encounter some issues in a particular problem, the trick is to recreate that issue in a very localised way: set up the exact situation that made it break (with very little overhead), and then re-run only the bit where the error occurred.

So instead of re-running the whole sampling routine as a test you would
1. Isolate what the problem is
2. Find some way to recreate that problem that has waaaay less overhead
3. Write a test (that fails)
4. Fix the bug

MichaelClerx: So in this case you'd ideally have a test where you brought the sampler into some state that you know used to break it, then do a few steps via ask-and-tell, check the output is correct. No controller involved!

pints/_mcmc/__init__.py
MichaelClerx: I'll do this one!

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+        self.assertEqual(mcmc._r, None)
+        self.assertEqual(mcmc._temp_l, None)
+        self.assertEqual(mcmc._temp_r, None)
+        self.assertEqual(mcmc._k, None)
+        self.assertEqual(mcmc._j, None)
+        self.assertEqual(mcmc._fx_l, None)
+        self.assertEqual(mcmc._fx_r, None)
+
+        self.assertEqual(mcmc._m, 50)
+        self.assertEqual(mcmc._active_param_index, 0)
```
MichaelClerx: These are OK for now, but ideally a unit test should test the API of a class works the way it should: for example:

```
x.set_value(5)
self.assertEqual(x.get_value(), 5)
```

How the object stores that value internally is not something you test. In fact, it should be easy to change the way things are handled internally without changing the test status: Just as long as it still works!

So internally I could have `self._value = 5`, or `self._super_value = 5` or `self._values = [5]`. It doesn't matter for the test, as long as it acts correctly _via the API_

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+import pints.toy
+
+debug = False
+
+
+class TestSliceStepout(unittest.TestCase):
+    """
+    Tests the basic methods of the Slice Sampling with Stepout routine.
+
+    Please refer to the _slice_stepout.py script in ..\_mcmc\_slice_stepout.py
```
MichaelClerx: This is a nice touch!
Unfortunately, it could go out of date if we refactor the code at some point, so maybe best to remove it?

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+        Tests the very first run of the sampler.
+        """
+        # Create log pdf
+        log_pdf = pints.toy.GaussianLogPDF([2, 4], [[1, 0], [0, 3]])
+
+        # Create mcmc
+        x0 = np.array([2., 4.])
+        mcmc = pints.SliceStepoutMCMC(x0)
+
+        # Ask should fail if _ready_for_tell flag is True
```
MichaelClerx: Don't change the internals in a test! When writing the test, you don't know about the internals, or this flag!
So do something like:
```
mcmc.ask()
mcmc.ask()
```
and if the docstrings etc. say the second call should fail then test for that.

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+        with self.assertRaises(RuntimeError):
+            mcmc._ready_for_tell = False
+            mcmc.tell(fx)
+
+        # Undo changes
+        mcmc._ready_for_tell = True
+
+        # Test first iteration of tell(). The first point in the chain
+        # should be x0
+        self.assertTrue(np.all(mcmc.tell(fx) == x0))
```
MichaelClerx: This line is perfect!

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # We update the current sample
+        self.assertTrue(np.all(mcmc._current == x0))
```
MichaelClerx: These should all go!

(But you _can_ test this is there are already methods that give you the current sample!)

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+        # Create mcmc
+        x0 = np.array([1, 1])
+        mcmc = pints.SliceStepoutMCMC(x0)
+
+        mcmc.set_w(30)
+        mcmc.set_a(2)
+
+        # Run multiple iterations of the sampler
+        chain = []
+        while len(chain) < 300:
```
MichaelClerx: Why 300? Please make this number as low as possible without ruining the test. Remember everyone needs to run this test every time they change a bit of PINTS code

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+            x = mcmc.ask()
+            fx = log_pdf.evaluateS1(x)[0]
+            sample = mcmc.tell(fx)
+            if sample is not None:
+                chain.append(np.copy(sample))
```
MichaelClerx: You're creating a chain here, but then not doing anything with it. Can you somehow test if the result is correct?

For example, you could simply test if the shape of the created array is correct.

Maybe have a look at the other MCMC method tests as well! You typically don't need to use an MCMCController for the MCMCSampler methods tests (the controller has its own test, and we test if the returned chains are correct in our separate "functional testing" project)

pints/tests/test_slice_stepout.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # Test set_m
+        mcmc.set_m(3)
+        self.assertEqual(mcmc._m, 3.)
+        with self.assertRaises(ValueError):
+            mcmc.set_m(-1)
+
+        # Test set_prob_overrelaxed
+        mcmc.set_prob_overrelaxed(.5)
+        self.assertEqual(mcmc._prob_overrelaxed, .5)
```
MichaelClerx: Is there a `get_prob_overrelaxed` you could use here instead?

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class SliceDoublingMCMC(pints.SingleChainMCMC):
+    """
+    *Extends:* :class:`SingleChainMCMC`
```
MichaelClerx: This line should go right to the bottom, just before the references.

(We changed this a while ago, but apparently we failed to change it in the file you copy-pasted :D)

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+
+    Implements Slice Sampling with Doubling, as described in [1]. This is a
+    univariate method, which is applied in a Slice-Sampling-within-Gibbs
+    framework to allow MCMC sampling from multivariate models.
+
+    Generates samples by sampling uniformly from the volume underneath the
```
MichaelClerx: Great docstring!

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+        self._init_check = False
+        self._continue_check = False
+
+        return None
+
+    def name(self):
+        """ See :meth:`pints.MCMCSampler.name()`. """
+        return 'Slice Sampling - Doubling'
+
+    def set_w(self, w):
```
MichaelClerx: For parameters that have a well-defined meaning, we use a long name.

So `set_width()` or `set_learning_rate()`.

Occasionally there's no way around it and we end up with a `set_epsilon`, but this should be avoided wherever possible

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+        Sets width w for generating the interval.
+        """
+        if type(w) == int or float:
+            w = np.full((len(self._x0)), w)
+        if any(n < 0 for n in w):
+            raise ValueError("""Width w must be positive
+                            for interval expansion.""")
+        self._w = w
+
+    def set_p(self, m):
```
MichaelClerx: There should be a name for this! (Also set_p sets m?)

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+        return 'Slice Sampling - Doubling'
+
+    def set_w(self, w):
+        """
+        Sets width w for generating the interval.
+        """
+        if type(w) == int or float:
+            w = np.full((len(self._x0)), w)
+        if any(n < 0 for n in w):
+            raise ValueError("""Width w must be positive
```
MichaelClerx: Does this not result in a lot of spaces in the message? Please use:

```
raise ValueError('Hello this line is just too long for the 79 character limit'
                 'that we use.')
```
or
```
raise ValueError(
    'Hello again, this line is also too long for the 79 character limit used'
    'in PINTS.')
```

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+
+        return None
+
+    def name(self):
+        """ See :meth:`pints.MCMCSampler.name()`. """
+        return 'Slice Sampling - Doubling'
+
+    def set_w(self, w):
+        """
+        Sets width w for generating the interval.
```
MichaelClerx: What should the width be? A single number or an array? If an array, then what shape?

pints/_mcmc/_slice_doubling.py
```diff
@@ -0,0 +1,583 @@
...
+        """
+        return self._p
+
+    def current_log_pdf(self):
+        """ See :meth:`SingleChainMCMC.current_log_pdf()`. """
+        return self._current_log_pdf
+
+    def current_slice_height(self):
+        """
+        Returns current log_y used to define the current slice.
```
MichaelClerx: Don't use variable names in the docstrings please!

pints/_mcmc/_slice_stepout.py
```diff
@@ -0,0 +1,681 @@
...
+            else:
+                self._r = self._proposed[self._active_param_index]
+                self._temp_r[self._active_param_index] = self._r
+            return None
+
+    def name(self):
+        """ See :meth:`pints.MCMCSampler.name()`. """
+        return 'Slice Sampling - Stepout'
+
+    def set_w(self, w):
```
MichaelClerx: See comments above

pints/_mcmc/_slice_stepout.py
```diff
@@ -0,0 +1,681 @@
...
+        self._first_expansion = False
+
+        # Flag indicating whether the interval expansion is concluded
+        self._interval_found = False
+
+        # Number of steps used for expanding the interval ``I``
+        self._j = None
+        self._k = None
+
+        # Flags used to calculate log_pdf of initial interval edges ``l,r```
```
MichaelClerx: There's a lot of values stored here!
Are all of these needed (for example, they get set in ask() and then used in tell())? Or can some of them just be local variables in the methods that use them? Please check for both classes

pints/_mcmc/_slice_doubling.py
lorcandelaney: @MichaelClerx Hi Michael! Thanks for the comment. For many parameters (such as ``w``, ``m``, ``p``, ``k``,  ``j``) I used the notation presented in Neal's original paper, thinking that a user could refer to the paper and to the relevant pseudocode to have a deeper understanding of their functioning and impact on the performance of the samplers (if needed). Would you like me to change the name of some variables anyway or do you think that preserving the same notation as the original paper would be a good idea?

MichaelClerx: The internal variables can be whatever you like, but wherever possible the method and class names should be something descriptive. You can always refer to the original notation in the docstring, e.g.

```
set_learning_rate(self, eta):
    """
    Sets the learning rate (eta in the original paper [1], see equation 17)
```

pints/_mcmc/_slice_doubling.py
MichaelClerx: See comment above! If you think it helps then mention it.

MichaelClerx:
Just added a test to fix a coverage issue. Be sure to git pull before committing new stuff!

lorcandelaney:
@MichaelClerx Hi Michael! I hope that you are having a great weekend! Thanks for all the comments - it's great to be able to learn from you guys! 

I have made the relevant changes to the files. What do you think?

pints/_mcmc/_slice_stepout.py
```diff
@@ -0,0 +1,684 @@
...
+    def bisection_steps(self):
+        """
+        Returns integer limit overrelaxation endpoint accuracy to
+        ``2^(-bisection steps) * width``.
+        """
+        return self._a
+
+    def current_log_pdf(self):
+        """ See :meth:`SingleChainMCMC.current_log_pdf()`. """
+        return self._current_log_pdf
```
MichaelClerx: Hey @lorcandelaney 

There's something worth thinking about whenever you write these methods, and that is: Will this make it easier for users to (accidentally) break the integrity of my class? I.E. by changing something, and accidentally creating some invalid state.

In this case, because you're returning an array, I can do this:
```
x = sampler.current_log_pdf()
x[2] = 12
```
which updates your internal state.
So its usually better to do ```return np.copy(self._current_log_pdf)```, or to make the object read-only (if you don't need to modify it yourself). That doesn't _prevent_ the user from unsetting the read only flag and modifying it, but at least they'll be doing it on purpose.
Not that for e.g. ```x = 5; return x``` this isn't a problem, because you can't modify an integer. (You could write ```x = method.get_x(); x = 12```, but that would change your local variable to point at a new integer 12, without affecting the x in the class). (Strings are similar, as they're defined as _immutable_: you can't change a string, doing e.g. ```x = '123'; x += '4'``` creates a new string '1234' and points x at it)

lorcandelaney:
@MichaelClerx @ben18785 @martinjrobins Hi all, I hope you had a great weekend!
First of all, thanks a lot Michael for your comments and changes, they have been extremely helpful. I have made the last two small changes to the code and the checks have passed. Shall we proceed with merging? (Small correction: in commit 0832b2b the message includes the word "setters" but this should have been "getters")

lorcandelaney:
> This all looks excellent, thanks @lorcandelaney! I'm happy to merge

@martinjrobins Great, thanks a lot Martin! Looking forward to merging the branch and working on the next methods!

lorcandelaney:
@martinjrobins @ben18785 @MichaelClerx Unfortunately, I am not authorised to merge the branch. Do you have the authorisation by any chance?

ben18785:
I don't but have asked Martin about this -- hopefully he should pick it up.

martinjrobins:
done! :)

lorcandelaney:
Thanks a lot!!!

ben18785:
Thanks Martin! Congrats Lorcan -- your persistence with this has paid off!

PR #884, opened 2019-07-29
stochastic degradation toy model

danielfridman98:
I've coded up the stochastic degradation toy model (Radek Erban et al., 2007) which will be a useful toy model for ABC. I've uploaded the Jupyter notebook file into examples under the name toy-model-stochastic-degradation (Erban et al., 2007)


pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,74 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import numpy as np
+import pints
+
+from . import ToyModel
+
+
+class StochasticDecayModel(pints.ForwardModelS1, ToyModel):
```
chonlei: @danielfridman98 Actually, can you change it to `pints.ForwardModel` here instead.

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,74 @@
...
+            time.append(t)
+            A = A - 1
+            mol_conc.append(A)
+
+
+        return mol_conc, time
+
+    def suggested_parameter(self):
+        """ See :meth:`pints.toy.ToyModel.suggested_parameters()`. """
+
```
chonlei: Remove line here -- remember to check coding style with `flake8`! (I know you haven't done that yet, but just put a reminder here for you.)

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,74 @@
...
+        while A > 0:
+            r = np.random.uniform(0, 1)
+            tao = (1 / (A * k)) * np.log(1 / r)
+            t += tao
+            time.append(t)
+            A = A - 1
+            mol_conc.append(A)
+
+
+        return mol_conc, time
```
chonlei: This should *not* return `time` here, as it should be determined by the input argument (see above comment).

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,74 @@
...
+        super(StochasticDecayModel, self).__init__()
+        self._n0 = float(initial_concentration)
+        if self._n0 <= 0:
+            raise ValueError('Initial concentration must be positive.')
+
+    def n_parameters(self):
+        """ See :meth:`pints.ForwardModel.n_parameters()`. """
+        return 1
+
+    def simulate(self, parameter):
```
chonlei: This should include `times` as the third argument (i.e. after `parameter`).

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,74 @@
...
+class StochasticDecayModel(pints.ForwardModelS1, ToyModel):
+
+    """
+    Stochastic decay model of a single chemical reaction [1].
+
+    .. math::
+
+
+    Has one parameter: Rate constant :math:`k`.
+    The initial concentration :math:`A(0) = n_0` can be set using the
```
chonlei: Can you add the equations here to make this a bit more self-explanatory (i.e. equations that explain `k`, `A(t)` etc.).

pints/toy/_stochastic_decay_model.py
chonlei: We want to do it in the way that we discussed, that is to have `times` (an array of time points) to determine the sampling time points for this particular simulation. (And which will miss the _exact_ decay/degradation time, and make the problem a bit harder...)

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,74 @@
...
+        if self._n0 <= 0:
+            raise ValueError('Initial concentration must be positive.')
+
+    def n_parameters(self):
+        """ See :meth:`pints.ForwardModel.n_parameters()`. """
+        return 1
+
+    def simulate(self, parameter):
+        """ See :meth:`pints.ForwardModel.simulate()`. """
+        return self._simulate(parameter)
```
chonlei: Remove this `_simulate`, you don't have to wrap it in this toy model. Just put all the code in `_simulate` here.

danielfridman98:
Just updated my stochastic degradation toy model, including all of Chon's comments as well as additional changes. I've written up an example notebook as well for the stochastic degradation model.

pints/toy/_stochastic_decay_model.py
ben18785: Yep, times please. Thanks!

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,69 @@
...
+            raise ValueError('Initial concentration must be positive.')
+
+    def n_parameters(self):
+        """ See :meth:`pints.ForwardModel.n_parameters()`. """
+        return 1
+
+    def simulate(self, parameter):
+        """ See :meth:`pints.ForwardModel.simulate()`. """
+        if parameter <= 0:
+            raise ValueError('rate constant must be postive')
```
ben18785: Can we make consistent with other ValueError messages, so 'Rate constant must be positive.'?

pints/toy/_stochastic_decay_model.py
ben18785: Yep, so we want to interpolate mol_conc/mol_count so that we obtain the values at the given times. A word of caution, be careful when you do this interpolation as it's easy to obtain mol_count(t)-1 rather than mol_count(t) -- I've made that mistake a few times! To this end, I recommend writing a unit test that checks that the interpolation is returning the correct counts at each time.

pints/toy/_stochastic_decay_model.py
```diff
@@ -0,0 +1,69 @@
...
+
+        A = self._n0
+        k = float(parameter)
+
+        t = 0
+        mol_conc = []
```
ben18785: Can we change the name of mol_conc to mol_count? Seeing as we are dealing with a integer variable rather than a concentration.

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,90 @@
+#
+# Logistic toy model.
```
ben18785: Name update please

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,90 @@
...
+from scipy.interpolate import interp1d
+import pints
+
+from . import ToyModel
+
+
+class StochasticDegradationModel(pints.ForwardModel, ToyModel):
+
+    """
+    Stochastic decay model of a single chemical reaction [1].
```
ben18785: Ok, can we have a bit more information here. For example, reproducing the A -> phi reaction equation from the paper.

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,90 @@
...
+            raise ValueError('Initial concentration cannot be negative.')
+
+    def n_parameters(self):
+        """ See :meth:`pints.ForwardModel.n_parameters()`. """
+        return 1
+
+    def simulate(self, parameters, times):
+        """ See :meth:`pints.ForwardModel.simulate()`. """
+        if parameters <= 0:
+            raise ValueError('rate constant must be positive')
```
ben18785: Make value error consistent with others in terms of capitalisation and full stops.

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,90 @@
...
+        k = np.array([float(parameters)])
+        times = np.asarray(times)
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+        if self._n0 == 0:
+            return np.zeros(times.shape)
+        a = np.array([float(self._n0)])
+
+        t = 0
+        mol_conc = [a]
```
ben18785: Rename mol_conc to mol_count. Also, I suggest making mol_conc/count a class property. The reason for this is to allow us to write unit tests that can check that the interpolation is working as desired.

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,90 @@
...
+            r = np.random.uniform(0, 1)
+            tao = ((1 / (a * k)) * np.log(1 / r))[0]
+            t += tao
+            time.append(t)
+            a = a - 1
+            mol_conc.append(a[0])
+
+        # Interpolate as step function, decreasing mol_conc by 1 at each
+        # reaction time point
+        f1 = interp1d(time, mol_conc, kind='previous')
```
ben18785: Cool. I suggest writing unit tests to check that the correct molecular concentration is correctly being interpolated at each time step. (It's easy to mistakenly have mol_count-1 rather than mol_count without noticing -- I've done it lots!)

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # Compute concentration ('a') values at given time points using f1
+        # at any time beyond the last reaction, concentration = 0
+        values = f1(times[np.where(times <= max(time))])
+        zero_vector = np.zeros(len(times[np.where(times > max(time))]))
+        values = np.concatenate((values, zero_vector))
+
```
ben18785: As Chon suggested when we met, I really think it'd be a good idea to have extra functions for this model to return useful things. For example, a function which returns the deterministic mean and variance at a given point in time.

I am less sure how to handle the exact posterior here; whether it's best to have a separate toy distribution (rather than toy model) which returns the likelihood of the data or whether to include it here? For clarity, what this function would do is take a series of data - times and values - and return their probability given the rate parameter k. To convert this into a Bayesian problem, we would need to combine this with a prior on k.

pints/toy/_stochastic_degradation_model.py
chonlei: Ben I guess what you mean here (for Daniel to understand) is to have `self.mol_count` to keep track on it?

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,124 @@
...
+        return values
+
+    def StochasticSimulationAlgorithm(self, parameters):
```
MichaelClerx: I'd change this to:

```def _simulate_stochasticly(self, rate):```

Couple of reasons:
- starting with a capital letter is a convention used for class names, so not appropriate here
- similarly, in PINTS we've gone with_this_notation to indicate variables and methods/functions
- starting with an underscore indicates that something is a private method, not to be called directly by people using the toy model.
- you expect a single parameter, the rate, here, so no need for the full parameters vector
- starting with an underscore indic

File: pints/toy/_stochastic_degradation_model.py [hunk 16]
```diff
@@ -0,0 +1,124 @@
...
+        t = 0
+        if parameters <= 0:
+            raise ValueError('Rate constant must be positive.')
+        k = np.array([float(parameters)])
+
+        a = np.array([float(self._n0)])
```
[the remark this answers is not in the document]
MichaelClerx: If you stop using arrays, you can probably also lose the cast to float :-)
(The `float(x)` line is called "casting" x to the type float)

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,124 @@
...
+        mol_count = [a[0]]
+        time = [t]
+
+        # Run stochastic degradation algorithm, calculating time until next
+        # reaction and decreasing concentration by 1 at that time
+        while a[0] > 0:
+            r = np.random.uniform(0, 1)
+            tao = ((1 / (a * k)) * np.log(1 / r))[0]
+            t += tao
```
MichaelClerx: If you like, this could be shortened to

```t += 1 / (a * k) * np.log(1 / r)```

so with fewer brackets and with a and k just being floats!
or even 

`t += 1 / (a * k) np.log(1 / r)` --> `t += np.log(1 / r) / (a * k)` --> `t+= -np.log(r) / (a * k)`

File: pints/toy/_stochastic_degradation_model.py [hunk 18]
```diff
@@ -0,0 +1,124 @@
...
+            time.append(t)
+            a[0] = a[0] - 1
+            mol_count.append(a[0])
+
+        return time, mol_count
+
+    def DeterministicMean(self, parameters, times):
```
MichaelClerx: Naming again :-)

Also, not sure this needs a `parameters` vector instead of a `rate` float ? Then again, might be convenient for the user, I don't know!

If this method _is_ public, it needs a docstring to explain how to use it, e.g.

```
def deterministic_mean(self, parameters, times):
    """
    One line description of what it does

    More info if needed.
    """
```

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,124 @@
...
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+
+        mean = self._n0*np.exp(-k*times)
+
+        return mean
+
+    def suggested_parameter(self):
+        """ See :meth:`pints.toy.ToyModel.suggested_parameters()`. """
+        return 0.1
```
MichaelClerx: This should be an array

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,124 @@
...
+
+        times = np.asarray(times)
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+
+        mean = self._n0*np.exp(-k*times)
+
+        return mean
+
+    def suggested_parameter(self):
```
MichaelClerx: missing an `s` here in `suggested_parameters`

(Its true it only returns one, but lets stick to the common interface!)

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,124 @@
...
+    def DeterministicMean(self, parameters, times):
+        if parameters <= 0:
+            raise ValueError('Rate constant must be positive.')
+        k = parameters
+
+        times = np.asarray(times)
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+
+        mean = self._n0*np.exp(-k*times)
```
MichaelClerx: The style checking will complain unless you put spaces in, like this: ```mean = self._n0 * np.exp(-k * times)```

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,201 @@
+{
```
MichaelClerx: Notebook looks lovely, but some minor cleaning up to be done (2 empty cells and one equation that isn't rendering)

danielfridman98:
Hi Michael, thanks for your feedback!! I'm fixing the issues you pointed
out.
Also, Just a question regarding the "stochastic_simulation_algorithm" or
"simulate_stochastically" function, you suggested making this private, but
I was initially thinking that I would make this a public function that the
user could call if they wanted to see the exact times of the reaction
according to our algorithm. The reason for this would be that the step
function the user gets from our simulate function isn't the exact step
function, but rather an approximation based on the number of molecules at
each of their measured time points.
Do you still think I should make it private?

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Compute concentration values at given time points using f1
+        # at any time beyond the last reaction, concentration = 0
+        values = self._interp_func(times[np.where(times <= max(time))])
+        zero_vector = np.zeros(len(times[np.where(times > max(time))]))
+        values = np.concatenate((values, zero_vector))
+
+        return values
+
+    def simulate_stochastically(self, parameters):
```
ben18785: I would not have a separate function `simulate_stochastically` as it makes it seem like the function `simulate` is deterministic. Rather, I would include the code for `simulate_stochastically` in the function `simulate` and have mol_count as a class member attribute: so call it self._mol_count. This means that you can still test its values in unit testing to check that interpolation is working as desired.

ben18785: I would not have a separate function `simulate_stochastically` as it makes it seem like the function `simulate` is deterministic. Rather, I would include the code for `simulate_stochastically` in the function `simulate` and have mol_count as a class member attribute: so call it self._mol_count. This means that you can still test its values in unit testing to check that interpolation is working as desired.

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def deterministic_mean(self, parameters, times):
+        """ Calculates deterministic mean of infinitely many stochastic
+        simulations, which follows :math:: n0*exp(-kt)"""
+        if len(parameters) != self.n_parameters():
+            raise ValueError('This model should have only 1 parameter.')
+        k = parameters[0]
+
+        if k <= 0:
+            raise ValueError('Rate constant must be positive.')
```
ben18785: From Mathematica (see email), we can calculate the variance and it turns out to me E^(-2 k t) (-1 + E^(k t)) n0. Can we add a function which returns this?

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,186 @@
...
```
ben18785: Yep, agreed with Michael here. Looking great! A few things (all minor):

- Can you change the code so that the mean  n_0exp[-kt] renders ok in the notebook? For some reason, this is not displaying as math.
- Once the variance is estimated and included as a function, can we overlay this (or, probably better, the standard deviation) on the plot with the mean to show that most trajectories are within this envelope? Alternatively, if this is too messy, you could plot it separately.

pints/toy/_stochastic_degradation_model.py
danielfridman98: Ok, I've put the simulate_stochastically code into the simulate function

pints/toy/_stochastic_degradation_model.py
danielfridman98: Cool! I'll add that in

danielfridman98:
Just uploaded my latest versions of _stochastic_degradation_model, unit tests, and example notebook!

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,190 @@
...
+   ],
+   "source": [
+    "mean = model.deterministic_mean(k, times)\n",
+    "variance = model.deterministic_variance(k, times)\n",
+    "std_dev = np.sqrt(variance)\n",
+    "\n",
+    "plt.plot(times, mean, '-', label = 'mean')\n",
+    "plt.plot(times, mean + std_dev, '--', label = 'standard deviation upper bound')\n",
+    "plt.plot(times, mean - std_dev, '--', label = 'standard deviation lower bound')\n",
+    "plt.legend(loc = 'upper right')\n",
```
chonlei: Can you add labels to both axes here? And make sure all labels are consistent across plots (all three in this notebook).

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,190 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Stochastic degradation model\n",
+    "\n",
+    "This example shows how the stochastic degradation model can be used.\n",
+    "This model describes the stochastic process of a single chemical reaction, in which the concentration of a substance degrades over time as particles react.\n",
+    "The substance degrades starting from an initial concentration, n_0, to 0 following a rate constant, k, according to the following model (Erban et al., 2007):\n",
```
chonlei: Can you make sure all the notation are consistent? E.g. n_0 here, you can do $n_0$ to make it equation-like, and so as k to $k$ etc.

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,190 @@
...
+    "   $$A \\xrightarrow{\\text{k}} \\emptyset$$\n",
+    "\n",
+    "The model is simulated according to the Gillespie stochastic simulation algorithm (Gillespie, 1976)\n",
+    "   1. Sample a random value r from a uniform distribution: r ~ unif(0,1)\n",
```
chonlei: Make `r ~ unif(0,1)` to math mode as well.

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,190 @@
...
+    "model = pints.toy.StochasticDegradationModel(n_0)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "times = np.linspace(0, 100, 100)\n",
```
chonlei: Can you merge cell `In[2]` to `In[4]` into one cell? All seems to me doing one thing.

pints/toy/_stochastic_degradation_model.py
```diff
@@ -0,0 +1,124 @@
...
+    3. Update the molecule count at time t + .. math:: $\tau$ as:
+    .. math:: $A(t + \tau) = A(t)-1$
+    4. Return to step (1) until molecule count reaches 0
+
+    Has one parameter: Rate constant :math:`k`.
+    :math:`r` is a random variable, which is part of the stochastic model
+    The initial concentration :math:`A(0) = n_0` can be set using the
+    (optional) named constructor arg ``initial_concentration``
+
+    [1] Erban et al., 2007
```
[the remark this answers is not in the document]
chonlei: Yes, make it a full citation, e.g. APA. Or anything consistent with the other reference in PINTS.

pints/tests/test_toy_stochastic_degradation_model.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_errors(self):
+        model = pints.toy.StochasticDegradationModel(20)
+        # parameters, times cannot be negative
+        times = np.linspace(0, 100, 101)
+        parameters = -0.1
+        self.assertRaises(ValueError, model.simulate, parameters, times)
+        self.assertRaises(ValueError, model.deterministic_mean, parameters,
+                          times)
+        self.assertRaises(ValueError, model.deterministic_variance, parameters
+                          times)
```
chonlei: Syntax error here!!! `...parameters *,* times)`! Make sure you run all tests locally before you push it here... Then it will tell you the coverage (which we want it to be 100%).

danielfridman98:
Got it! Yes I'll look over all these things either tomorrow if I chance, or
once I'm back in New York. I'll make sure to run the tests locally.

pints/tests/test_toy_stochastic_degradation_model.py
```diff
@@ -0,0 +1,128 @@
...
+    def test_start_with_twenty(self):
+        # Run small simulation
+        model = pints.toy.StochasticDegradationModel(20)
+        times = [0, 1, 2, 100, 1000]
+        parameters = [0.1]
+        values = model.simulate(parameters, times)
+        self.assertEqual(len(values), len(times))
+        self.assertEqual(values[0], 20)
+        self.assertEqual(values[-1], 0)
+
```
MichaelClerx: Maybe add a check like ```self.assertTrue(values[1:] <= values[:-1]``` to check the values are monotonically decreasing

pints/tests/test_toy_stochastic_degradation_model.py
```diff
@@ -0,0 +1,128 @@
...
+    def test_simulate(self):
+        model = pints.toy.StochasticDegradationModel(20)
+        parameters = [0.1]
+        times = np.linspace(0, 100, 101)
+        values = model.simulate(parameters, times)
+
+        # Test output of Gillespie algorithm
```
MichaelClerx: These tests need to go!
A unit test should test that a class does what its supposed to do, *but not how it actually does it*!
So you never check internals / private properties of a class, you only check that the methods return what they ought to return. (It gets a bit trickier if the methods update some internal state of the object, but that's not the case here!)
(And don't solve this by adding a method to return these properties either ;-) )

So all your tests should be tests of the `values` returned

pints/tests/test_toy_stochastic_degradation_model.py
```diff
@@ -0,0 +1,128 @@
...
+        values = model.simulate(parameters, times)
+        self.assertEqual(len(values), len(times))
+        self.assertEqual(values[0], 20)
+        self.assertEqual(values[-1], 0)
+
+    def test_suggested(self):
+        model = pints.toy.StochasticDegradationModel(20)
+        times = model.suggested_times()
+        parameters = model.suggested_parameters()
+        self.assertTrue(np.all(times == np.linspace(0, 100, 101)))
```
MichaelClerx: I think value based tests are overkill here. The model returns some suggested values, what they are is determined in the model, we don't need to re-specify that here.
Instead, just check if they have the correct shape and/or type.
A very easy way to do this is to use *pass them back into the model*. That's what we want really, that `suggested_parameters()` returns something - whatever that is - that can be used as parameters.
So just run a simulation with the suggested times and parameters, if that works all is fine, if that raises an exception the test will fail.

docs/source/toy/index.rst
```diff
@@ -34,4 +34,6 @@ Some toy classes provide extra functionality defined in the
     simple_egg_box_logpdf
     toy_classes
     twisted_gaussian_logpdf
+    stochastic_degradation_model
```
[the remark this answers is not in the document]
ben18785: Done

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,185 @@
...
+   "source": [
+    "# Stochastic degradation model\n",
+    "\n",
+    "This example shows how the stochastic degradation model can be used.\n",
+    "This model describes the stochastic process of a single chemical reaction, in which the concentration of a substance degrades over time as particles react.\n",
+    "The substance degrades starting from an initial concentration, $n_0$, to 0 following a rate constant, $k$, according to the following model (Erban et al., 2007):\n",
+    "   $$A \\xrightarrow{\\text{k}} \\emptyset$$\n",
+    "\n",
+    "The model is simulated according to the Gillespie stochastic simulation algorithm (Gillespie, 1976)\n",
+    "   1. Sample a random value r from a uniform distribution: $r \\sim uniform(0,1)$\n",
```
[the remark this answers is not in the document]
ben18785: changed

examples/toy-model-stochastic-degradation.ipynb
```diff
@@ -0,0 +1,185 @@
...
...um3MZxVcexuHEx33zhm9vl1xTXUJZXxtLmpdy26DYAPPHsNjoeM0bOoCK/gqbOJt6qf6s7PeSEcMUl7dtLhcNumNKc0u7/my3ltpw7GhMdw+kjT8cTD9dxccXFczzyIyHCrse0EVMZUTp86/9fcbpHHLz44POYNmJK9+c64hByQlRF7ZN3rz76W7QlL8MVu27HcQg7YYYWFgLwy8/fRMqkupd1xcV1XHI9eznwI6ffv9PzXLOH3r/DPICHq+7Zaf4Dw+/caf791b/Zaf6Emlk7zT921I07yR3AlDE/2Onyu0N/dDF52JPUnwfWYk9Sf8kY826PMl8HDu5xkvqLxph/3dW6tYtp32GMoTPVSVuiDddxGZg7EICX6l4ilozRmeokno4TT8cZWTKSYyqPwTc+Ny+42f769VPdv4SPqTyG6TXT6Ux18v1Xvo9vtj5XMLV6KtOqp7E5vpkfz/sxTvDrURAccZheM53JlZNp6mzi9rdu/2jnFezkThx2IoeUHUJjZyNP1j7ZvePcsgOeMHgCw4uG09TZxIKNC7baeXuOx+jS0ZTmlNKWaGNd+zpCToiQE8Jz7I66NFJKyA2R9tP4+Hji7fXnddTeY686BwEgIicDv8Be5nq3MeY6EfkhsMAY85SI5AD3A+OxRw7nGGOW72q9GiCUUuqT2dvOQWCMeQZ4Zpu07/d43wXM3NP1Ukop9ZG958J0pZRSexUNEEoppTLSAKGUUiojDRBKKaUy0gChlFIqIw0QSimlMtIAoZRSKqOsGjBIRBqAVb1cfCCw04cBZiHd5uy3r20v6DZ/UsONMWWZMrIqQHwaIrJgR3cTZivd5uy3r20v6Db3Je1iUkoplZEGCKWUUhlpgPjIHf1dgX6g25z99rXtBd3mPqPnIJRSSmWkRxBKKaUy0gChlFIqo30+QIjINBFZKiK1IvLt/q7P7iAiVSIyV0SWiMi7InJFkB4Vkb+JyIfBa2l/17WviYgrIgtF5OlgvkZE5gfb/IiIhHe1js8SESkRkcdE5P2gvSdmezuLyJXB93qxiDwkIjnZ1s4icreI1IvI4h5pGdtVrF8G+7S3ReSw3n7uPh0gRMQFfgVMBw4EzhWRA/u3VrtFCviWMeYA4Gjg68F2fhuYY4wZBcwJ5rPNFcCSHvM3Aj8PtrkZ+Gq/1Gr3uRV41hizPzAOu+1Z284iUgn8F3
```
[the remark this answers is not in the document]
ben18785: changed

pints/tests/test_toy_stochastic_degradation_model.py
ben18785: Done

pints/tests/test_toy_stochastic_degradation_model.py
ben18785: Done

pints/tests/test_toy_stochastic_degradation_model.py
ben18785: Done

pints/tests/test_toy_stochastic_degradation_model.py
MichaelClerx: What would happen to `values` if it went wrong? Can we check for that?

ben18785: Not without knowing the internals, no.

ben18785: Actually I've figured a work-round. Give me a few mins

ben18785: Actually, no I haven't -- can we keep this as is please?

MichaelClerx: Why don't you seed the random generator, print the output, check it manually, and then hardcode those values in the test?

ben18785: I've created separate functions that do the interpolation and return the raw values (and added tests for these). Let us know what you think @MichaelClerx

MichaelClerx: I still don't quite understand why that's needed. Will users benefit from having these methods as part of the API? If so, then great, if not then this is not the way to go.
But if we can't tell whether its working from the output of simulate(), then why do we care?

(I found this article https://enterprisecraftsmanship.com/2017/10/23/unit-testing-private-methods/ that says the stuff I'm saying a bit better, even if it doesn't give that many arguments)

PR #891, opened 2019-08-05
Added CMA-ES and the ForwardModel

naunauyoh:


MichaelClerx:
Thanks Arnaud,

Could you put these in separate PRs please? I think we already have one for the
sgd model?




On Mon, 05 Aug 2019 03:27:35 -0700 Arnaud Massenet <notifications@github.com>
wrote

Issue #500, opened 2018-09-10
Implement (with a warning) sensitivity methods using finite differences


ben18785:
For sensitivities, this requires 2x number of parameters per iteration if we use central differences; 1x the number of parameters if we use forward/backward differences.

For 2nd order sensitivities, more function evaluations are required.

ben18785:
[This](http://www.mathematik.uni-dortmund.de/~kuzmin/cfdintro/lecture4.pdf) is a good reference on finite differences.

Note that I don't expect this approach to be as computationally efficient as using automatic differentiation or calculating the sensitivities outside of Pints. It will, however, allow users to quantify, in terms of the number of effective samples per _iteration_ not _computational time_ whether it is worth solving for sensitivities.

MichaelClerx:
I have some serious doubts about this! 

Apart from my gut feeling (backed up by some bad experiences) that they're a terrible idea for serious problems, there's the question of where do we stop with Pints.
Yes it's something we _could_ add, but I'm much happier saying "pints is for optimising / sampling" and leaving the rest for the users and other specialist packages. (See #501 )

MichaelClerx:
Example of how this could be implemented in a wrapper: https://github.com/pints-team/pints/pull/892

PR #892, opened 2019-08-07
ForwardModel wrapper to approximate sensitivities

naunauyoh:
See #500


MichaelClerx:
I like this idea (see also #500), but don't think this is the best algorithm for it

MichaelClerx:
Keeping this branch for reference

PR #893, opened 2019-08-07
Stochastic gradient descent optimiser

naunauyoh:


MichaelClerx:
See also #930

MichaelClerx:
I think this PR provides:

- a simple gradient descent optimiser which includes gradient guessing

so we can replace it by

- a simple gradient descent optimiser #1104 
- a ForwardModelS1 that wraps around a ForwardModel and guesses gradients #500 #892

and so close this.

Any objections @ben18785 @naunauyoh @martinjrobins @chonlei ?

PR #894, opened 2019-08-07
PINTS implementation of CMA-ES optimiser

### LINKED ISSUE #415: Add bare bones CMAES
In some cases, CMAES beats XNES by a very wide margin. It'd be interesting to see if this is due to the core method, or due to all the tricks in the python `cma` module.

naunauyoh:
Fixes #415

CMA-ES without using the `cma` module


MichaelClerx:
Hi Arnaud,
This is a great start, but still some things to do:

- Add documentation (RST file)
- Add an example notebook
- Add a test
- Update the coding style to match the PEP8 guidelines

See CONTRIBUTING.md and the presentations for details!

MichaelClerx:
Hey all!
I've tidied up the code by @naunauyoh and added the missing bits (tests, documentation, notebook example). Could you have a look and see if you think this is ready to go in?

@naunauyoh comments from you are also welcome! Thanks again for coding this up

chonlei:
Thanks Michael! I will have a look today!

MichaelClerx:
@DavAug

MichaelClerx:
@fcooper8472 here's the thesis I'd like to cite for this method https://www.cs.ox.ac.uk/msctheses/1724/

fcooper8472:
> @fcooper8472 here's the thesis I'd like to cite for this method https://www.cs.ox.ac.uk/msctheses/1724/

Can't see any way of getting public access to that, and in any case there doesn't seem to be a DOI or similar...

@MichaelClerx best bet is probably to get Arnaud to upload it to [zenodo](https://zenodo.org/) so it has a permanent public home and a DOI.  Maybe suggest a change to the front page too so it has his name on!  Think that would be possible?

pints/_optimisers/_cmaes_bare.py
```diff
@@ -0,0 +1,379 @@
...
+        if isinstance(self._boundaries, pints.RectangularBoundaries):
+            self._boundary_transform = pints.TriangleWaveTransform(
+                self._boundaries)
+        elif self._boundaries is not None:
+            self._manual_boundaries = True
+
+        # Parent generation population size
+        # The parameter parent_pop_size is the mu in the papers. It represents
+        # the size of a parent population used to update our paramters.
+        # Not sure if the limitation to dim is good
```
DavAug, line 173: This comment is not clear to me. What is `dim` and where is it used to limit the `parent_pop_size`

DavAug: In line 176 it seems as if the `parent_pop_size` is simply always the smaller half of `population_size`

pints/tests/test_opt_cmaes_bare.py
```diff
@@ -0,0 +1,146 @@
...
+        opt.set_log_to_screen(debug)
+        found_parameters, found_solution = opt.run()
+        self.assertTrue(found_solution < 1e-3)
+
+    def test_bounded(self):
+        # Runs an optimisation with boundaries.
+        r, x, s, b = self.problem()
+
+        # Rectangular boundaries
+        b = pints.RectangularBoundaries([-1, -1], [1, 1])
```
DavAug, line 57: Why define boundaries again, when `b` has been defined as those boundaries already in `problem`?

DavAug: Ok, probably to make the difference to the `CircularBoundaries` more explicit.

Issue #735, opened 2019-02-21
Maths in docstrings


fcooper8472:
At the moment all maths in docstrings requires the escaping of each backslash character, which ends up looking messy.  For instance, here's the start of one docstring:

```python
"""
    Defines a gamma (log) prior with given shape parameter ``a`` and rate
    parameter ``b``, with pdf

    .. math::
        f(x|a,b)=\\frac{b^a x^{a-1} e^{-bx}}{\\mathrm{\\Gamma}(a)}
...
```

If you make the docstring a raw string literal by adding `r` to the start of it, the escape characters are no longer needed.  This makes the formulas look a lot nicer, and is how (for instance) numpy and scipy do their docstrings:

```python
r"""
    Defines a gamma (log) prior with given shape parameter ``a`` and rate
    parameter ``b``, with pdf

    .. math::
        f(x|a,b)=\frac{b^a x^{a-1} e^{-bx}}{\mathrm{\Gamma}(a)}
...
```

I think we should adopt this going forward and, eventually, rework the existing docstrings.

ben18785:
That’s good to know. Thanks Fergus!

MichaelClerx:
Anyone feel like going over the docs for this? Or making a list of docs that need adapting?

MichaelClerx:
Nobody?

MichaelClerx:
It seems that if we use raw strings, we can drop one of our exceptions from the flake8 configuration too:

https://github.com/ModellingWebLab/project_issues/issues/43#issuecomment-521605747

MichaelClerx:
In fact, @fcooper8472 do you think we should update our flake8 config to complain about bad escape sequences again? That would maybe help us identify cases where we need to use a raw string

fcooper8472:
I didn't know we had explicitly turned it off!  I'd say, yes, let's not allow invalid escape sequences, and any docstrings that contain maths can be `r""""""` strings.

PR #901, opened 2019-08-15
Let flake8 check for invalid escape sequences again

MichaelClerx:
because latex docstrings should be in raw strings.

See #735


MichaelClerx:
@fcooper8472 could you have a look at the updates please?

PR #911, opened 2019-08-26
Issue 501 sensitivity

### LINKED ISSUE #501: Include examples involving sensitivity calculated by automatic differentiation
as well as more traditional Jacobian matrix based approaches. It'll be interesting to see what works more efficiently. Perhaps have a look at [this](https://pythonhosted.org/algopy/) program.

martinjrobins:
Fixes #501 

autograd seems very slow, so I've just made an example showing how to use it, and showing how slow it is for the fitzhugh nagumo model


MichaelClerx:
I'm guessing this would need to be added `.slow-books` as well

PR #924, opened 2019-09-12
Adaptive base class

ben18785:
Proposed structure for adaptive covariance methods:

AdaptiveCovarianceMCMC (abstract class)

- GlobalAdaptiveCovarianceMCMC (abstract class)
  - RemiACMCMC 
  - HaarioACMCMC
  - RaoBlackWellACMCMC
  - DRAMACMCMC
  - ComponentwiseACMCMC
  - DualAvgMCMC (see Sanmitra's email and NUTS [paper](https://arxiv.org/pdf/1111.4246.pdf))

- LocalAdaptiveCovarianceMCMC (abstract class)
  - LocalACMCMC
  - others that I've not remembered

At the moment, I have just implemented the global side with RemiACMCMC and HaarioACMCM to give an idea of how this might work.


pints/_mcmc/_global_ac.py
```diff
@@ -0,0 +1,85 @@
...
+#  For licensing information, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class GlobalACMCMC(pints.GlobalAdaptiveCovarianceMCMC):
```
MichaelClerx: Could we use a different name for this one, to distinguish it from the wider class of global methods? I'd be fine with using some nickname like we do with the Remi method

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -154,67 +136,52 @@ def tell(self, fx):
...
-        return self._current
+    def _update_sigma(self):
+        """
+        Updates the covariance matrix used to generate proposals.
+        Note that this default is overidden in some of the methods
+        """
+        raise NotImplementedError
 
-    def replace(self, current, current_log_pdf, proposed=None):
+    def replace(self, x, fx):
```
MichaelClerx: I think we needed to be able to update the proposal to make SMC work.
(If we decide to drop SMC I suppose this is fine?)

MichaelClerx: Either way, the replace method implementation can wait until after this is done, in my opinion :-|

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -154,67 +136,52 @@ def tell(self, fx):
             # Clear proposal
             self._proposed = None
 
+            # Set r to zero
+            self._r = float('-Inf')
```
MichaelClerx: r needs to be defined in the constructor too

pints/_mcmc/_global_adaptive_covariance.py
```diff
@@ -0,0 +1,68 @@
...
+    acceptance rate.
+
+    *Extends:* :class:`AdaptiveCovarianceMCMC`
+    """
+    def __init__(self, x0, sigma0=None):
+        super(GlobalAdaptiveCovarianceMCMC, self).__init__(x0, sigma0)
+
+    def ask(self):
+        """ See :meth:`SingleChainMCMC.ask()`. """
+        super(GlobalAdaptiveCovarianceMCMC, self).ask()
```
MichaelClerx: That can't be right :D

MichaelClerx: There doesn't seem to be anything in that method?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -96,6 +69,8 @@ def _initialise(self):
         # Adaptation
         self._loga = 0
```
MichaelClerx: Is `_loga` still used?

Now that these are used by multiple methods, we should probably add a comment with some explanation above each

pints/_mcmc/_global_ac.py
ben18785: Agreed. Changed it to 'SimpleACMCMC'

pints/_mcmc/_adaptive_covariance.py
ben18785: Yep, done.

pints/_mcmc/_global_adaptive_covariance.py
ben18785: There is an initialise call in the method above that method (it also calls super, to reach AdaptiveCovarianceMCMC.ask()).

pints/_mcmc/_adaptive_covariance.py
ben18785: Yep, you're right -- this was redundant. Have updated now.

pints/_mcmc/_adaptive_covariance.py
ben18785: I've left it in for now. Can discuss later...

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -1,8 +1,8 @@
 #
-# Adaptive covariance MCMC method
+# Base class for Adaptive covariance MCMC methods
 #
 # This file is part of PINTS.
-#  Copyright (c) 2017-2019, University of Oxford.
+#  Copyright (c) 2017-2018, University of Oxford.
```
martinjrobins: why 2019?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -39,6 +31,7 @@ def __init__(self, x0, sigma0=None):
         self._current = None
         self._current_log_pdf = None
         self._proposed = None
+        self._r = None
```
martinjrobins: what is r? variable name more descriptive or an explanatory comment

pints/_mcmc/_adaptive_covariance.py
ben18785: Good spot!

ben18785:
> This looks good, thanks @ben18785. My one comment is that it is difficult to know what are the class variables in `AdaptiveCovarianceMCMC`, since many of them are only created in `_initialise` and are not defined in the constructor.

Yep, good point. Have moved these into the constructor.

pints/_mcmc/_adaptive_covariance.py
ben18785: Have changed this.

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -58,21 +65,6 @@ def ask(self):
         if not self._running:
             self._initialise()
 
-        # Propose new point
```
MichaelClerx: So... the base class has an ask(), and the implementing classes have to override it, before calling the base class ask() , doing their own stuff and actually returning.

That's OK. But wondering if it would be tidier to have the base class call a 2nd method, e.g. _ask() or _propose() that the base class defines but sticks a `raise NotImplementedError` in. That way we have made it more explicit that this is something implementing classes need to do?

Not sure! Thoughts @martinjrobins @ben18785 ?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -83,26 +75,32 @@ def _initialise(self):
...
+        """
+        Returns ``eta`` which controls the rate of adaptation decay
+        ``adaptations**(-eta)``, where ``eta > 0`` to ensure asymptotic
+        ergodicity.
+        """
+        return self._eta
 
-        # Update sampler state
-        self._running = True
+    def set_eta(self, eta):
```
MichaelClerx: Move down to make alphabetical

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -58,21 +65,6 @@ def ask(self):
         if not self._running:
             self._initialise()
```
MichaelClerx: Can the code from _initialise can't just go in the ask() method, instead of a separate function?

MichaelClerx: Tell has some initialisation code as well

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -154,44 +149,29 @@ def tell(self, fx):
...
-                self._current - self._mu, (self._n_parameters, 1))
-            self._sigma = (
-                (1 - gamma) * self._sigma + gamma * np.dot(dsigm, dsigm.T))
+        self._log_acceptance_ratio = fx - self._current_log_pdf
 
-        # Update acceptance rate (only used for output!)
-        self._acceptance = ((self._iterations * self._acceptance + accepted) /
-                            (self._iterations + 1))
-
-        # Increase iteration count
```
MichaelClerx: Iteration count was part of the base class tell() method for first point, so should be same for remaining points?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -154,44 +149,29 @@ def tell(self, fx):
...
-            # Update mu, log acceptance rate, and covariance matrix
-            self._mu = (1 - gamma) * self._mu + gamma * self._current
-            self._loga += gamma * (accepted - self._target_acceptance)
-            dsigm = np.reshape(
-                self._current - self._mu, (self._n_parameters, 1))
-            self._sigma = (
-                (1 - gamma) * self._sigma + gamma * np.dot(dsigm, dsigm.T))
+        self._log_acceptance_ratio = fx - self._current_log_pdf
 
-        # Update acceptance rate (only used for output!)
```
MichaelClerx: Again, can we write this so that the base class by itself will generate an error if used? So have some kind of _tell method that the base class calls and the subclasses implement?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -154,44 +149,29 @@ def tell(self, fx):
...
-        self._acceptance = ((self._iterations * self._acceptance + accepted) /
-                            (self._iterations + 1))
-
-        # Increase iteration count
-        self._iterations += 1
+    def _update_mu(self):
+        """
+        Updates the current running mean used to calculate the sample
+        covariance matrix of proposals. Note that this default is overidden in
```
MichaelClerx: Which default?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -154,44 +149,29 @@ def tell(self, fx):
...
+        some of the methods
+        """
+        raise NotImplementedError
 
-        # Return new point for chain
-        return self._current
+    def _update_sigma(self):
+        """
+        Updates the covariance matrix used to generate proposals.
+        Note that this default is overidden in some of the methods
```
MichaelClerx: See above

pints/_mcmc/_global_adaptive_covariance.py
```diff
@@ -0,0 +1,62 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class GlobalAdaptiveCovarianceMCMC(pints.AdaptiveCovarianceMCMC):
+    """
+    Base class for single chain MCMC methods that globally adapt a
```
MichaelClerx: A lot of these files seem to do line breaks arbitrarily instead of at 79 chars

pints/_mcmc/_haario_ac.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def name(self):
+        """ See :meth:`pints.MCMCSampler.name()`. """
+        return 'Haario adaptive covariance MCMC'
+
+    def n_hyper_parameters(self):
+        """ See :meth:`TunableMethod.n_hyper_parameters()`. """
+        return 1
+
+    def set_hyper_parameters(self, x):
```
MichaelClerx: This is the same in both, and eta belongs to the base class (Adaptive), so maybe this method should go there?

pints/_mcmc/_global_adaptive_covariance.py
```diff
@@ -0,0 +1,62 @@
...
+    proposal covariance matrix when running, in order to control the
+    acceptance rate.
+
+    *Extends:* :class:`AdaptiveCovarianceMCMC`
+    """
+    def tell(self, fx):
```
MichaelClerx: I can't really tell from this code where the implementing classes get to do their thing. Can that happen in some appropriately named method that's called by this class?

E.g. self._adapt_log_lambda ?

pints/_mcmc/_haario_ac.py
```diff
@@ -0,0 +1,103 @@
...
+            self._proposed.setflags(write=False)
+
+        # Return proposed point
+        return self._proposed
+
+    def tell(self, fx):
+        """ See :meth:`pints.AdaptiveCovarianceMCMC.tell()`. """
+        super(HaarioACMCMC, self).tell(fx)
+
+        self._acceptance_prob = (
```
MichaelClerx: Why is this stored?

pints/_mcmc/_haario_ac.py
```diff
@@ -0,0 +1,103 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class HaarioACMCMC(pints.GlobalAdaptiveCovarianceMCMC):
+    """
+    Adaptive Metropolis MCMC, which is algorithm 4 in [1].
```
MichaelClerx: Ref to Haario again. And maybe some words on how this method is different? Not how the maths differs, but maybe what the intuition was or something? Not required if too hard

pints/_mcmc/_remi_ac.py
```diff
@@ -0,0 +1,105 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class RemiACMCMC(pints.GlobalAdaptiveCovarianceMCMC):
+    """
+    Adaptive Metropolis MCMC, which is algorithm in SOM of [1].
```
MichaelClerx: I've never seen the abbrev. SOM before

pints/tests/test_mcmc_adaptive.py
```diff
@@ -60,36 +58,26 @@ def setUpClass(cls):
...
+        self.assertEqual(0.234, mcmc.target_acceptance_rate())
+        self.assertTrue(mcmc.in_initial_phase())
+
+    def test_ask_tell(self):
+
+        # ask only initialises
+        x0 = self.real_parameters * 1.1
+        mcmc = pints.AdaptiveCovarianceMCMC(x0)
+        mcmc.ask()
+        self.assertTrue(mcmc._running)
```
MichaelClerx: Never check private properties.

pints/tests/test_mcmc_adaptive.py
```diff
@@ -60,36 +58,26 @@ def setUpClass(cls):
...
-            if np.all(sample == x):
-                self.assertEqual(mcmc.current_log_pdf(), fx)
-
-        chain = np.array(chain)
-        rate = np.array(rate)
-        self.assertEqual(chain.shape[0], 50)
-        self.assertEqual(chain.shape[1], len(x0))
-        self.assertEqual(rate.shape[0], 100)
+        # tell
+        mcmc._proposed = None
```
MichaelClerx: Again. The unit tester doesn't know about `._proposed`. All unit tests are of the public API. It should be possible to completely strip out all the internals and rewrite the class, without having to change the tests

pints/tests/test_mcmc_adaptive.py
```diff
@@ -60,36 +58,26 @@ def setUpClass(cls):
...
+        self.assertRaises(RuntimeError, mcmc.tell, 0.0)
```
MichaelClerx: Just check that tell raises an error if you call it without a tell. That's what those internals are trying to do.

pints/tests/test_mcmc_haario_ac.py
```diff
@@ -0,0 +1,198 @@
...
+            if np.all(sample == x):
+                self.assertEqual(mcmc.current_log_pdf(), fx)
+
+        chain = np.array(chain)
+        rate = np.array(rate)
+        self.assertEqual(chain.shape[0], 50)
+        self.assertEqual(chain.shape[1], len(x0))
+        self.assertEqual(rate.shape[0], 100)
+
+    def test_replace(self):
```
MichaelClerx: Think this is a method of a base class, and so should go in test for that base class, not here (or in the Remi one)

pints/tests/test_mcmc_haario_ac.py
```diff
@@ -0,0 +1,198 @@
...
+        # Proposal can be changed too
+        mcmc.ask()
+        mcmc.replace([1, 2, 3], 10, [3, 4, 5])
+
+        # New proposal must have correct size
+        self.assertRaisesRegex(
+            ValueError, '`proposed` has the wrong dimensions',
+            mcmc.replace, [1, 2, 3], 3, [3, 4])
+
+    def test_flow(self):
```
MichaelClerx: Not sure how much (if any) of this is changed by Haario/Remi classes, and how much should go in single test for base class

pints/tests/test_mcmc_haario_ac.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # test hyperparameter setters and getters
+        self.assertEqual(mcmc.n_hyper_parameters(), 1)
+        self.assertRaises(ValueError, mcmc.set_hyper_parameters, [-0.1])
+        mcmc.set_hyper_parameters([0.3])
+        self.assertEqual(mcmc.eta(), 0.3)
+
+        self.assertEqual(mcmc.name(), 'Haario adaptive covariance MCMC')
+
+    def test_logging(self):
```
MichaelClerx: Again, might have to go into test for base class

examples/readme-example.ipynb
```diff
@@ -182,7 +182,7 @@
    "source": [
     "# Set up a sampling routine with 3 MCMC chains, all starting from the same point\n",
     "mcmc = pints.MCMCController(\n",
-    "    log_likelihood, 3, [initial_point]*3, method=pints.AdaptiveCovarianceMCMC)\n",
+    "    log_likelihood, 3, [initial_point]*3, method=pints.RemiACMCMC)\n",
```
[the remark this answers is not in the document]
ben18785: Happy to use a different name options are:

- RemiACMCMC
- BardenetACMCMC
- RemiACMC
- BardenetACMC -- my favourite!

@martinjrobins @MichaelClerx -- which of these is yours? If you let me know, I can change it over quickly.

examples/sampling-adaptive-covariance-haario.ipynb
```diff
@@ -0,0 +1,171 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Inference: Haario adaptive covariance MCMC\n",
```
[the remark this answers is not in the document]
ben18785: Yep, good point. Have included this reference in the docstrings too.

examples/sampling-adaptive-covariance-haario.ipynb
```diff
@@ -0,0 +1,171 @@
...
...IpYKWH2dTVhids6lJpQoXLEX4m/2MPWoTh7rC7W\nlxXjtI+nDxkriMce/yW9pXw528JVPNHXwVPSyuUNN7FWThAtuoo3bbqVhw62EfSECfcLzR6bjCfE\nfp+fVTXrYeBXlBr3+6g07OO9N76L//j+TzGeNPUb7mXLwSaahpqwxBANlkLC7SRQYUL0k4BIJbVl\nEY71LaPY6SZtG8I+DzeW1nAgGABj8PjnP7kCTbAWzFStLXuG+1iYoXbqfGjrmlLqdMaYD5zlrdvO\nsnx+DDRD83NukYrT50oumaJVqrAatn5wISJTp7M8blILcM1Hx+fGMrY7kbJlQe0W8IUhEMG36V2w\n634es9zEYdgpZmPtFdwQrCNQsYlXB4/R5zuKFX+O4qDbetFjFzAcrIX6Ssr8N9M/cpCg9wAA6yJv\n4W67j2+lj0xqQSgN++nxN5AqKKYiFOHudCs7eo5TmYkQCV5OrGILv31ZiK8c+hrx4T4sTxmVK3+T\nVckdHOltxQ7bvD2yio5Ukh9EbmLd0D62emJ8zhskE67Bzjo0mEJGwsmxbm8D3ioqnR6KKq+lYNhH\nfeFm7r3md0inBjiWjPNKR4ADx1tprvxjttZ00mcPIiLcvWUrP9qZ5erBnaw0xbSU9btd0QbhzkgD\nmaThZ7bbda0wN26pqe5DvGX7Bso7HodjLVDaiFVUR1nL86wLldE+mCQpNp1l76IWi1D8dSKNNWw3\nq+ls38WbCfPD7HFsx7DSROjzCx0k6S64jKrYYUJei6Phq6gf3k1NYQCfx6LKG+bjpavwWF4oW8WL\nweWUlP0uJYlXuaskyRuTNYQ23Il99BdcPbyMjshNvGGbReexvaT6j5MI1tBQKtx85e9R/9pPeP1k\nlN7wGu7duJ2aUuEnewdYbo7iDD7K7UX1xBMOsXAJvwyVYlnCfXWXEzVp/n6kGANElq1Btn8AXuug\nqHOYrlz3vZBVgsfq5+q1d2Ha2qn2xjhCPx6BUt9yKnyrSYVeAivFtSURytNFXB0oYRcD3O6s4Jee\nFrKh1QxlulmTHMCHxeZ1b4H+E6yKRImMrOa6rR9Brh7/cvp/K34f0/Q0I4ef5cEhYbdfqCz0g9fP\nyrWfIt3VxSp+RVVRAIqCfGDNPdSnT+Bd+27Sle28Ye+3CJZU87/Sw7Rna7h3qIvba+p
```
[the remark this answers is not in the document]
ben18785: Yep, done.

examples/sampling-adaptive-covariance-remi.ipynb
```diff
@@ -0,0 +1,158 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Inference: Remi adaptive covariance MCMC\n",
+    "\n",
+    "This example shows you how to perform Bayesian inference on a time series, using a variant of [Adaptive Covariance MCMC](http://pints.readthedocs.io/en/latest/mcmc_samplers/adaptive_covariance_mcmc_remi.html) detailed in SOM of [1].\n",
```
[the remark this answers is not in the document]
ben18785: Yep good point -- have changed this TLA ("three letter acronym") to be "supplementary materials" instead.

examples/sampling-adaptive-covariance-remi.ipynb
```diff
@@ -0,0 +1,158 @@
...
...byRBx9rPsrm2lBfbBjnq3zM6NM4y\nFvUleU4BmoIK1tUU4AulODjs4njpO7hsxSVYBx8lHO1ljyfCCdcN1HuX47NCbF5QxL7mAcKxJCZQ\nxOLyHHL73TQNuWnNW0Pt/EIO7z1FUf47yBl8FE93gA+t/Cg5r36Pck8uj+bcxP7wo5QXF3JFyS24\nJcTPDzll8evLLLzG5tIF8yl3r6Opb4Dl65bA9nsoDnl5a089zRVv5eY1a9jeOEiqNUbx0mtZ1jrg\nLAydWwHAx5bezv09S/FYTq9kJNVJZe4wmys3cXzgOEF3kCurrzpzseVppglWNqXXvmIa17Ka697s\nVfje7O9fqTnmN8AngK+lvz887a8Y7oQ99ztzmUZYblh9G+SUnnv/plfgyJPO7UCh8/VG+fPgqs+B\nsSdOKEYU1kN+NfQcc75GenKK5ju9XheLXT89vTaTNcHnUbUOuhug68j4nrwVt8CJ/3Z6uUYsfQeb\nai+hbOAEXpeXVSX1Tgn3dBK7tmwtOenewkdea2EoluLSikpCPjc5nhy8Lich8laupOlkLtXxHFaX\nFHPD0it55XgP9WGnzEF9cZAT3TXUhG7jXaUHiQwJlxWsZm+wmo21ZfjcFoPRJMOJFPe/7AwnLfDU\nEAjVk7dqIcXuFES7+fRVH8ZtueFgH0tbXiWcn2BDwTKa5l/G7xsfZdWS23CVrQAR/mDxH9DbcZR5\nBdXk+/KhZBnUtJNTs4n3i02xvxhXuiDK9eUf5pAZoHRFBZ/NGeB3jR2sX7iBijUVox+VhQsRoT/m\nzDkqCnmBIhLBMlyFddT6+wh5clhTk4/VG+MWdxnGAn/5EhbUX4OIcNvS2wCIJVP43FcAcG3NtTT4\nilhsAqxYtAmASxcU4bKE+uIQC0tzuH5pGZZ1+md53ZIy9rX4+J8rP4/LchZHxrKgcB45wN03fZe4\nWAzGXXjcFnl+D3dsuoYlRYu4f1uE7bV3sPaqevyB0wsNX3b1R5y5jsESKtw+KsrKAbiy6ko2b7Dp\nanicp061M7jwg7C8ANr3Qu1mgk3bYd4VcAC8bovLq6+E6iuJRTpYEh/gQJufA83D/Olbnd+pslwf\nj7zWSlHIi9/tYl5xiMprb2dlsAYJLeHKuhRXCfg9byVlG+f9rbmN3twI7h6L5ZW5lOQEec+aRRhj\n
```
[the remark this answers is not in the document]
ben18785: Yep, changed.

pints/_mcmc/_adaptive_covariance.py
ben18785: I like doing it the way we currently are because it avoids second methods and also, to me at least, this seems like a good way to use the inheritance of base classes. @martinjrobins what do you think?

pints/_mcmc/_adaptive_covariance.py
ben18785: Done and in all other AC code too.

pints/_mcmc/_adaptive_covariance.py
ben18785: Done!

pints/_mcmc/_adaptive_covariance.py
ben18785: I've removed this now.

pints/_mcmc/_adaptive_covariance.py
ben18785: This section of the code has been removed now.

pints/_mcmc/_adaptive_covariance.py
ben18785: Good spot!

pints/_mcmc/_adaptive_covariance.py
ben18785: Good spot again!

pints/_mcmc/_global_adaptive_covariance.py
ben18785: Have changed some of the text so that it runs to 79 instead.

pints/_mcmc/_haario_ac.py
ben18785: Yep, good idea. Have changed.

pints/_mcmc/_global_adaptive_covariance.py
ben18785: I could, but feel that, before implementing the other AC methods, I'd be a bit unsure whether self._adapt_log_lambda is needed? Any chance we could wait with this then update later if needed/

pints/_mcmc/_haario_ac.py
ben18785: Good point! Changed.

pints/_mcmc/_haario_ac.py
ben18785: Done.

pints/_mcmc/_remi_ac.py
ben18785: Changed!

pints/tests/test_mcmc_adaptive.py
ben18785: Removed.

pints/tests/test_mcmc_adaptive.py
ben18785: Ok, I have just copied what we were doing in the old tests. I don't see how to test this runtime error otherwise...

ben18785: Saw your comment below -- now changed. Thanks!

pints/tests/test_mcmc_adaptive.py
ben18785: Cool, thanks. Done.

pints/tests/test_mcmc_haario_ac.py
ben18785: Done!

pints/tests/test_mcmc_haario_ac.py
ben18785: The problem is that the base class is just abstract. Would you call Haario or Remi in a test of the base class?

pints/tests/test_mcmc_haario_ac.py
ben18785: Ok. At the moment, it tests that the name of each method is printed though...

examples/readme-example.ipynb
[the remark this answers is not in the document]
MichaelClerx: Happy with BardenetACMC too!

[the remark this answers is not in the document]
MichaelClerx: @mirams any opinion on this? Ben's rewriting the bit of PINTS that has our trusty AdaptiveCovarianceMCMC method in it, because there's several adaptive covariance methods out there. Remi's one will still be the default, but needs a more specific name

pints/_mcmc/_global_adaptive_covariance.py
MichaelClerx: Sure!

pints/tests/test_mcmc_haario_ac.py
MichaelClerx: Yup! It's not super nice, I agree, but I think it's the best compromise. We have similar things already in the tests for the OptimisationController and the tests for the MCMCController (both of which pick a single method and test with that)

pints/tests/test_mcmc_haario_ac.py
MichaelClerx: Good point!
I'm a bit conflicted, but I think the cleanest thing to do here is to think of the base class methods _using_ the inheriting methods. So we can maybe check that the log for 2 different methods has 2 different names in it, but this is a check of the base class, not the inheriting classes.
The unit tests for the inheriting classes, meanwhile, should check that their name() method works

MichaelClerx:
Looking good @ben18785 ! Maybe @martinjrobins can rule on the final issues, we can fix whatever the 2 uncovered lines are, and then get this in?

pints/tests/test_mcmc_haario_ac.py
ben18785: Yep, that's a good idea -- have changed this now so that the daughter class tests check for name and the abstract base class checks everything else (iterations, accept and so on).

pints/tests/test_mcmc_haario_ac.py
ben18785: Yep, you're right. I have changed this now.

examples/readme-example.ipynb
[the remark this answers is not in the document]
mirams: What's the difference between remi and the other adaptive covariance one?

ben18785:
Hi Gary,

The proposed structure is of adaptive methods is something like:

AdaptiveCovarianceMCMC (abstract class)

   -

   GlobalAdaptiveCovarianceMCMC (abstract class)
   - RemiACMCMC
      - HaarioACMCMC
      - RaoBlackWellACMCMC
      - DRAMACMCMC
      - ComponentwiseACMCMC
      - DualAvgMCMC (see Sanmitra's email and NUTS paper)
   -

   LocalAdaptiveCovarianceMCMC (abstract class)
   - LocalACMCMC
      - others that I've not remembered

RemiACMCMC is the old AdaptiveMCMC. Now AdaptiveMCMC is a base class.

mirams:
I've asked Remi!

rbardenet:
Hi everyone. I'm impressed with pints, well done! As for RemiACMCMC vs HaarioACMCMC, they are almost the same algorithm (from what I see in the code of the adaptive-base-class). The only difference is in the adaptation of the stepsize, where RemiACMCMC used the indicator of acceptance at the current iteration, while HaarioACMCMC uses the current running estimate of the acceptance rate instead. Is that right?

If yes, then from a mathematical point of view, the modification is very minor and the proof of convergence is nearly identical. I would thus call both algorithms HaarioACMCMC, with an argument to specify whether the indicator or the estimated rate is used. Default could be the indicator (as in the current RemiACMCMC). I certainly don't deserve my name to be at the same level as Haario's in the code, as I merely helped Ross implement my favourite version of the algorithm from (Haario, Saksman, and Tamminen, Bernoulli 2001), I didn't invent anything.

Out of curiosity, did you observe a significant change in performance between HaarioACMCMC and RemiACMCMC on some problems? If yes, which ones?

ben18785:
Hi Remi,

Glad to hear you like the PINTS idea!

Re: performance, we haven’t run extensive tests of everything yet although are planning to do so in our second paper.

In terms of naming, we toyed with having both algorithms within a single interface (with a switch dictating behaviour) but from a hyperparameter optimisation perspective, it’s easier to have them as separate entities, I think. Somewhat less convincingly, it also boosts overall algorithm count having them separate.

We could call them HaarioACMC1 and HaarioACMC2 (personally, I’d prefer BardenetACMC for one :). What do people think?

Thanks,

Ben

MichaelClerx:
Would `HaarioACMC` and `HaarioBardenetACMC` be acceptable?

pints/_mcmc/_adaptive_covariance.py
martinjrobins: I'm happy with what @ben18785 has done, its a fairly traditional way of using inheritance.

MichaelClerx:
@ben18785 I've made a few updates to the docstrings. If you're OK with them let me know and I'll merge!

MichaelClerx:
Steps to fix:

- [ ] Rename the new `AdaptiveCovarianceMCMC` class to something else
- [ ] Add a new `AdaptiveCovarianceMCMC` that acts like the old one, but raises a deprecation warning. (Also add a comment in the code with the date we deprecated it)
- [ ] Add a tiny test (does it exist), and list it in the docs

I'm happy to do steps 2 and 3 @ben18785 !

ben18785:
These are good points -- do we have to ensure backwards compatibility though? No one really uses PINTS currently, so I wouldn't worry about it. The reason for me being pro staying with AdaptiveCovarianceMCMC is that I think this is the most informative base class name. Also, not sure we want to be beholden to all previous choices we've made in the past.

What do you think?

MichaelClerx:
> No one really uses PINTS currently, so I wouldn't worry about it

67 stars. Also published work and code by me, you, Chon, Martin :-)

Most of that uses AdaptiveCovarianceMCMC somewhere

ben18785:
Ha, fair enough! Still not sure really what to call the new class.

MichaelClerx:
AdaptiveCovarianceMC ?

MichaelClerx:
It's annoying I'll grant you!

ben18785:
> Steps to fix:
> 
> * [x]  Rename the new `AdaptiveCovarianceMCMC` class to something else
> * [ ]  Add a new `AdaptiveCovarianceMCMC` that acts like the old one, but raises a deprecation warning. (Also add a comment in the code with the date we deprecated it)
> * [ ]  Add a tiny test (does it exist), and list it in the docs
> 
> I'm happy to do steps 2 and 3 @ben18785 !

@MichaelClerx I've done the name name change, if you want to sort the other two, that'd be dandy!

MichaelClerx:
@chonlei Could you please checkout this branch and test if your code still works? E.g. the tutorial stuff or the hERG code? It should work, but give you warnings! (If you don't see any warnings, you might have a previous pints installed with pip that you'll have to remove)

MichaelClerx:
@martinjrobins has checked some of his code, and that seems fine, so merging this!

ben18785:
Thanks Michael!

PR #925, opened 2019-09-12
ABC rejection - rebased

### LINKED ISSUE #96: Add ABC routine
Like @sanmitraghosh 's matlab or @adaly 's python

ben18785:
Closes #96


ben18785:
@MichaelClerx A nudge here -- @danielfridman98 wants to start working on adding ABC-SMC etc. on this, so would be good to get your thoughts on this.

MichaelClerx:
Hey @ben18785 @danielfridman98 , this one seems to have slipped past me. Do you still need a review of this? Could try having a look in the next few days probably...

examples/sampling/abc-rejection.ipynb
```diff
@@ -0,0 +1,154 @@
+{
```
MichaelClerx, line 1: This notebook could do with some more explanation!

- `rejection sampling` --> `[rejection sampling](https://pints.readthedocs.io/en/latest/abc_samplers/rejection_abc_sampler.html)`
- replace the `####` headings with some text explaining what the code block is going to do ("First, we set up a single series problem" etc)
- Add some comments near the log. Is it the "acceptance rate" we're wanting to monitor here?
- Explain the final figure
- Why only one parameter when we've got a 2d problem?

pints/_abc/__init__.py
```diff
@@ -0,0 +1,348 @@
...
+        that no parameters were accepted (tell allows for multiple evaluations
+        per iteration).
+
+        """
+        raise NotImplementedError
+
+
+class ABCController(object):
+    """
+    Samples from a :class:`pints.LogPrior`.
```
MichaelClerx, line 50: I don't think that's the best one-line description of an ABC routine we can come up with?
Where's the `error_measure` in this, for example?

pints/_abc/__init__.py
```diff
@@ -0,0 +1,348 @@
...
+        Performs an iteration of the ABC-rejection algorithm, using the
+        parameters specified by ask.
+
+        Returns the accepted parameter values, or ``None`` to indicate
+        that no parameters were accepted (tell allows for multiple evaluations
+        per iteration).
+
+        """
+        raise NotImplementedError
+
```
MichaelClerx, line 46: Is the threshold a universal property of all ABC methods?
If so, it could be added here

pints/_abc/__init__.py
```diff
@@ -0,0 +1,351 @@
...
+    def parallel(self):
+        """
+        Returns the number of parallel worker processes this routine will be
+        run on, or ``False`` if parallelisation is disabled.
+        """
+        return self._n_workers if self._parallel else False
+
+    def run(self):
+        """
+        Runs the ABC sampler.
```
MichaelClerx, line 199: Please add a paragraph explaining the return type

pints/_abc/__init__.py
```diff
@@ -0,0 +1,366 @@
...
+        """
+        raise NotImplementedError
+
+
+class ABCController(object):
+    """
+    Samples from a :class:`pints.LogPrior`.
+
+    Properties related to the number of iterations, parallelisation,
+    threshold, and number of parameters to sample can be set directly on the
```
MichaelClerx, line 53: "number of parameters to sample" ??

pints/_abc/__init__.py
```diff
@@ -0,0 +1,366 @@
...
+
+        """
+        raise NotImplementedError
+
+
+class ABCController(object):
+    """
+    Samples from a :class:`pints.LogPrior`.
+
+    Properties related to the number of iterations, parallelisation,
```
MichaelClerx, line 52: Before going into this, give a _very_ brief idea of what's going to happen. E.g. what is the "target number of samples", and what does the threshold do?

MichaelClerx: After that, but still before this paragraph, please add a code example where the controller is created

MichaelClerx: Also seems threshold is _not_ set here?

pints/_abc/__init__.py
```diff
@@ -0,0 +1,366 @@
...
+    threshold, and number of parameters to sample can be set directly on the
+    ``ABCController`` object, e.g.::
+
+        abc.set_max_iterations(1000)
+
+    Finally, to run an ABC routine, call::
+
+        posterior_estimate = abc.run()
+
```
MichaelClerx, line 61: There's a paragraph about logging you can pretty much copy-paste from MCMC or optimisation

pints/_abc/__init__.py
```diff
@@ -0,0 +1,366 @@
...
+    def max_iterations(self):
+        """
+        Returns the maximum iterations if this stopping criterion is set, or
+        ``None`` if it is not. See :meth:`set_max_iterations()`.
+        """
+        return self._max_iterations
+
+    def n_target(self):
+        """
+        Returns the target number of samples to obtain in the estimated
```
MichaelClerx, line 185: When you say "target", is this guaranteed to be met? If so, maybe leave out the word target in both the dosctring and the name and just say its the number of samples to sample?
(See also set_n_target)

MichaelClerx: Guessing from the code we might return _more_ than the target?

pints/_abc/__init__.py
```diff
@@ -0,0 +1,362 @@
...
+            accepted_vals = None
+            while accepted_vals is None:
+                xs = self._sampler.ask(self._n_workers)
+                fxs = evaluator.evaluate(xs)
+                evaluations += self._n_workers
+                accepted_vals = self._sampler.tell(fxs)
+
+            # Store the accepted samples
+            accepted_count += len(accepted_vals)
+            for val in accepted_vals:
```
MichaelClerx, line 271: Depending on the exact types used, this could a one line `samples.extend(accepted_vals)`

pints/tests/test_abc_rejection.py
```diff
@@ -0,0 +1,101 @@
...
+        samples = np.array(samples)
+        self.assertEqual(samples.shape[0], niter)
+
+    def test_errors(self):
+        # Test errors in abc rejection
+
+        abc = pints.RejectionABC(self.log_prior)
+        abc.ask(1)
+        # test two asks raises error
+        self.assertRaises(RuntimeError, abc.ask, 1)
```
MichaelClerx, line 82: Please use assertRaisesRegex here and below

pints/_abc/_abc_rejection.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy as np
+
+import pints
+
+
+class RejectionABC(pints.ABCSampler):
+    """
+    Rejection ABC sampler.
+
+    See, for example, [1]_. In each iteration of the algorithm, the following
```
MichaelClerx, line 20: I don't like the "for example" here. Is this ref the _exact_ version we implement or not?

pints/_abc/_abc_rejection.py
```diff
@@ -0,0 +1,89 @@
...
+    steps occur::
+
+        theta* ~ p(theta), i.e. sample parameters from prior distribution
```
MichaelClerx, line 23: Math block?

pints/_abc/_abc_rejection.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def name(self):
+        """ See :meth:`pints.ABCSampler.name()`. """
+        return 'Rejection ABC'
+
+    def ask(self, n_samples):
+        """ See :meth:`ABCSampler.ask()`. """
+        if self._ready_for_tell:
+            raise RuntimeError('ask called before tell.')
+        self._xs = self._log_prior.sample(n_samples)
```
MichaelClerx, line 50: Please add `self._xs.setflags(write=False)` before returning.
The reason is, if you return an array, I can modify it, e.g.:

xs = abc.ask()
xs[0] = 'yellow'

pints/_abc/_abc_rejection.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self._ready_for_tell = True
+        return self._xs
+
+    def tell(self, fx):
+        """ See :meth:`ABCSampler.tell()`. """
+        if not self._ready_for_tell:
+            raise RuntimeError('tell called before ask.')
+        self._ready_for_tell = False
```
MichaelClerx, line 59: Please add a blank line after this to show we're done checking the ready-for-tell business and are now doing something else

pints/_abc/_abc_rejection.py
```diff
@@ -0,0 +1,89 @@
...
+        if isinstance(fx, list):
```
MichaelClerx, line 60: Don't think this check should be here.

Other methods (e.g. optimisers, mcmc) always use a sequence-type (e.g. a list) for `xs` and `fxs`, even if there's only a single sample in it. This makes everything much easier, because you don't get ifs like this!

pints/_abc/_abc_rejection.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            if fx < self._threshold:
+                return self._xs
+            else:
+                return None
+
+    def threshold(self):
+        """
+        Returns threshold error distance that determines if a sample is
+        accepted (is error < threshold).
```
MichaelClerx, line 76: is --> if (and see below)

MichaelClerx:
🎵 PR 9 2 5! What a way to make a living

MichaelClerx:
@ben18785 @phumtutum are we closing and deleting this one then?

ben18785:
Yep, I think this one can go after @phumtutum's new PR.

Issue #849, opened 2019-06-05
Use this sphinx doc layout


MichaelClerx:
https://salib.readthedocs.io/en/latest/api.html#sobol-sensitivity-analysis

Looks great, w. references etc.

Would require:

- [x] Find out what the theme is called
- [x] Start using it
- [x] Update docstrings, if required
- [x] Update contributing.md with a basic template for docstrings

abhidg:
Theme is https://github.com/bitprophet/alabaster - says it's the new default Sphinx theme.

abhidg:
We should also settle on a docstring format: either Sphinx, Google or numpy/scipy style.

MichaelClerx:
Prioritising: Should be easily done

MichaelClerx:
Overlaps with #923

MichaelClerx:
Turns out the nice bits are from sphinx extensions (specifically `napoleon`, so am trying that)

MichaelClerx:
Fxied in #944

Issue #99, opened 2017-11-29
End of __init__ method not clear in documentation


chonlei:
![image](https://user-images.githubusercontent.com/26725646/33403883-066de36c-d55a-11e7-90b7-e97a14379f82.png)

Can we somehow add a line or indentation or any other thing to separate the class `__init__` arguments and the first class method? (i.e. separate `values` and `dimension()`.) It took me awhile to realise one is an argument and the other is a method (I guess even harder for the user!).

MichaelClerx:
I agree, but think that would mean customising the css, so very low priority!

MichaelClerx:
https://github.com/guzzle/guzzle_sphinx_theme/issues/30#issuecomment-406206923

We can fix it with a bit of custom CSS!

MichaelClerx:
See #849

MichaelClerx:
This should be fixed if we start using Numpy style parameter lists & literature references

MichaelClerx:
Fixed in #944

PR #944, opened 2019-09-24
Testing out alabaster theme, with napoleon extension for numpy style argument lists.

MichaelClerx:
See #849 #99

@martinjrobins please have a look at some of the example changes in the diff and see if that's OK

Will be a bit of work to start using this everywhere


MichaelClerx:
Todo:

- [ ] Everything that shows up in `grep "Arguments:" ./pints -Ir`
- [ ] Everything that shows up in `grep "   \[1" ./pints -Ir`
- [ ] Lots of manual checking

MichaelClerx:
> Is there a way to get sphynx to automatically say what the base classes are (rather than manually putting in "Extends....")?

I've looked into `:show-inheritance` which does exactly that, but unfortunately outputs names like `pints._mcmc.MCMCSampler`. It also fails to make links for those as it doesn't recognise it as an indexed class

PR #968, opened 2019-09-26
Bug in loglikelihoods test.

MichaelClerx:
@ben18785 could you have a look at this? I spotted a bug:

    self.assertTrue(np.abs(y - ll(x) < 1e-6))

this asserts that np.abs(z) is true, where z is the expression `y - ll(x) < 1e-6`
As it turns out, `y - ll(x)` is negative, so that z = True, and abs(z) = 1...

Corrected the bracket mistake, but now the tolerance is no longer met

### DIFF
--- a/pints/tests/test_log_likelihoods.py
+++ b/pints/tests/test_log_likelihoods.py
@@ -618,7 +618,7 @@ class TestLogLikelihood(unittest.TestCase):
 
         # Note: y and ll(x) differ a bit, because the solver acts slightly
         # different when evaluating with and without sensitivities!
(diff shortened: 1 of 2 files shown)


MichaelClerx:
These should be written e.g. like this, btw:

    self.assertAlmostEqual(y, ll(x), places=6)

ben18785:
@MichaelClerx I've tried playing with the ODE solver tolerance but can't seem to make these values exactly equal (they are very close however). I think this is fine and likely just reflects the different routines that are being used to solve the two systems (one with sensitivities, one without). So have just made the test less onerous.

Issue #976, opened 2019-09-29
Acceptance not logging correctly for Adaptive covariance methods


ben18785:

MichaelClerx:
Fixed in #991 using code from #977

PR #977, opened 2019-09-29
Acceptance rate logging for adaptive methods now works

ben18785:
Addresses #976


pints/_mcmc/_adaptive_covariance.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, x0, sigma0=None):
         # Acceptance rate monitoring
         self._iterations = 0
         self._acceptance = 0
+        self._accepted_count = 0
```
MichaelClerx: I don't like how this is split up: Why is this variable defined here, but only used by a different class?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -165,6 +166,9 @@ def tell(self, fx):
         # Ensure fx is a float
         fx = float(fx)
 
+        # Increase iteration count
+        self._iterations += 1
```
MichaelClerx: Again, we don't use these here, so why are they this class's responsibility?

ben18785: Not sure where else to put these -- they are used by the controller, so I put them in the highest class I could?

pints/_mcmc/_adaptive_covariance.py
ben18785: Ok. Have moved these to the daughter class since this is where they are used.

pints/_mcmc/_adaptive_covariance.py
MichaelClerx: No they're not :D
The controller should not, and as far as I can see does not, access other objects private properties.
The iteration count inside the original AdaptiveCovarianceMCMC was maintained purely to calculate an acceptance rate for user output

ben18785: Ok. The way I have it at the moment, all adaptive methods require an iteration count so that we can calculate an acceptance rate (``self._acceptance``) which is logged by the controller. To me, to avoid having to define an iteration counter in each method, it seems sensible to have this in the base class. What do you think?

MichaelClerx:
This was cheekily incorporated in #991

PR #982, opened 2019-09-30
DRAM ACMC

### LINKED ISSUE #295: Implement delayed acceptance adaptive Metropolis (DRAM)
Described [here](https://bmcsystbiol.biomedcentral.com/articles/10.1186/s12918-017-0433-1).

ben18785:
Closes #295


MichaelClerx:
Hey @ben18785 
Looks like this needs updating to the new structure? Imports a GlobalAdaptiveCovarianceMCMC at some point that's no longer define

pints/_mcmc/_dram_ac.py
```diff
@@ -0,0 +1,292 @@
...
+    DRAM (Delayed Rejection Adaptive Covariance) MCMC, as described in [1]_.
+
+    In this method, rejections do not necessarily lead an iteration to end.
+    Instead, if a rejection occurs, another point is proposed although
+    typically from a narrower (i.e. more conservative) proposal kernel than was
+    used for the first proposal.
+
+    In this approach, in each iteration, the following steps return the next
+    state of the Markov chain (assuming the current state is ``theta_0`` and
+    that there are 2 proposal kernels)::
```
MichaelClerx: Replace `::` with `..math:` and all that

ben18785:
> Looks good, but notebook could do with a bit of interpretation of the results.
> 
> "Try with 2 proposal kernels instead." Could you point out the issues in the first run in the images? And show how that gets better?
> 
> Will make some small suggestions in the code now too :-)

Very good point -- the notebook was pretty bare before! haha

Have now added more detail and made the notebook a bit more readable. See what you think.

pints/_mcmc/_dram_ac.py
ben18785: This isn't supposed to be a math section though -- writing pseudocode in math expressions is OTT in my view -- and also, here we use the actual parameter names as they are called in the code, which I think is a benefit. I'm also hesitant to change this because I was careful to make sure things matched up with the paper before and am fearful of introducing a mistake!

MichaelClerx: Oh, sorry! Good point!

MichaelClerx:
Thanks Ben, looks great!
To me the traces look wild! Is that a good thing? Massively-mixing mcmc?

MichaelClerx:
@ben18785 are you happy for me to merge this?

ben18785:
Yep!

Issue #983, opened 2019-09-30
Add simple harmonic oscillator ODE model


ben18785:
Stan uses this for their examples, so think we should too.

MichaelClerx:
Maybe more something for performance comparison? We've got much nicer ODEs already :-)

MichaelClerx:
Cool! Don't forget these guys though: https://github.com/pints-team/pints/projects/8 :D

MichaelClerx:
Closed in #984

PR #984, opened 2019-09-30
Simple harmonic oscillator toy model

ben18785:
Addresses #983


MichaelClerx:
A very useful resource for really getting to grips with the harmonic oscillator: https://youtu.be/lQ6jZgMaZk4?t=26

pints/toy/_sho_model.py
```diff
@@ -0,0 +1,115 @@
...
+
+            values = 1 / (2 * sqrt) * (
+                2 * y2_0 * (exp2 - exp1) + y1_0 * theta * (exp2 - exp1) +
+                y1_0 * sqrt * (exp1 + exp2)
+            )
+            values = np.real(values)
+
+            if sensitivities:
+                exp3 = np.exp(-0.5 * theta * times)
+                exp4 = np.exp(times * sqrt)
```
MichaelClerx: Looks like we could cache some of these, but might not be worth the effort (and reduced legibility)!

docs/source/toy/index.rst
```diff
@@ -32,6 +32,7 @@ Some toy classes provide extra functionality defined in the
     repressilator_model
     rosenbrock
     simple_egg_box_logpdf
+    simple_harmonic_oscillator_model
```
MichaelClerx: I prefer this to the "sho" model filenames used below. Unless that's a common abbrev.?

ben18785:
Ha, it's a common abbreviation in physics, at least!

pints/toy/_sho_model.py
```diff
@@ -0,0 +1,115 @@
...
+class SimpleHarmonicOscillatorModel(pints.ForwardModelS1, ToyModel):
+
+    r"""
+    Simple harmonic oscillator model for a particle that experiences a force
+    in proportion to its displacement from an equilibrium position,
+    and, in addition, a friction force. The system's behaviour is determined by
+    a second order ordinary differential equation (from Newton's second law):
+
+    .. math::
+        \frac{d^2y}{dt^2} = -y(t) - \theta \frac{dy(t)}{dt}
```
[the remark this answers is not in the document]
ben18785: Ha, I made up this particular inference form, so don't think a citation is necessary! Though it was likely to reflect reality -- we may know the mass of the particle with high precision (so why not make it 1) but may be uncertain about the initial state of the particle, its velocity and the magnitude of the frictional force.

pints/toy/_sho_model.py
ben18785: Yeah, not sure it's worth it since, whenever theta changes (as would occur in inference), they would need to be updated.

docs/source/toy/index.rst
ben18785: Yep, common abbreviation!

pints/toy/_sho_model.py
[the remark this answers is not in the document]
MichaelClerx: Cool!

Issue #974, opened 2019-09-27
Error in toy annulus notebook


MichaelClerx:
```
Running...
Done!
Running...
Using Hamiltonian Monte Carlo
Generating 4 chains.
Running in sequential mode.
Iter. Eval. Accept.   Accept.   Accept.   Accept.   Time m:s
0     4      0         0         0         0          0:00.0
1     84     0.333     0.333     0.333     0.333      0:00.0
2     164    0.5       0.5       0.5       0.5        0:00.0
3     244    0.6       0.6       0.6       0.6        0:00.1
20    1604   0.909     0.909     0.864     0.909      0:00.3
40    3204   0.952381  0.952381  0.929     0.952381   0:00.6
60    4804   0.968     0.968     0.952     0.968      0:00.8
80    6404   0.976     0.976     0.963     0.976      0:01.1
100   8004   0.98      0.98      0.971     0.98       0:01.3
120   9604   0.984     0.984     0.975     0.984      0:01.6
140   11204  0.986     0.986     0.971831  0.986      0:01.8
160   12804  0.988     0.988     0.975     0.988      0:02.1
180   14404  0.989011  0.989011  0.978022  0.989011   0:02.3
200   16004  0.990099  0.990099  0.980198  0.990099   0:02.6
220   17604  0.990991  0.990991  0.977     0.990991   0:02.8
240   19204  0.992     0.992     0.979     0.992      0:03.1
260   20804  0.992     0.992     0.980916  0.992      0:03.3
280   22404  0.993     0.993     0.982     0.993      0:03.6
300   24004  0.993     0.993     0.983     0.99       0:03.8
320   25604  0.991     0.994     0.984472  0.991      0:04.1
340   27204  0.991     0.994152  0.985     0.991      0:04.3
360   28804  0.992     0.994     0.986     0.992      0:04.6
380   30404  0.992     0.992     0.986911  0.992      0:04.8
400   31924  0.993     0.993     0.988     0.993      0:05.1
Halting: Maximum number of iterations (400) reached.
Done!
Running...


-- stderr ---------------------------------------------------------------------

Traceback (most recent call last):
  File "<string>", line 165, in <module>
  File "/home/travis/build/pints-team/pints/pints/_mcmc/__init__.py", line 561, in run
    fxs = evaluator.evaluate(xs)
  File "/home/travis/build/pints-team/pints/pints/_evaluation.py", line 107, in evaluate
    return self._evaluate(positions)
[...]

PR #985, opened 2019-10-01
Check that method in MCMC controller is not abstract base class

ben18785:
Addresses #974


ben18785:
Fair play -- please do! I wasn't sure either. Other that this, it would
require us to implement some sort of class method that says whether or not
an object is abstract, perhaps?

MichaelClerx:
I was hoping it would follow naturally somehow!

MichaelClerx:
Fixed instead in #991

Issue #329, opened 2018-04-05
Rewrite "Results object" ideas as methods on the MCMCController


ben18785:
I am opening up this can of worms again, sorry. 

After reflecting on this, I now agree with Michael that it's probably best to return just a numpy array as a result of sampling. 

I do think, however, that it's still useful to have a sampling result object/table function that basically summarises the result of sampling, if a user calls it. This ranges from parameter quantiles, to R-hats (for MCMC only) to effective sample sizes (per second), to other things which are sampler dependent. This is standard in other software and I certainly find it useful and less cumbersome than histograms for many-parameter models.

I have my MCMC result object from before but am considering simplifying it as well as generalising it to other sampling methods (and making tests for it, sorry!).

mirams:
Hmmmm, quite nice to be able to calculate ESS etc. on a results object rather than the MCMC thing itself, as you might want to manually thin the chain, discard more burn in, or something first. So it's not clear to me that you want these methods on the MCMCsampler rather than MCMCresults. An subclass of numpy array that had such methods might work, if practical?

MichaelClerx:
Still think this should be done by adding methods to the Controller class to return anything we'd like to know. See #271

MichaelClerx:
See also #862

MichaelClerx:
Closed in #986

PR #986, opened 2019-10-02
Results method for MCMC -- prints lovely summaries to the screen

ben18785:
Addresses #329.


ben18785:
@MichaelClerx Any idea about the Travis error messages here? I've had a Google but it doesn't really make sense to me.

ben18785:
@MichaelClerx Why? It's just an extension of diagnostic methods? Did you see the example notebook? It's not a results object - just a method that is called on the chains.  I don't like having it as an option for the controller object because, if you fail to select it, you then can't use the methods afterwards.

ben18785:
So, to be clear, the MCMC still returns the same thing. All we are doing is wrapping the diagnostic methods into a useful helper function.

MichaelClerx:
OK, sorry, I'll take a detailed look asap!

setup.py
```diff
@@ -60,6 +60,7 @@ def load_version():
         # on systems without an attached display, it should never be imported
         # outside of plot() methods.
         'matplotlib>=1.5',
+        'tabulate>0.1',
```
MichaelClerx: Why the version number?

pints/_mcmc/_results.py
```diff
@@ -0,0 +1,204 @@
...
+#
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+from tabulate import tabulate
+
+
+class MCMCResults(object):
```
MichaelClerx: I'm wondering if there's a better name, like MCMCResultAnalysis, or MCMCResultSummary or something

MichaelClerx: MCMCSummary, MCMCDiagnosis, MCMCAnalysis ?

setup.py
ben18785: Removed version number

pints/_mcmc/_results.py
ben18785: Yep, good idea! Have changed it to MCMCSummary.

ben18785:
Great — cheers Michael!

PR #991, opened 2019-10-08
Simplified adaptive covariance mcmc class hierarchy.

MichaelClerx:
Simplification of the recent changes to adaptive covariance MCMC.

Changes w.r.t. previous:

- There is no longer a separate GlobalAdaptiveCovarianceMC, we can re-introduce that later if needed
- The task of the subclasses is more clearly defined (although they still use internal parameters of the parent classes)
- You can read and understand the base class without having to look at the adaptations
- Trying to use the base class on its own generates NotImplementedErrors (see #974, #985 )


MichaelClerx:
Cheers Ben!

PR #993, opened 2019-10-08
Repeated calls to DiffEv ask() increased iteration count.

MichaelClerx:
We're not explicit about this in the docs, but MCMC methods have two options for ask():

1. Only allow one call (we do this for hamiltonian methods) between tells
2. Return the same proposal if ask() is called repeatedly

DiffEv followed pattern number 2, but updated an iteration count with each call

### DIFF
--- a/pints/_mcmc/_differential_evolution.py
+++ b/pints/_mcmc/_differential_evolution.py
@@ -96,13 +96,14 @@ class DifferentialEvolutionMCMC(pints.MultiChainMCMC):
         if not self._running:
             self._initialise()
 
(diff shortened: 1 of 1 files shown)


MichaelClerx:
Good point, I've updated it to read "Only allow one call (we do this for hamiltonian methods) between tells"

So mode 1) ask > tell > ask > tell etc..
mode 2) ask+ > tell > ask+ > tell, where ask+ means ask as often as you like but at least once :D 

You're right we don't have a use case for this though. Maybe we should make a ticket?

PR #994, opened 2019-10-08
Make MALA allow repeated ask()

MichaelClerx:
There's no reason not!

Also cleaned up the tests a bit

### DIFF
--- a/pints/_mcmc/_mala.py
+++ b/pints/_mcmc/_mala.py
@@ -85,7 +85,6 @@ class MALAMCMC(pints.SingleChainMCMC):
 
         # Set initial state
         self._running = False
-        self._ready_for_tell = False
 
         # Current point and proposed point
         self._current = None
@@ -166,9 +165,6 @@ class MALAMCMC(pints.SingleChainMCMC):
         if not self._running:
             self._initialise()
 
-        if self._ready_for_tell:
-            raise RuntimeError('Ask() called when expecting call to tell().')
-
         # Propose new point
         if self._proposed is None:
 
@@ -187,8 +183,6 @@ class MALAMCMC(pints.SingleChainMCMC):
             # Set as read-only
             self._proposed.setflags(write=False)
 
-        self._ready_for_tell = True
-
         # Return proposed point
         return self._proposed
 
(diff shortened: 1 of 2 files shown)


pints/_mcmc/_mala.py
```diff
@@ -166,9 +165,6 @@ def ask(self):
         if not self._running:
             self._initialise()
 
-        if self._ready_for_tell:
-            raise RuntimeError('Ask() called when expecting call to tell().')
-
         # Propose new point
```
ben18785: Not sure I follow why we'd want to remove this check? Probably missing the point, but what would be a use case for repeatedly calling ask?

pints/_mcmc/_mala.py
```diff
@@ -187,18 +183,15 @@ def ask(self):
...
         return self._proposed
 
     def tell(self, reply):
         """ See :meth:`pints.SingleChainMCMC.tell()`. """
 
         # Check if we had a proposal
-        if not self._ready_for_tell:
+        if self._proposed is None:
             raise RuntimeError('Tell called before proposal was set.')
-        self._ready_for_tell = False
```
ben18785: Sorry, probably not getting something but don't quite get why we'd remove these checks.

MichaelClerx: It's more in line with what we have in the other MCMC methods, with the exception of the ones that do more complicated stuff in ask() than just set a proposal.

So no very strong reason! Just that it's possible to let users ask() as often as they like in this case (and always get the same result), and it lets us remove a variable...

We might want to disallow this everywhere instead though? In which case we'd have to update the documentation for SingleChainMCMC and MultiChainMCMC a bit to make this explicit...

ben18785:
Cool. Not sure about this one currently, to be honest. Let me think about
it...

MichaelClerx:
Made a ticket here #996

MichaelClerx:
Can we re-open discussion on this, @ben18785 and @martinjrobins ? See especially https://github.com/pints-team/pints/issues/996

MichaelClerx:
No thoughts on this?
It needs to be resolve one of two ways, I think:

1. Make it so that you can only ask once, in every method: a second call to ask() breaks everything
2. Make it so that you can ask as often as you like, in every method.

Issue #980, opened 2019-09-30
Break up large `test_log_x` files


MichaelClerx:
For example, `test_log_priors` starts like this:

```
class TestPrior(unittest.TestCase):

    def test_beta_prior(self):

        # Test input parameters
        self.assertRaises(ValueError, pints.BetaLogPrior, 0, 0)
        self.assertRaises(ValueError, pints.BetaLogPrior, 2, -2)
        self.assertRaises(ValueError, pints.BetaLogPrior, -2, 2)

        p1 = pints.BetaLogPrior(0.123, 2.34)
        p2 = pints.BetaLogPrior(3.45, 4.56)

        points = [-2., 0.001, 0.1, 0.3, 0.5, 0.7, 0.9, 0.999, 2.]

        # Test means
        self.assertAlmostEqual(p1.mean(), 0.04993909866017051)
        self.assertAlmostEqual(p2.mean(), 0.4307116104868914)

        # Test n_parameters
        self.assertEqual(p1.n_parameters(), 1)

        # Test specific points
        for point in points:
            to_test = [point]
            self.assertAlmostEqual(
                scipy.stats.beta.logpdf(to_test[0], 0.123, 2.34),
                p1(to_test),
                places=9)
            self.assertAlmostEqual(
                scipy.stats.beta.logpdf(to_test[0], 3.45, 4.56),
                p2(to_test),
                places=9)

        # Test derivatives
        p1_derivs = [0., -878.341341341342, -10.25888888888889,
                     -4.837619047619048,
                     -4.434, -5.719523809523809, -14.37444444444445,
                     -1340.877877877876,
                     0.]

        p2_derivs = [0., 2446.436436436437, 20.54444444444445,
                     3.080952380952382,
                     -2.219999999999999, -8.36666666666666, -32.87777777777778,
                     -3557.547547547544, 0.]

        for point, deriv in zip(points, p1_derivs):
            calc_val, calc_deriv = p1.evaluateS1([point])
            self.assertAlmostEqual(calc_deriv[0], deriv)

        for point, deriv in zip(points, p2_derivs):
            calc_val, calc_deriv = p2.evaluateS1([point])
            self.assertAlmostEqual(calc_deriv[0], deriv)

        # Test pathological edge cases
        p3 = pints.BetaLogPrior(1.0, 1.0)
        calc_val, calc_deriv = p3.evaluateS1([0.0])
        self.assertAlmostEqual(calc_deriv[0], 0.0)

[...]

Issue #640, opened 2019-02-03
Return log-likelihood at each sampled value


ben18785:
I remember we had a discussion about this previously (I've looked through the issues but couldn't find the relevant one). 

We need code that returns the log-likelihood at each sample value. Do we have this implemented currently? I couldn't see it in the mcmc class.

Quite a few things depend on this: marginal likelihood estimators and model comparison stuff for example. Happy to implement this myself but want to check that no one else has done so yet.

MichaelClerx:
Hi @ben18785 !

We've got code for logging this to disk, but not returning in memory just yet

https://github.com/pints-team/pints/blob/master/pints/_mcmc/__init__.py#L444-L470
https://github.com/pints-team/pints/blob/master/pints/_mcmc/__init__.py#L568-L578

We can certainly discuss adding this tomorrow, perhaps in light of #329 as well

MichaelClerx:
Discussing during pints meeting with @martinjrobins @ben18785 and @MichaelClerx 

We think the best thing to do is to start storing this (and other things!) in the MCMCSampling and Optimisation classes. They are "single-use" objects already, so it makes sense to start using them as their own "result objects" (#329 #154 #129)

So the changes to the interface would be e.g.:

- run() still outputs the chains
- add a `set_loglikelihood_storage(store)` or similar, plus a get method for the setting
- add a `get_stored_loglikelihoods()` method or similar

Further, we could add a 
- `get_chains` to get the chains a second time (or in case people want to get everything via get_...)
- `set_chain_storage()` or similar, to disable chain storage in memory #584

MichaelClerx:
> We've got code for logging this to disk, but not returning in memory just yet

See if there's some way to use the logging methods for this --> log to screen, disk, or memory

MichaelClerx:
Hey @martinjrobins , what do you think:

`run()` returns all chains, except if `set_chain_storage(False)` is called, in which case it returns None

I'm now adding a `set_evaluation_storage()` method. Should that also update the output of `run()` ? Or should the evaluations be retrieved using a separate method?

MichaelClerx:
Going with `set_log_pdf_storage()` and `log_pdfs()` for now to enable storage and retrieve the evaluations

martinjrobins:
sounds good. I would have `set_evaluation_storage()` update the output of `run()` to be consistent with `set_chain_storage()`

MichaelClerx:
[...]

PR #1011, opened 2019-11-12
More granular MCMCController test structure, and bugfix

MichaelClerx:
Made MCMCController tests more granular (similar to open issue #980)

In the process saw that one of the tests was wrong, fixed it, and uncovered bug (then fixed that)

### DIFF
--- a/pints/_mcmc/__init__.py
+++ b/pints/_mcmc/__init__.py
@@ -567,8 +567,10 @@ class MCMCController(object):
             if self._single_chain:
 
                 # Check and update the individual chains
+                xs_iterator = iter(xs)
                 fxs_iterator = iter(fxs)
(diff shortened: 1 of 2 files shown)


MichaelClerx:
@ben18785 if you could have a quick look it'd help me get to #640

MichaelClerx:
There was a bug for the situation where we had `n` samplers, but only `m <= n` of them were active. In this case the index of a sampler in the array of samplers was given by `i`, but the index of a sample `x` was not necessarily `i`. So I changed the code to stop using `i` when referring to a sample, but instead to use an iterator over the `xs` (so that we don't need to know the indices, just that its the next available sample)

ben18785:
Makes sense! Good spot.

Issue #584, opened 2018-12-10
Add option to MCMCSampling to _not_ store the chain in memory


MichaelClerx:
Now that we have a write-do-disk option, it should be possible to avoid storing the chain in memory. For example, we could have a switch

```
mcmc.set_return_chain(False)
```

that stops the method from storing and returning the chain?

MichaelClerx:
See also #640

PR #1012, opened 2019-11-12
584 dont store chains

MichaelClerx:
See #584


MichaelClerx:
Another step towards #640 @ben18785 !
In this one we add the option to disable storage of samples in memory (in case you want to only store them on disk). There's still an array of samples in this code, but now it's an array with a single entry for every chain (instead of `n_iterations` entries per chain)

MichaelClerx:
Review would be much appreciated!

MichaelClerx:
Psssst! @martinjrobins @ben18785 !

MichaelClerx:
Thanks both!

Issue #787, opened 2019-04-16
The examples for `pints.plot` do not use `pints.plot`


MichaelClerx:
- [ ] [Should use `pints.plot.trace`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-trace-plots.ipynb)
- [ ] [Should use `pints.plot.series`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-predicted-time-series.ipynb)
- [ ] [Should use `pints.plot.histogram` and `pints.plot.pairwise`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-pairwise-scatterplots.ipynb)
- [ ] [Should also use `pints.plot.histogram` and `pints.plot.pairwise`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-pairwise-kde-plots.ipynb)
- [ ] [Should use `pints.plot.autocorrelation`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-autocorrelation.ipynb)

The link at the top of all these examples is broken too

chonlei:
Hmmm?! These 'plot-mcmc-' example series are the break down of `pints.plot` codes, so we didn't use any `pints.plot` there.

But now think about it, I guess we could, for all of them, add a link saying user can use `pints.plot`, a one line code, to plot the same thing?

MichaelClerx:
Well yeah, they should all show how to use `pints.plot`, not how to replace it with custom code :-)

MichaelClerx:
@chonlei :D

rcw5890:
> The link at the top of all these examples is broken too

@MichaelClerx Which of the three adaptive covariance MCMC examples should be linked to? Maybe 
https://github.com/pints-team/pints/blob/pints-plot-examples/examples/sampling-adaptive-covariance-haario.ipynb

MichaelClerx:
Good question! I'd suggest the haario-bardenet one for now.

@ben18785 we messed up quite a few examples I think in the adaptive covariance renaming :D
Should we add some kind of general adaptive covariance notebook, or just link to the haario-bardenet one?

rcw5890:
> * [ ]  [Should also use `pints.plot.histogram` and `pints.plot.pairwise`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-pairwise-kde-plots.ipynb)

[...]

PR #1017, opened 2019-11-13
787 examples use pints plot

rcw5890:
See #787


MichaelClerx:
Thanks! Could you also update the links at the top of each notebook to point to https://pints.readthedocs.io/en/latest/mcmc_samplers/adaptive_covariance_mc.html ?
And remove the link to the KDE example + delete the file

Issue #1018, opened 2019-11-14
Add diagnostics for autocorrelated errors


rcw5890:
After fitting a model assuming independent errors, it would be useful to have a diagnostic to check if the assumption of independent errors was appropriate. 

One simple diagnostic is to produce an autocorrelation plot of the residuals.

`pints.plot.series` already encodes much of the functionality (simulating for potentially thinned samples). The calculation and display of the residuals autocorrelation needs to be added.

rcw5890:
Closed in #1020

PR #1020, opened 2019-11-19
1018 autocorrelated errors diagnostics

rcw5890:
See #1018


MichaelClerx:
Hi both! @ben18785 if you want to review this first in terms of functionality, I can have a look at the coding after?

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,255 @@
...
+                                   max_lag=10,
+                                   thinning=None,
+                                   significance_level=0.05,
+                                   posterior_interval=0.95):
+    """
+    Generate an autocorrelation plot of the residuals.
+
+    This function can be used to analyze the results of either optimisation or
+    MCMC Bayesian inference. When multiple samples of the residuals are present
+    (corresponding to multiple MCMC samples), the plot can illustrate the
```
ben18785: "...the plot illustrates..."

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,255 @@
...
+    distribution of autocorrelations across the MCMC samples. At each lag,
+    a dot is drawn at the median autocorrelation, and a line is drawn giving
```
ben18785: change "dot" to "point", I reckon

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,255 @@
...
+    the extent of the posterior interval specified as an argument (by default,
+    the 2.5th to the 97.5th percentile).
```
ben18785: percentile range

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # If there are multiple samples of the parameters, calculate the
+    # percentiles of the posterior to plot
+    if n_samples > 1:
+        if posterior_interval > 1 or posterior_interval < 0:
+            raise ValueError('posterior interval must fall between 0 and 1')
+        upper_pctle = (0.5 + posterior_interval / 2) * 100
+        lower_pctle = (0.5 - posterior_interval / 2) * 100
+
+    # Find the location of the confidence interval dotted lines
```
ben18785: Do we have a citation for the confidence intervals? If so, can we include it in the docstrings? Alternatively / additionally, it may be worth giving the mathematical form of this in the docstrings for this method.

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,170 @@
...
+        mcmc.set_max_iterations(300)  # make it as small as possible
+        mcmc.set_log_to_screen(False)
+        self.samples2 = mcmc.run()
+
+    def test_calculate_residuals(self):
+        """
+        Test the calculate_residuals function.
+        """
+        # Test that it runs on an optimisation result
+        pints.residuals_diagnostics.calculate_residuals(
```
ben18785: Can we check that the residual calculation works? I.e. that the difference between model predictions and actual values is actually that which is calculated.

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,170 @@
...
+            self.problem2,
+            thinning=0
+        )
+
+    def test_acorr(self):
+        """
+        Test the acorr function.
+        """
+        # Test that it runs without error
+        pints.residuals_diagnostics.acorr(np.random.normal(0, 1, 50), 11)
```
ben18785: Can we test this using series of a known autocorrelation structure? To check that the autocorrelations match the true values...

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,170 @@
...
+        # Test that it runs on MCMC samples without thinning
+        pints.residuals_diagnostics.calculate_residuals(self.samples2[0],
+                                                        self.problem2)
+
+        # Test that it runs on MCMC samples with thinning
+        pints.residuals_diagnostics.calculate_residuals(self.samples2[0],
+                                                        self.problem2,
+                                                        thinning=10)
+
+        # Check the message when the input is wrong dimension
```
ben18785: Do we test somewhere that the plots work with a multi-output problem? Here ``self.model2`` seems to be a multiple output model but we don't seem to test whether the plot works with it?

ben18785:
@MichaelClerx Have suggested a few changes, so may be best having a look through afterwards. @rcw5890 If you could let Michael know when these are done, that'd be great.

pints/residuals_diagnostics.py
rcw5890: I added a citation which I think is applicable here. I also added the basic equations to the docstring.

pints/tests/test_residuals_diagnostics.py
rcw5890: Is something like this what you had in mind? A fixed series with precomputed autocorrelation is now hardcoded into the test

pints/tests/test_residuals_diagnostics.py
rcw5890: I added an assertion for the multi-output figure, checking that it does indeed make multiple axes (one for each output) when a multi-ouput problem is provided. Currently, it doesn't look any deeper into the figure

rcw5890:
>  In the last example in the example notebook, can we use a higher thinning of samples so that it is easier to visually discern the uncertainty in the autocorrelations?

I managed to increase the visibility of the autocorrelation uncertainties by increasing the noise and decreasing the number of time points in the problem itself (thinning the MCMC samples didn't help).

pints/residuals_diagnostics.py
ben18785: Looks good!

pints/tests/test_residuals_diagnostics.py
ben18785: Cool! Thanks

pints/tests/test_residuals_diagnostics.py
ben18785: Yep, perfect!

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,272 @@
...
+        values are provided, this gives the size of the credible region of the
+        posterior to plot.
+
+    References
+    ----------
+    .. [1] Brockwell, P. J., & Davis, R. A. (1991). Time series: Theory and
+           methods (2nd ed.). New York: Springer.
+    """
+    import matplotlib.pyplot as plt
+    import numpy as np
```
MichaelClerx: Numpy and scipy can be imported at module level.

We avoid importing matplotlib at the module level because users sometimes need to configure matplotlib before it's first imported. Module-level imports get executed when pints itself (or in this case `pints.residuals_diagnostics`) is imported, which would means users could no longer configure matplotlib after they've imported pints. To avoid that, we stick the matplotlib import in the method, because that only gets run when the method is executed.

(There's no overhead difference by the way, imported modules are cached globally so writing an import in a method only does the hard work once, the 2nd time the `import` statement returns a reference)

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,272 @@
...
+
+    This function returns a ``matplotlib`` figure.
+
+    Parameters
+    ----------
+    parameters
+        The parameter values with shape ``(n_samples, n_parameters)``. When
+        passing a single best fit parameter vector, ``n_samples`` will be 1.
+    problem
+        The problem given by a :class:``pints.SingleOutputProblem`` or
```
MichaelClerx: For mysterious reasons, the authors of restructuredtext/sphinx decided that it should be:

```
Reference to ``a_variable`` or ``1 + general_code()`` (double ticks)
A reference to a class or method (gets converted to a hyperlink) :meth:`a_single_tick()`
```

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,199 @@
...
+#
+import pints
+import pints.residuals_diagnostics
+import pints.toy as toy
+import unittest
+import numpy as np
+import matplotlib
+
+# Select matplotlib backend that doesn't require a screen
+matplotlib.use('Agg')
```
MichaelClerx: This is an example of the above! If you had an `import matplotlib.pyplot` anywhere above this line (including inside pints), matplotlib would complain

(This code is fine, by the way, just continuing my above comment :-) )

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,199 @@
+#
+# Test the methods in pints.residuals_diagnostics
+#
+# This file is part of PINTS
+#  Copyright (c) 2017-2019, University of Oxford.
+#  For licensing informating, see the LICENSE file distributed with the PINTS
+#  software package.
+#
+import pints
```
MichaelClerx: We tend to do imports alphabetically, but I'm not sure we've always been consistent

MichaelClerx: (Except `from .. import ...`, those always go below the `import ...`)

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class TestResidualsDiagnostics(unittest.TestCase):
+    """
+    Tests Pints residuals diagnostics methods.
+    """
+    def __init__(self, name):
+        super(TestResidualsDiagnostics, self).__init__(name)
+
+        # Create a single output optimization toy model
```
MichaelClerx: Test initialisation should happen inside a `setupClass` method, see https://docs.python.org/3.5/library/unittest.html#class-and-module-fixtures

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,199 @@
...
+        self.values1 += np.random.normal(0, self.noise1, self.values1.shape)
+
+        # Set up optimisation problem
+        self.problem1 = pints.SingleOutputProblem(
+            self.model1, self.times1, self.values1)
+        self.score1 = pints.SumOfSquaresError(self.problem1)
+        self.boundaries1 = pints.RectangularBoundaries([0, 200], [1, 1000])
+        self.x01 = np.array([0.5, 500])
+
+        # Run the optimisation
```
MichaelClerx: Not sure about having this inside the test initialisation!

MichaelClerx: You could just generate some random "found parameters" and speed things up _a lot_

pints/tests/test_residuals_diagnostics.py
```diff
@@ -0,0 +1,199 @@
...
+            self.real_parameters2 * 0.9,
+            self.real_parameters2 * 1.05
+        ]
+        mcmc = pints.MCMCController(self.log_posterior2, 3, self.x02)
+        mcmc.set_max_iterations(300)  # make it as small as possible
+        mcmc.set_log_to_screen(False)
+        self.samples2 = mcmc.run()
+
+    def test_calculate_residuals(self):
+        """
```
MichaelClerx: There's an annoying thing with unit testing where it doesn't show filename or method name if your test has a docstring, so we try to _not_ have docstrings in tests, but instead just add comments at the top (e.g. `# Test the calculate_residuals function.`)

docs/source/residuals_diagnostics.rst
```diff
@@ -0,0 +1,30 @@
+*********************
+Residuals diagnostics
```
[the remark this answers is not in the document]
rcw5890: I have "Noise model diagnostics" now in the docs. Were you thinking of changing the module name as well?

examples/plot-residuals-autocorrelation.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "fig.set_size_inches(15, 7.5)\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Plotting autocorrelation of the residuals\n",
+    "Next we use the autocorrelation plot of the residuals to evaluate the noise model. This functionality is provided by the `plot_residuals_autocorrelation` method in `pints.residuals_diagnostics`."
```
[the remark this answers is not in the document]
rcw5890: I'm now using a link to the method documentation on readthedocs. It's a new method, so the link is broken until the docs get updated.

pints/tests/test_residuals_diagnostics.py
rcw5890: I used fixed found_parameters here, and I also replaced the MCMC chain with independent samples.

docs/source/residuals_diagnostics.rst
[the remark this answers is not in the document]
MichaelClerx: No that's fine with me :-)

docs/source/residuals_diagnostics.rst
```diff
@@ -0,0 +1,30 @@
...
+*********************
+
+.. module:: pints.residuals_diagnostics
+
+Pints includes functionality to generate diagnostic plots of the residuals.
+These tools may be useful to evaluate the validity of a noise model.
+
+Plotting functions:
```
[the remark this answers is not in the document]
MichaelClerx: Are these some kind of index?

pints/residuals_diagnostics.py
```diff
@@ -0,0 +1,272 @@
...
+def plot_residuals_autocorrelation(parameters,
+                                   problem,
+                                   max_lag=10,
+                                   thinning=None,
+                                   significance_level=0.05,
+                                   posterior_interval=0.95):
+    r"""
+    Generate an autocorrelation plot of the residuals.
+
+    This function can be used to analyze the results of either optimisation or
```
MichaelClerx: analyse

It's no big deal but we'll need to be consistent

docs/source/residuals_diagnostics.rst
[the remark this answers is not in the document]
rcw5890: Yeah, here I was roughly imitating [diagnostic_plots.rst](https://github.com/pints-team/pints/blob/master/docs/source/diagnostic_plots.rst) by categorising and listing the functions at the top. It may be unnecessary given the small size of the file however.

pints/residuals_diagnostics.py
rcw5890: Wasn't this fixed in b9fd6b4c76ac04d89f3adfcea79f56fbec573446

docs/source/residuals_diagnostics.rst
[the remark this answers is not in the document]
rcw5890: I could remove for now, and add later if ever the file becomes big enough that it would be helpful?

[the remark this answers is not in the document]
MichaelClerx: Thanks. Now wondering if we shouldn't just merge this with the existing `diagnostics.rst`. Maybe even move the functions into the main `pints` module like we did with the other diagnostics?
Any thoughts @ben18785 ?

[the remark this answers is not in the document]
MichaelClerx: https://pints.readthedocs.io/en/latest/diagnostics.html
https://pints.readthedocs.io/en/latest/diagnostic_plots.html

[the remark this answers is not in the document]
MichaelClerx: Thoughts @martinjrobins ?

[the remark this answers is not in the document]
martinjrobins: probably keep them in the submodule if that is where they currently are. I don't think it matters either way

PR #1025, opened 2019-12-03
WIP - Stochastic Logistic Growth (ABC Toy problem)

jarthur36:
Adds the _spatially implicit birth-only_ logistic growth model [(Simpson et al., 2019)](https://doi.org/10.1101/533182) to pints. Simulated using the interpretation of the Gillespie algorithm from [(Erban et al., 2007)](https://arxiv.org/abs/0704.1908)

This version is work in progress as the empirical mean of n stochastic runs does not converge to the deterministic mean as n->∞ as shown below
MeanDifference


pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,131 @@
...
+        if self._n0 < 0:
+            raise ValueError('Initial molecule count cannot be negative.')
+
+    def n_parameters(self):
+        """ See :meth:`pints.ForwardModel.n_parameters()`. """
+        return 2
+
+    def simulate_raw(self, parameters):
+        """
+        Returns raw times, mol counts when reactions occur
```
jarthur36: Remove reference to mol counts since i am using population size elsewhere

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,131 @@
...
+        t = 0
+        a = self._n0
+
+        # Run stochastic logistic birth-only algorithm, calculating time until next
+        # reaction and increasing population count by 1 at that time
+        mol_count = [a]
+        time = [t]
+        while a < k:
+            r = np.random.uniform(0, 1)
+            t += np.log(1/r) / (a * b * (1 - a / k))
```
jarthur36: This line may be causing the issue with means not lining up as expected, investigate whether having the growth rate per capita be population dependant breaks any of the maths

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,131 @@
...
+            a = a + 1
+            time.append(t)
+            mol_count.append(a)
+        return time, mol_count
+
+    def interpolate_mol_counts(self, time, mol_count, output_times, parameters):
```
jarthur36: Probably change mol count terminology throughout here too to avoid confusion

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,131 @@
...
+        time, mol_count = self.simulate_raw(parameters)
+
+        # interpolate
+        values = self.interpolate_mol_counts(time, mol_count, times, parameters)
+        return values
+
+    def mean(self, parameters, times):
+        r"""
+        Returns the deterministic mean of infinitely many stochastic
+        simulations, which follows :math:`A(0) \exp(-kt)`.
```
jarthur36: Change comment to display the true expected mean( not the one copies from stochastic degradation)

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError('Negative times are not allowed.')
+
+        return (self._n0 * k) / (self._n0 + np.exp(-b * times) * (k - self._n0))
+
+    def variance(self, parameters, times):
+        r"""
+        Returns the deterministic variance of infinitely many stochastic
+        simulations.
+        """
+        return NotImplementedError()
```
jarthur36: Need to work out how to calculate the variance for this model, or remove it as an option if this turns out to be too tricky

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,110 @@
...
+import pints
+import pints.toy
+from pints.toy import StochasticLogisticModel
+
+
+class TestStochasticLogistic(unittest.TestCase):
+    """
+    Tests if the stochastic logistic growth (toy) model works.
+    """
+    def test_start_with_zero(self):
```
MichaelClerx: If this is a stochastic method, we should seed the random generator and compare output values (search other tests for `seed(` to find examples)

ben18785: yep agreed

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,132 @@
...
+from __future__ import print_function, unicode_literals
+import numpy as np
+from scipy.interpolate import interp1d
+import pints
+
+from . import ToyModel
+
+
+class StochasticLogisticModel(pints.ForwardModel, ToyModel):
+    r"""
```
ben18785: We need docstrings that provide full information (including references) for the model. At a minimum, this should include the model equations and description of how the model works. It should also describe the parameters that the model takes as inputs. See the deterministic logistic model or the Goodwin-oscillator for examples of this.

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+#
+# Tests if the stochastic logistic growth (toy) model works.
+#
+# This file is part of PINTS.
```
MichaelClerx, line 5: There's a new template for these headers now (which is why the "copyright" test is failing)

pints/toy/_stochastic_logistic_model.py
chonlei: @ben18785 Do you remember which paper was it? It says "Simpson et al. 2019" in the note, but wasn't easy to search...

chonlei:
Just found out why there was the discrepancy between the deterministic model and the 1000-realisation average. Long in short, it all depends on the choice of parameters (in particular the _initial_ size of the population)!

With the one that Jack used, the population was small, so each step gave a big variation, so it had a large 'noise'. Although we thought by running more simulations it'd converge to the deterministic solution, it'd not be the case when the _initial_ population size is small (with some quick experiments, seems like n_0 > 10 would a reasonable size, but need some proper sensitivity anaylsis to give a bigger quote to that number). The reason seems to be that with a small initial population, the big noise at the beginning can drift (or shift?) the whole process (say to start the growth at a later time), then the average from these shifted solutions would give a bias.

So a fix to that is simply scale up the whole population and that'd solve the bias seen in the plot above, see below:

image
image

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,160 @@
+#
+# Stochastic degradation toy model.
```
ben18785, line 2: can we change to stochastic degradation model?

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,160 @@
...
+        super(StochasticLogisticModel, self).__init__()
+        self._n0 = float(initial_molecule_count)
+        if self._n0 < 0:
+            raise ValueError('Initial molecule count cannot be negative.')
+
+    def n_parameters(self):
+        """ See :meth:`pints.ForwardModel.n_parameters()`. """
+        return 2
+
+    def simulate_raw(self, parameters):
```
ben18785, line 62: Do we want this function to be _simulate_raw?

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,160 @@
...
+        time = [t]
+        while a < k:
+            r = np.random.uniform(0, 1)
+            t += np.log(1 / r) / (a * b * (1 - a / k))
+            a = a + 1
+            time.append(t)
+            mol_count.append(a)
+        return time, mol_count
+
+    def interpolate_values(self, time, pop_size, output_times, parameters):
```
ben18785, line 90: Same: not sure we'll want interpolate_values exposed? (So maybe _interpolate_values?)

ben18785 (changes requested):
Thanks @chonlei -- looks really good. Only a few minor comments then fine to go.

Just wondering -- did you see #1067 ? This was Jack's attempt to create a more generic stochastic simulation approach (basically it should automate the Gillespie algorithm for many types of stochastic reaction equation). I just remembered it, sorry. Do you think merge this then later someone tries to adapt it to fit into such a framework?

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,118 @@
...
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+import unittest
+import numpy as np
+import pints
+import pints.toy
+
+
+class TestStochasticLogistic(unittest.TestCase):
```
MichaelClerx, line 15: TestStochasticLogisticModel

docs/source/toy/stochastic_logistic_model.rst
```diff
@@ -0,0 +1,7 @@
+*************************
+Stochastic Logistic Model
```
MichaelClerx, line 2: Seems to have multiple names? Which one are we using?

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # Test the special case where the initial population count is zero
+        model = pints.toy.StochasticLogisticModel(0)
+        times = [0, 1, 2, 100, 1000]
+        parameters = [0.1, 50]
+        values = model.simulate(parameters, times)
+        self.assertEqual(len(values), len(times))
+        self.assertTrue(np.all(values == np.zeros(5)))
+
+    def test_start_with_one(self):
```
MichaelClerx, line 31: Maybe add comment saying why this is a testcase?

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,118 @@
...
+        times = [0, 1, 2, 100, 1000]
+        parameters = [0.1, 50]
+        values = model.simulate(parameters, times)
+        self.assertEqual(len(values), len(times))
+        self.assertEqual(values[0], 1)
+        self.assertEqual(values[-1], 50)
+        self.assertTrue(np.all(values[1:] >= values[:-1]))
+
+    def test_suggested(self):
+        np.random.seed(1)
```
MichaelClerx, line 46: Please stick in some 1 line comments in each of the tests

MichaelClerx: So that developers later can see e.g. what the difference between start-at-1 and simulate is

MichaelClerx, line 46: Why seeding here? (it's quite a slow operation)

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,118 @@
...
+        temp_time = np.array([np.random.uniform(time[0], time[1])])
+        self.assertTrue(model.interpolate_values(time, raw_values, temp_time,
+                                                 params)[0] == 1)
+        temp_time = np.array([np.random.uniform(time[1], time[2])])
+        self.assertTrue(model.interpolate_values(time, raw_values, temp_time,
+                                                 params)[0] == 2)
+
+    def test_mean_variance(self):
+        # test mean
+        np.random.seed(1)
```
MichaelClerx, line 78: why seed?

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,118 @@
...
+        model = pints.toy.StochasticLogisticModel(1)
+        v_mean = model.mean([1, 10], [5, 10])
+        self.assertEqual(v_mean[0], 10 / (1 + 9 * np.exp(-5)))
+        self.assertEqual(v_mean[1], 10 / (1 + 9 * np.exp(-10)))
+
+    def test_errors(self):
+        np.random.seed(1)
```
MichaelClerx, line 85: why seed?

pints/toy/_stochastic_logistic_model.py
MichaelClerx: and make the name consistent throughout!

pints/toy/_stochastic_logistic_model.py
MichaelClerx: Needs better docstring

pints/toy/_stochastic_logistic_model.py
MichaelClerx: Needs better docstring

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,160 @@
...
+            return np.zeros(times.shape)
+
+        # run Gillespie
+        time, pop_size = self.simulate_raw(parameters)
+
+        # interpolate
+        values = self.interpolate_values(time, pop_size, times, parameters)
+        return values
+
+    def mean(self, parameters, times):
```
MichaelClerx, line 123: Needs a better name, not really clear now that this is a simulation.

Also args need to be mentioned, plus return type etc.

pints/toy/_stochastic_logistic_model.py
```diff
@@ -0,0 +1,160 @@
...
+        if k <= 0:
+            raise ValueError("Carrying capacity must be positive")
+
+        times = np.asarray(times)
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+        c0 = self._n0
+        return (c0 * k) / (c0 + np.exp(-b * times) * (k - c0))
+
+    def variance(self, parameters, times):
```
MichaelClerx, line 147: Why is this here then? This isn't from an interface?

MichaelClerx (changes requested):

chonlei:
> Just wondering -- did you see #1067 ? This was Jack's attempt to create a more generic stochastic simulation approach (basically it should automate the Gillespie algorithm for many types of stochastic reaction equation). I just remembered it, sorry. Do you think merge this then later someone tries to adapt it to fit into such a framework?

Think I would go for merging this first then adapt it to fit into a better framework later, since we've been quite slow on SDE and/or ABC...

pints/toy/_stochastic_logistic_model.py
chonlei: Done.

pints/toy/_stochastic_logistic_model.py
chonlei: OK, I will change this to _simulate_raw. And not changing to docstring, because some of the other ones like the logistic model, its _simulate doesn't have a docstring either.

pints/toy/_stochastic_logistic_model.py
chonlei: Same as above.

docs/source/toy/stochastic_logistic_model.rst
chonlei: It's logistic.

pints/tests/test_toy_stochastic_logistic_model.py
chonlei: This is not done else where...

chonlei: Because we have run simulations...

chonlei: Why do we not want seeding here then?

pints/tests/test_toy_stochastic_logistic_model.py
chonlei: Just a simple test that tries to run and see it works.

pints/toy/_stochastic_logistic_model.py
chonlei: Think we might need a new interface for SDE.

chonlei: But tempted to leave this for #1067 instead of fixing it in this PR

pints/toy/_stochastic_logistic_model.py
chonlei: Similarly, for the name, I will leave this for #1067.

chonlei: But did change it to mention the args and return type now.

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertTrue(model._interpolate_values(time, raw_values, temp_time,
+                                                  params)[0] == 2)
+
+    def test_mean_variance(self):
+        # Check the mean is what we expected
+        model = pints.toy.StochasticLogisticModel(1)
+        v_mean = model.mean([1, 10], [5, 10])
+        self.assertEqual(v_mean[0], 10 / (1 + 9 * np.exp(-5)))
+        self.assertEqual(v_mean[1], 10 / (1 + 9 * np.exp(-10)))
+
```
MichaelClerx, line 85: Add test that variance returns notimplemented?

pints/tests/test_toy_stochastic_logistic_model.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_errors(self):
+        # Check the model is raising expected errors
+        model = pints.toy.StochasticLogisticModel(1)
+        times = np.linspace(0, 100, 101)
+
+        # parameters, times cannot be negative
```
MichaelClerx, line 91: All of these should be somewhere else!
Errors are part of a method's interface, so a test that mean.simulate() raises a certain error goes in the test for mean.simulate.

pints/toy/_stochastic_logistic_model.py
MichaelClerx: ok!

MichaelClerx (approved):
Thanks @chonlei and @jarthur36 !

Two tiny comments now, happy for this to be merged without further review once they're done

ben18785 (approved):
Thanks @chonlei happy for this to go in now (after tests complete). I'm not sure why the "build-and-test" check isn't executing?

ben18785:
@MichaelClerx Any idea why the "build-and-test" thing isn't running?

MichaelClerx:
> @MichaelClerx Any idea why the "build-and-test" thing isn't running?

I think that's an old test name that no longer exist. Will investigate!

Issue #1021, opened 2019-11-20
Add support for multiplicative Gaussian noise


rcw5890:
Add a likelihood in which the noise variance magnitude scales with the output, and add a diagnostic plot which may be useful to detect this type of noise.

MichaelClerx:
Closed in #1026

PR #1026, opened 2019-12-10
1021 multiplicative noise

rcw5890:
Here is the current code for #1021


pints/_log_likelihoods.py
```diff
@@ -521,6 +521,71 @@ def __init__(self, problem, sigma):
...
+
+    .. math::
+        \log{L(\theta, \sigma, \eta | \boldsymbol{x})} =
+            -\frac{n_t}{2} \log{2 \pi}
+            -\sum_{i=1}^{n_t}{\log{f(t_i, \theta)^\eta \sigma}}
+            -\frac{1}{2}\sum_{i=1}^{n_t}
+                \frac{(X(t_i) - f(t_i, \theta))^2}
+                {(f(t_i, \theta)^\eta \sigma)^2}
+
+    where ``n_t`` is the number of time points in the series.
```
MichaelClerx: Please add note saying this introduces 2*no new parameters

pints/_log_likelihoods.py
```diff
@@ -521,6 +521,71 @@ def __init__(self, problem, sigma):
...
+
+    This likelihood is applicable to a model given by
+
+    .. math::
+        X(t) = f(t; \theta) + f(t; \theta)^\eta v(t)
+
+    where v(t) is iid Gaussian:
+
+    .. math::
+        v(t) \sim \text{ iid } N(0, \sigma)
```
MichaelClerx: Please explain what eta and sigma are (and what dimensions they have)

pints/_log_likelihoods.py
```diff
@@ -521,6 +521,71 @@ def __init__(self, problem, sigma):
...
+                \frac{(X(t_i) - f(t_i, \theta))^2}
+                {(f(t_i, \theta)^\eta \sigma)^2}
+
+    where ``n_t`` is the number of time points in the series.
+
+    Extends :class:`ProblemLogLikelihood`.
+
+    Parameters
+    ----------
+    problem
```
MichaelClerx: `problem` --> ` ``problem`` `

pints/noise.py
```diff
@@ -230,3 +230,48 @@ def arma11_unity(rho, theta, sigma, n):
...
+
+    .. math::
+        v(t) \sim \text{ iid } N(0, \sigma)
+
+    The output magnitude ``f`` is required as an input to this function. The
+    noise terms are returned in an array of the same length as ``f``.
+
+    Parameters
+    ----------
+    eta
```
MichaelClerx: Please add ticks again

pints/noise.py
```diff
@@ -230,3 +230,48 @@ def arma11_unity(rho, theta, sigma, n):
...
+        The exponential power controlling the rate at which the noise scales
+        with the output
```
MichaelClerx: Tiny detail, but these sentences can be punctuated :-)

pints/noise.py
```diff
@@ -230,3 +230,48 @@ def arma11_unity(rho, theta, sigma, n):
...
+    sigma
+        The baseline standard deviation of the noise (must be greater than
+        zero)
+    f
+        A numpy array giving the time-series for the output over time
+    """
+    import numpy as np
+
+    if sigma <= 0:
+        raise ValueError('Standard deviation must be positive.')
```
MichaelClerx: "strictly positive"? or just "greater than zero"?

pints/noise.py
```diff
@@ -230,3 +230,48 @@ def arma11_unity(rho, theta, sigma, n):
     for i in range(1, n):
         e[i] = (1 - rho) + rho * e[i - 1] + v[i] + theta * v[i - 1]
     return e
+
+
+def multiplicative_gaussian(eta, sigma, f):
```
MichaelClerx: Does this work for multi-output data?

pints/residuals_diagnostics.py
```diff
@@ -173,6 +173,94 @@ def plot_residuals_autocorrelation(parameters,
...
+
+    When multiple samples of the parameters are provided (from an MCMC chain),
+    the residuals are calculated and plotted relative to the posterior median
+    of the solution outputs.
+
+    This function returns a ``matplotlib`` figure.
+
+    Parameters
+    ----------
+    parameters
```
MichaelClerx: ` ``parameters`` `

pints/tests/test_residuals_diagnostics.py
```diff
@@ -183,6 +183,44 @@ def test_plot_residuals_autocorrelation(self):
...
+            self.problem2
+        )
+
+        # Test that the multiple output figure has multiple axes
+        self.assertGreaterEqual(len(fig.axes), 2)
+
+        # Check the message when the input is wrong dimension
+        self.assertRaisesRegexp(
+            ValueError,
+            r'\`parameters\` must be of shape \(n_samples\, n_parameters\)\.',
```
MichaelClerx: You don't have to stick the whole message in here. Just enough so that we know that it's definitely the error we wanted to check for. (The exact wording isn't important, the fact that we check some of the wording is a bit of a side-effect)

pints/noise.py
rcw5890: Currently it doesn't - I edited the docstring to explicitly state that, and also check that `f` is a single time-series. For now, you would have to call the function separately for each output dimension. However, it wouldn't be too hard to generalize this function to handle multi output data in one call. For multi output data, I think we would want to allow a different value of `eta` and `sigma` for each output (the multiplicative likelihood certainly assumes that they can be different for each output).

Let me know if you think this would be a useful and worthwhile feature.

ben18785:
@rcw5890 How are you getting on with the suggestions that @MichaelClerx makes? Would be good to merge this soonish, if possible.

rcw5890:
Hi @ben18785 , as of d468929, I think I covered all of Michael's suggestions, except the question of making the noise generator multi-output.

Assuming there's no reasons not to, I'll just go ahead and add the multi-output functionality to the noise generator.

ben18785:
@rcw5890 Cool -- yep, add that then am sure we can merge.

pints/_log_likelihoods.py
```diff
@@ -521,6 +521,77 @@ def __init__(self, problem, sigma):
...
+
+    where ``n_t`` is the number of time points in the series.
+
+    Extends :class:`ProblemLogLikelihood`.
+
+    Parameters
+    ----------
+    ``problem``
+        A :class:`SingleOutputProblem` or :class:`MultiOutputProblem`. For a
+        single-output problem two parameters are added (eta, sigma),
```
MichaelClerx: ``eta`` etc

MichaelClerx:
Heteroscedastic, ey? Am I the only one who doesn't know what that means? :smile:

MichaelClerx:
In general, we're hoping PINTS is a useful tool for people who aren't statisticians (they can roll their own samplers with lots of experimental features etc.), so should aim to keep everything as simple / applied / non-statsy as possible.

Also, if you feel you've addressed a comment @rcw5890 feel free to hit the "Resolve conversation" button.

Good work!

rcw5890:
Ok, makes sense!

MichaelClerx:
Also, do you have the rights to merge? Or should I do it @rcw5890 ?

rcw5890:
I don't have merge authorization, so you'll have to be the one who handles that for now

MichaelClerx:
Aha! I've pressed some buttons. Does that change the situation?

rcw5890:
Yes, it looks like I now do have the permission, thanks!
@ben18785  let me know if you want to review the final commits too, or just go ahead and merge

MichaelClerx:
Thanks @rcw5890 !

If you write something like "Fixes #1021" or "Closes #1021" in any commit message (including the final merge commit) github will pick it up and automatically close the issue.

Also don't forget to hit "Delete branch" after merging :-)

Issue #282, opened 2018-03-23
Recreate Multinest in Pints


ben18785:
Again to avoid dependency issues it should be possible (and informative) to recreate the Multinest sampling algorithm in Pints. See [here](https://arxiv.org/abs/0809.3437).

ben18785:
Note to self: add in importance nested sampling as an option. See: https://arxiv.org/pdf/1306.2144.pdf

MichaelClerx:
Current notebook is here: https://github.com/pints-team/pints/blob/master/studies/multinest.ipynb (in master)

MichaelClerx:
For an alternative plan of adding a wrapper to existing software, see #161

ben18785:
See f42dad013fb54630bfb3804ddf6918a3aea309a2

ben18785:
Am adding to project for PINTS paper as nearly done.

MichaelClerx:
PR: #1319

PR #1033, opened 2020-02-10
CDF and inverse-CDFs for (log-)priors

ben18785:
To do #282, it is necessary to have cumulative distribution functions and their inverses for the priors.


ben18785:
@martinjrobins Could you have a look at this? My main issue concerns the multivariate normal -- in this case inverse cdfs are not available, so I created a pseudo-inverse-cdf function that can be used to convert samples uniformly distributed within the unit cube to those in the same space as the prior (this is necessary for multinest). Given that the cdf and icdf functions I use for this distribution are not actually what they say, I'm wondering about endowing each prior with functions "convert_to_unit_cube" and "convert_from_unit_cube". For all distributions but the mv-normal, the convert_to_unit_cube function would call the cdf and convert_from_unit_cube would call icdf. Do you think this sounds reasonable? This would avoid the mv-normal having functions that don't correspond to what they're named, at least.

ben18785:
Actually, have gone ahead and think this is ready for review.

pints/_log_pdfs.py
```diff
@@ -59,16 +59,31 @@ class LogPrior(LogPDF):
 
     Extends :class:`LogPDF`.
     """
-    def sample(self, n=1):
+    def cdf(self, x):
         """
-        Returns ``n`` random samples from the underlying prior distribution.
+        Returns the cumulative density function at a point ``x``.
```
martinjrobins: put the dimensions required for x in the docstring. from the docstring it looks like x has to be a single point? Would it be possible to generalise this to allow multiple points, similar to the cdf and icdf functions in scipy? Same with docstring of other new functions below

ben18785: @martinjrobins Good idea! I have now tweaked the cdf and icdf functions to allow them to handle multiple points (most required no changes since they use the scipy functions that generalise this way automatically).

martinjrobins: thanks @ben18785. Can we also put something in the docstring indicating the shape of x that is required. I.e. ``x`` is an n x d array, where n in the number of input samples and d is the dimension of parameter space

ben18785: Good point -- done!

ben18785:
@martinjrobins -- prod, if happy, can you merge please?

Issue #965, opened 2019-09-25
Organise examples folder into subfolders


ben18785:
The examples directory has so many example notebooks (great!) that it is now quite hard to navigate. Can we instead put these into folders? I'd suggest just using the names that appear in the file names to do this. So, sampling, optimisation and so on.

MichaelClerx:
Feel free!

ben18785:
I'm thinking about moving to a simple folder structure like:

- optimisation
- sampling
- other-functionality

I'd rather keep it simple, so things aren't miscategorised.

If we rearrange into folders, do we still want the prefix names? I'm guessing not. So, for example, we would have ``dream-mcmc`` rather than ``sampling-dream-mcmc`` as the named example.

What do people think?

MichaelClerx:
Sounds good!

MichaelClerx:
Closed in #1037

PR #1037, opened 2020-02-13
Restructured example notebook folder

ben18785:
Addresses #965 

Have done the following:

- Sorted example notebooks into separate folders
- Updated internal links between notebooks so they now work
- Updated ReadMe so it points to new location of notebooks
- Have updated slow notebooks and error notebooks


ben18785:
@MichaelClerx Would you be able to take a look at the "run-tests.py" file here? With this new structure of example notebook folders, I think it is not picking up all the files. Also, I am not sure how error-books and slow-books work here but, again, with my new folder structure for the notebooks, it seems to be missing these.

ben18785:
@martinjrobins Could you have a look at the bits I raise above? I don't really get how the testing works at the mo, so not sure what I need to change to rectify things.

examples/README.md
```diff
@@ -8,103 +8,101 @@ relevant code.
...
 
 ## Getting started
-- [Optimisation: First example](./optimisation-first-example.ipynb)
-- [Sampling: First example](./sampling-first-example.ipynb)
-- [Writing a model](./writing-a-model.ipynb)
-- [Writing a custom LogPDF](./writing-a-logpdf.ipynb)
-- [Writing a custom LogPrior](./writing-a-prior.ipynb)
+- [Optimisation: First example](./optimisation/first-example.ipynb)
+- [Sampling: First example](./sampling/first-example.ipynb)
+- [Writing a model](./other-functionality/stats/custom-model.ipynb)
```
MichaelClerx: Why are these 3 in "stats" ?

MichaelClerx:
Hey @ben18785 , made some fixes to the test scripts, which seem OK now.

Could we dispense with the `other-functionality` folder? I'm happy to have plotting, stats, and toy on the same level as optimisation and sampling
Bit worried we'll end up sticking everything else in `other-functionality`!

MichaelClerx:
Hey @ben18785 ! Did you have time to look at the comment above?

ben18785:
Sounds ok with me. In that case, can you go through all example notebooks and check that the links all point to the new folder locations?

MichaelClerx:
Done!

PR #1038, opened 2020-02-13
Toy model sensitivities

ben18785:
Addresses #298 by making it much easier to add toy model sensitivities (via the forward sensitivities method).

This involves the following:

- Adding methods to ``ToyModel`` class that handle the formation of the RHS of the outputs + sensitivities combined ODE
- Added sensitivities for Goodwin Oscillator and restructured those for Fitzhugh-Nagumo to handle new structure of toy models
- Added short codes to example notebooks to highlight sensitivity functionality of these models


ben18785:
@martinjrobins Before I add tests, what are your thoughts on this restructuring? It should make adding sensitivities for the remaining toy models trivial (which I increasingly believe will be important to have in our paper, where we benchmark NUTS vs other samplers).

pints/toy/_toy_classes.py
```diff
@@ -61,3 +79,71 @@ def suggested_times(self):
...
+        """
+        raise NotImplementedError
+
+    def _rhs_S1(self, y_and_dydp, t, p):
+        """
+        Forms the RHS of ODE for numerical integration to obtain both
+        outputs and sensitivities. This should be a vector of length
+        ``n_outputs`` + ``n_parameters``.
+        """
+        y = y_and_dydp[0:self.n_outputs()]
```
martinjrobins: maybe can make this a bit more readable by using temp variables for n_outputs and n_parameters?
```python
n = self.n_outputs()
m = self.n_parameters()
```

ben18785:
@martinjrobins -- sorry -- another prod, if happy, can you merge please? I don't have the power to (which is a good thing!).

pints/toy/_toy_classes.py
ben18785: Done

pints/toy/_toy_classes.py
```diff
@@ -46,6 +48,22 @@ class ToyModel(object):
...
 
+    def _dfdp(self, y, t, p):
+        """
+        Returns the derivative of the ODE RHS with respect to parameters.
+        This should be a matrix of dimensions ``n_parameters`` by
+        ``n_parameters``.
+        """
+        raise NotImplementedError
+
+    def jacobian(self, y, t, p):
```
MichaelClerx: What's the rationale for having `jacobian` public and the others private?

MichaelClerx: Some description of `y, t, p` ?

pints/toy/_toy_classes.py
```diff
@@ -61,3 +79,73 @@ def suggested_times(self):
...
+        d_dydp_dt = (
+            np.matmul(self.jacobian(y, t, p), dydp) +
+            self._dfdp(y, t, p))
+        return np.concatenate((dydt, d_dydp_dt.reshape(-1)))
+
+    def simulate(self, parameters, times):
+        """ See :meth:`pints.ForwardModel.simulate()`. """
+        return self._simulate(parameters, times, False)
+
+    def _simulate(self, parameters, times, sensitivities):
```
MichaelClerx: This is all very useful and a Good Idea, but specific to ODE models and could be useful to users implementing new toy ODE models.

Should we have a subclass called e.g. `ODEToyModel` that has all these methods, possibly public, and has more documentation on how to use this functionality?

pints/toy/_toy_classes.py
ben18785: Jacobians are useful for other things (like some gradient-based optimisers and samplers), so thought I'd give them a public method.

pints/toy/_toy_classes.py
ben18785: True, yes, you may be right that it's better to create a new class (I was mistakenly thinking that we just had ODE toy models, whereas we have others too).

MichaelClerx: What do you think @martinjrobins ?

ben18785: I'm resolving this as think we decided to have a ToyODEModel class

pints/toy/_toy_classes.py
ben18785: Done

pints/toy/_toy_classes.py
```diff
@@ -61,3 +64,134 @@ def suggested_times(self):
         """
         raise NotImplementedError
 
+    def simulate(self, parameters, times):
+        """ See :meth:`pints.ForwardModel.simulate()`. """
+        raise NotImplementedError
+
+
+class ToyODEModel(ToyModel):
```
MichaelClerx: This implements some of the `ForwardModelS1` methods, and so should extend that class?

Unless we want to split into toy ode models that do and don't do derivatives, but I'm happy to only have one that does ODEs where derivatives are available, as long as we make that clear in the docstring!

ben18785: I've now updated things as we discussed in the meeting, with a separate ToyODEModel class.

MichaelClerx:
Made a few small tweaks to the docstrings @ben18785 

One last thing: We have the names "parameters" and "times" in some parts of the ToyODEModel class, but also generally throughout pints. So should we use them here too?

The order of the arguments (y, t, p) makes a lot of sense to me too, but isn't consistent with the other methods in the class (parameters, times), so wondering about that as well...

Thoughts @ben18785 @martinjrobins ?

ben18785:
@MichaelClerx Thanks for having a look -- I am still working on this so probably should have had it as a draft PR; sorry. Good point re: parameters and times. Let me take a look..

ben18785:
@MichaelClerx This is ready to review now, please. I'd suggest we make the parameters thing a separate issue, if you feel strongly about it. Having looked, I agree it is a bit ungainly, but not something I think we should address here.

pints/tests/test_toy_fitzhugh_nagumo_model.py
```diff
@@ -59,6 +60,16 @@ def test_values(self):
         self.assertAlmostEqual(values[200, 0], 1.675726, places=6)
         self.assertAlmostEqual(values[200, 1], -0.226142, places=6)
 
+    def test_sensitivities(self):
+        model = pints.toy.FitzhughNagumoModel([2, 3])
+        parameters = [0.2, 0.7, 2.8]
+        times_finer = np.linspace(0, 20, 500)
+        sols, sens = model.simulateS1(parameters, times_finer)
+        f = interp1d(times_finer, sens[:, 0][:, 2])
```
MichaelClerx: What's going on here???

pints/tests/test_toy_goodwin_oscillator_model.py
```diff
@@ -40,6 +41,48 @@ def test_values(self):
         self.assertAlmostEqual(values[100, 1], 0.1779547, places=6)
         self.assertAlmostEqual(values[100, 2], 2.6074527, places=6)
 
+    def test_sensitivity(self):
```
MichaelClerx: Could you add a comment in each of these? I don't get what the difference between test_sensitivity and test_sensitivities is

pints/toy/_goodwin_oscillator_model.py
```diff
@@ -9,27 +9,26 @@
...
+        \dot{y} = k_2 x - m_2 y
 
-        \dot{z} = k3 y - m3 z
+        \dot{z} = k_3 y - m_3 z
 
-    Parameters are ``[k2, k3, m1, m2, m3]``.
+    Parameters are :math`[k_2, k_3, m_1, m_2, m_3]`. The initial conditions
+    are hard-coded at ``[0.0054, 0.053, 1.93]``.
 
     Extends :class:`pints.ForwardModel`, :class:`pints.toy.ToyModel`.
```
MichaelClerx: All modified models will need to update this line of the docstring

pints/toy/_lotka_volterra_model.py
```diff
@@ -38,13 +41,41 @@ class LotkaVolterraModel(pints.ForwardModel, ToyModel):
     References
     ----------
     .. [1] https://en.wikipedia.org/wiki/Lotka%E2%80%93Volterra_equations
```
MichaelClerx: This could just say `https://en.wikipedia.org/wiki/Lotka-Volterra_equations`

MichaelClerx: Without the %E2% etc

pints/toy/_toy_classes.py
```diff
@@ -61,3 +64,155 @@ def suggested_times(self):
         """
         raise NotImplementedError
 
+    def simulate(self, parameters, times):
```
MichaelClerx: This shouldn't be here, as it's already defined in `ForwardModel`

MichaelClerx: (which this class doesn't inherit from)

pints/toy/_toy_classes.py
```diff
@@ -61,3 +64,155 @@ def suggested_times(self):
...
+        y = y_and_dydp[0:self.n_outputs()]
+        dydp = y_and_dydp[self.n_outputs():].reshape((self.n_parameters(),
+                                                      self.n_outputs()))
+        dydt = self._rhs(y, t, p)
+        d_dydp_dt = (
+            np.matmul(dydp, np.transpose(self.jacobian(y, t, p))) +
+            np.transpose(self._dfdp(y, t, p)))
+        return np.concatenate((dydt, d_dydp_dt.reshape(-1)))
+
+    def simulate(self, parameters, times):
```
MichaelClerx: This one at least is functionally necessary...
@martinjrobins what do you think?

At the moment we've got the models themselves extending both ToyODEModel and either ForwardModel or ForwardModelS1 (they get to choose, which is important!)

With ToyModel we can avoid having a simulate(), but for ToyODEModel it's kind of necessary that we do. I think the cleanest solution would be to split ToyODEModel into a derivative-free and S1 class, and have those extend ForwardModel and ForwardModelS1. But what we have now works just as well and is probably easier...

MichaelClerx:
But looks great @ben18785 ! The model classes look much tidier & I like the bonus improvements. Just need to fix a few small issues and work out how best to do the inheritance

pints/tests/test_toy_fitzhugh_nagumo_model.py
ben18785: I've added comments to these -- to obtain the sensitivity at a predefined time it's easiest to interpolate the sensitivity series.

MichaelClerx: I don't understand. Why don't you request evaluation at the times you want?

pints/tests/test_toy_goodwin_oscillator_model.py
ben18785: Good point -- I've combined these into a single test method as it didn't make sense to keep these separate.

pints/toy/_goodwin_oscillator_model.py
ben18785: Good spot -- done!

pints/toy/_lotka_volterra_model.py
ben18785: changed

pints/toy/_toy_classes.py
ben18785: removed this

pints/toy/_toy_classes.py
ben18785: I think let's avoid splitting things up as it's already quite a complicated class ecosystem.

ben18785:
@MichaelClerx I've addressed your comments (hopefully). I may just have a quick blitz to try and get the last (possible) sensitivities in there...will buzz you when more reviewing needs to be done!

pints/tests/test_toy_fitzhugh_nagumo_model.py
ben18785: I could but find it easier to do it this way as it means I don't have to keep tabs on which observation maps to which time -- feel free to change if you feel strongly though.

ben18785:
@MichaelClerx Can you have a look now? I decided it's probably best to keep the remaining toy model sensitivities as a separate PR as this one is quite big now.

pints/tests/test_toy_fitzhugh_nagumo_model.py
MichaelClerx: Made it easier, but found a bug!

MichaelClerx:
Have simplified the tests @ben18785 
And found a small bug!
The first point returned is always 0, regardless of whether you ask for t=0 to be included in the returned times.
Probably an issue in ToyODEModel itself, as all three models that were using interp1d now show it. Have added a specific test for this issue to the fitzhugh model

ben18785:
@MichaelClerx Thanks. It looks like the sensitivity tests are failing now though...Could you have a look at this as Martin and I need these for looking at NUTS?

MichaelClerx:
@ben18785 that's right! The code in this PR can't handle the case where the initial time point isn't in the requested time points so this needs fixing!

ben18785:
Good spot. It looks like this is a requirement for `odeint` in [Scipy](https://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.integrate.odeint.html). I.e. that the first time point corresponds to the value of `y0`. As such, this will be an issue for ode integration in general - i.e. for the `simulate` as well as `simulateS1`, no?

In that case, I'd suggest we make a separate issue/PR to handle this corner case in general (and adapt the tests for this PR so that they pass) as it'll take some work to fix this. What do you think?

MichaelClerx:
I'd rather not merge in code that has known bugs :-/
I've got some time now though, so let me have a go at fixing it...

PR #1053, opened 2020-03-06
Removed unneeded method from Lotka-Volterra model.

MichaelClerx:
Had an extra `_simulate()` which is no longer needed


MichaelClerx:
Thanks @ben18785 !

PR #1057, opened 2020-03-10
ESS bug in summary object corrected

ben18785:


ben18785:
@martinjrobins Thanks, have modified because it was failing tests because ess is not a scalar quantity.

ben18785:
@martinjrobins @MichaelClerx thanks both!

Issue #615, opened 2019-01-29
Move sigma0 guessing code to hidden utility function


MichaelClerx:
We have code that guesses a covariance matrix based on an input vector in a few different places, probably makes sense to write a single utility function that does this and call it where needed.

fcooper8472:
Can you point me to these?

MichaelClerx:
Here's one:
https://github.com/pints-team/pints/blob/master/pints/_optimisers/__init__.py#L101-L135

And another:
https://github.com/pints-team/pints/blob/master/pints/_mcmc/__init__.py#L87-L104

I think a few more are hidden in branches at the moment

They don't do exactly the same. It'd be good to have some utility function (or multiple?) with a name that makes sense, so that we can standardise this sort of thing.

PR #1058, opened 2020-03-11
MCMC automatic initialisation

### LINKED ISSUE #1050: MCMC initialisation functions
At the moment, we either do one of the following:

1. Initialise all chains at non-random points
2. Randomly sample a starting value from some distribution, then pass these values to the MCMC samplers

Point 1. would make most Bayesian practitioners balk since it isn't best practice: by setting starting points _a priori_ you essentially bias samplers towards areas in parameter space that you know.

Point 2. is fine theoretically but causes practical problems: if the values sampled correspond to non-finite values of the log-posterior, then sampling doesn't execute.

To get around both of these issues, I'd suggest we consider the following (which I suspect is what Stan does):

- An MCMC sampler/controller can either accept an initial value set (as it currently does) or a function that will (randomly) sample initialisation points.
- When sampling starts, the function executes and passes this parameter set to the MCMC sampler: if the log-posterior is finite, all's hunky dory; if it's not, then the function resamples parameter values and tries these; and so on.
- If a given number of tries do not result in finite initial log-posteriors, then sampling does not execute. This number of tries would have sensible defaults (say 100 such sets are trialled) but could also be set by the user.

The reason for needing some leeway in terms of allowing quite a few tries before success is that, for more complex models, a more narrow region of parameter space typically has (effectively) non-zero probability density.

What do people think of this? As I say, I am a bit concerned at the moment that the way we initialise our chains isn't best practice (especially in the notebooks where we are quite slapdash about it). 

If people think this is reasonable, how would I go about doing this? I'd need to pass a function to the MCMCController, I suppose (not something I've done before in Python but looks straightforward enough).

ben18785:
Fixes #1050. Doesn't address sigma0 issue.

See also: #615


ben18785:
@martinjrobins @MichaelClerx This isn't tested yet but could you have a look and see what you think about the structure?

pints/_mcmc/__init__.py
```diff
@@ -296,7 +296,8 @@ class MCMCController(object):
         :class:`HaarioBardenetACMC` is used.
     """
 
-    def __init__(self, log_pdf, chains, x0, sigma0=None, method=None):
+    def __init__(self, log_pdf, chains, x0=None, sigma0=None, method=None,
```
MichaelClerx: Would it make sense to remove the `init_function` for this and re-use `x0` ?

pints/_mcmc/__init__.py
```diff
@@ -600,6 +623,34 @@ def run(self):
         # Start sampling
         timer = pints.Timer()
         running = True
+
+        # initialisation (for single chain methods only and only if prior
```
MichaelClerx: I'm not super excited about sticking more stuff in the controller. It's a very complex class already, and this is a really easy chunk to break off and do externally?

ben18785: I think we should add this because it's a convenience function that is really useful in practice. Chon himself said it would be useful because it would save him having to take care of this. Otherwise, it makes people lazy -- we currently are initialising wrongly (as per best practices) in about 95% of our notebooks; I don't feel comfortable about this. I suspect if we don't rectify this that it will come up in review.

ben18785: I do agree that the run() function of the controller is complicated too and, as I've said before, I would like to simplify it by splitting up into separate methods. But that's for another day.

ben18785:
> I like the general idea of this. only thing I would change is that I don't like the name of `init_function`. Since the function you are talking about returns samples, I would call it a sampler, or even better specify that it has to be a `LogPrior` class with a `sample` method.

@martinjrobins Thanks. Re: the naming, I considered calling it a LogPrior but then realised that this would be quite confusing because the initialisation function need not necessarily be _the_ prior of the model; the model itself need not be Bayesian and not need a prior in the first place.

So, that's why I opted to call it what I did. I'd actually prefer to call it just `init` since this mirrors Stan and, I think, would make it easier for people coming across from Stan. As long as we keep it clear in the documentation that this need be of class LogPrior, do you think that'd be ok?

pints/_mcmc/__init__.py
MichaelClerx: Most of the things in `run` can't be isolated (I don't agree with that tool's analysis!), but this bit easily can. Maybe stick it in a hidden method inside then controller then and just call that?
e.g.
```
def __init__(self, x0, ...):
    if isinstance(x0, pints.LogPrior):
        self._log_prior = x0
        self._x0 = None
    elif len(x0) etc.

def run(..)
    ...
    if self._x0 is None:
        self._sample_x0()
    ...
def _sample_x0():
    set x0 based on log_prior passed at construction time
```

ben18785: Cool -- thanks, will do.

MichaelClerx:
I think `init`'s an even worse name! Why can't we just use `x0` for this? You have to specify either one or the other, both is an error, and if x0 isn't a point it's a function used to generate points

pints/_mcmc/__init__.py
```diff
@@ -311,14 +311,23 @@ def __init__(self, log_pdf, chains, x0, sigma0=None, method=None):
...
-        if len(x0) != chains:
-            raise ValueError(
-                'Number of initial positions must be equal to number of'
-                ' chains.')
-        if not all([len(x) == self._n_parameters for x in x0]):
-            raise ValueError(
-                'All initial positions must have the same dimension as the'
-                ' given LogPDF.')
+        self._x0_isfunction = hasattr(x0, '__call__')
+        if self._x0_isfunction:
```
MichaelClerx, line 315: I think you can use `if callable(x0):`

MichaelClerx: Or even just check if it's a LogPrior, otherwise assume it's a number

ben18785: Yep, good point. Have done the latter.

ben18785:
@martinjrobins @MichaelClerx @chonlei Can some of you have a look at this? I've added the following functionality to the MCMCController:

- if x0=None, then initialise by sampling from the LogPrior, if the log_pdf is of type LogPosterior; if it's not, then initialise all parameters on a U(-2, 2) scale, which is what Stan does by default
- if x0=LogPrior_a (which can be different to the LogPrior_b constituting the posterior), then LogPrior_a is used for initialisation

In all cases, the initialisation is tried a fixed number of tries (each "try" constitutes a new set of samples from whatever function is used to generate an initial point set): if this fixed count is reached, an error about "failing to initialise".

Of course, these changes still allow x0 to be a fixed set of points.

pints/_mcmc/__init__.py
```diff
@@ -300,22 +309,65 @@ class MCMCController(object):
...
+        if x0 is None:
+            # use prior if one's available
+            if isinstance(log_pdf, pints.LogPosterior):
+                x0 = log_pdf.log_prior()
+            # do what Stan does and initialise all parameters on U(-2, 2) scale
+            else:
+                x0 = pints.UniformLogPrior([-2] * self._n_parameters,
+                                           [2] * self._n_parameters)
+        if isinstance(x0, pints.LogPrior):
+            init = [x0.sample() for i in range(chains)]
```
MichaelClerx, line 339: https://pints.readthedocs.io/en/latest/log_pdfs.html#pints.LogPrior.sample

log prior sampling takes `n_samples` as argument

pints/_mcmc/__init__.py
```diff
@@ -300,22 +309,65 @@ class MCMCController(object):
...
+            raise ValueError('Number of chains must be at least 1.')
+
+        # Get functions which can generate initial position or set initial
+        # position points if numerical values given
+        self._x0_isfunction = False
+        if x0 is None:
+            # use prior if one's available
+            if isinstance(log_pdf, pints.LogPosterior):
+                x0 = log_pdf.log_prior()
+            # do what Stan does and initialise all parameters on U(-2, 2) scale
```
MichaelClerx, line 334: I'd prefer raising a ValueError at this point, saying users must specify an x0 or provide a log posterior. I've had some experience with people leaving things at their default values and then running into problems, so better to make them aware there's an important choice to be made

pints/_mcmc/__init__.py
```diff
@@ -300,22 +309,65 @@ class MCMCController(object):
...
         # sampler will see only the transformed search space and will know
         # nothing about the model parameter space.
         if transform is not None:
             # Convert log pdf
             log_pdf = transform.convert_log_pdf(log_pdf)
 
+        # Store transform for later detransformation: if using a transform, any
+        # parameters logged to the filesystem or printed to screen should be
+        # detransformed first!
+        self._transform = transform
```
MichaelClerx, line 363: The transform code was handled first, so that transformed log_pdfs would be used throughout. With this re-ordering we sample initial points from the untransformed space?

pints/_mcmc/__init__.py
```diff
@@ -300,22 +309,65 @@ class MCMCController(object):
...
+
+        # Store function
+        self._log_pdf = log_pdf
+
+        if transform is not None:
             # Convert initial positions
-            x0 = [transform.to_search(x) for x in x0]
+            init = [transform.to_search(x) for x in init]
```
MichaelClerx, line 370: Another transformation thing to check! I'd guess we want to sample from the transformed log prior, so that this conversion is unnecessary - so this `to_search` should happen only for an explicitly set x0 ?

pints/_mcmc/__init__.py
```diff
@@ -878,6 +932,37 @@ def samplers(self):
...
+        """
+        initialised_finite = False
+        init_fn = self._init_fn
+        current_active = list(active)
+        n_tries = 0
+        x0 = []
+        transform = self._transform
+        while not initialised_finite and (n_tries <
+                                          self._max_initialisation_tries):
+            xs = [init_fn.sample()[0] for i in current_active]
```
MichaelClerx, line 948: sample(n)

pints/_mcmc/__init__.py
```diff
@@ -878,6 +932,37 @@ def samplers(self):
...
         return self._samplers
 
+    def _sample_x0(self, active, evaluator):
+        """
+        Samples an initial starting point for the chains using the supplied
+        `init` function.
+        """
+        initialised_finite = False
+        init_fn = self._init_fn
+        current_active = list(active)
```
MichaelClerx, line 942: Are they not all active by default?

pints/_mcmc/__init__.py
```diff
@@ -878,6 +932,37 @@ def samplers(self):
         """
         return self._samplers
 
+    def _sample_x0(self, active, evaluator):
+        """
+        Samples an initial starting point for the chains using the supplied
+        `init` function.
+        """
+        initialised_finite = False
+        init_fn = self._init_fn
```
MichaelClerx, line 941: Would pass this as an argument, along with the other args

pints/_mcmc/__init__.py
```diff
@@ -371,6 +398,7 @@ def __init__(
                 ok = False
             if not ok:
                 raise ValueError('Given method must extend pints.MCMCSampler.')
+        self._method = method
```
MichaelClerx, line 401: Why this line?

MichaelClerx (changes requested):
Few small requests.

But also, seeing how involved this gets, would it be a better idea to just have a utility function that can do this? E.g. `x0 = pints.sample_starting_points(n, log_prior (or posterior), max_tries=100, parallel=False)` ?
And then have the user call it?

ben18785:
@MichaelClerx Thanks for the comments -- really helpful.

Re: utility function, I can see arguments for why this would be useful (making people be explicit about how they do initialisation) and arguments against (at the moment, because of the effort of doing random initialisation -- which is the correct way to do it -- my impression is that lots of people do it incorrectly and just start all chains at the same point).

Because of this, I'm more in favour of keeping this in the controller but documenting it.

But, I would like to hear what users like @chonlei @DavAug @martinjrobins think.

MichaelClerx:
Dave proposes the general rule: "where possible try to force the user to do it right", 

which would mean 

- the above belongs in the controller, 
- methods that rely sensitively on other correct instantiation (e.g. sigma0) should warn or even raise an error if no default is provided?

pints/_mcmc/__init__.py
ben18785: I would prefer to keep this as is: this is the default used by the major probabilistic programming languages, so think it's not really unsafe (what's more unsafe is allowing a user to specify a fixed x0 but we need that).

pints/_mcmc/__init__.py
ben18785: done

pints/_mcmc/__init__.py
ben18785: It's tricky because the initialisation function is defined during instantiation. So, I'd end up passing it to run (or calling it from within run) then passing to _sample_x0. So, either case, I think it'd have to be a self_.init_fn job no?

pints/_mcmc/__init__.py
ben18785: This is a good point but I just checked: doing log_pdf = transform.convert_log_pdf(log_pdf) transforms the pdf but does not transform the prior itself (indeed, as far as I can tell, a LogPosterior converted to a TransformedLogPDF does not have its own TransformedLogPrior). So, I think something like this needs to happen.

pints/_mcmc/__init__.py
ben18785: The init values are stored in a temp variable "init". It isn't stored to any object. Then a few lines down, they're converted to the correct space. These init values are needed for sigma0, so not sure how we could do the transform first.

pints/_mcmc/__init__.py
ben18785: Nope. Some samplers have intermediate steps, so we return a list of those 'active' samplers still waiting for their intermediate iterations to complete.

pints/_mcmc/__init__.py
ben18785: We use it to initialise each sampler in `run`.

pints/_mcmc/__init__.py
MichaelClerx: On the first step though :D 
But I'm happy for this to stay in

pints/_mcmc/__init__.py
MichaelClerx: @chonlei can you look at these last two transformation-related points? The easiest place is not here but on the "Files changed" tab.

pints/_mcmc/__init__.py
MichaelClerx: But you're already creating the samplers just below: `self._samplers = [method(x, sigma0)`

pints/_mcmc/__init__.py
```diff
@@ -1022,6 +1107,15 @@ def set_max_iterations(self, iterations=10000):
                     'Maximum number of iterations cannot be negative.')
         self._max_iterations = iterations
 
+    def set_max_initialisation_tries(self, max_initialisation_tries):
+        """
+        Sets number of attempts to initialise samplers by sampling.
```
MichaelClerx, line 1112: I'm a bit worried how this will look to a first (or 10th?) time user. Certainly needs a much clearer docstring. The current one assumes everyone's already aware that this is something the controller will do.

MichaelClerx: Plus it doesn't always do it. So needs some condition :D

pints/_mcmc/__init__.py
```diff
@@ -645,6 +680,25 @@ def run(self):
...
             n_samples = [0] * self._n_chains
 
+        # initialisation
+        if self._x0_isfunction:
+            if self._single_chain:
+                initialised_finite, init = self._sample_x0(active, evaluator)
+            else:
+                chain_r = list(range(self._n_chains))
+                initialised_finite, init = self._sample_x0(chain_r, evaluator)
+            if not initialised_finite:
```
MichaelClerx, line 690: I'd just raise this error in `_sample_x0` instead of returning a bool to check?

pints/_mcmc/__init__.py
MichaelClerx: Even a case to be made here for keeping this fixed for the controller, to avoid making it too clunky / adding too many bells & whistles?

MichaelClerx: We could even do this:

1. Write a public `pints.sample_initial_point` method that does the sampling.
2. Have the controller use it with some fixed sensible max
3. Let the user who wants to customise things generate their own x0 _using this method_, so that they can pass in all the arguments and customisation they like, without cluttering the controller interface

MichaelClerx:
What about sigma0? Are we assuming a one-size fits all here then?

pints/_mcmc/__init__.py
chonlei: For the first point, the convert_log_pdf() only considers the PDF as one thing (it doesn't care if it's a PDF/likelihood/posterior). There is the TransformedLogPrior object which can be called via convert_log_prior() to do the conversion first(?), then we could sample from there (but note that there's no way that we can sample directly from the transformed prior easily, so it samples from the model/original parameter space and does the transform afterward).

chonlei: But I see Michael's point that we've now shifted things to "transforming parameters in the middle of the code" which is something that i was trying to avoid. I don't see the code will break things at the moment, but I somehow wanted to have transformation happens only at the start and end, so that the middle bit doesn't involve any transformation in case of (future) confusion...

pints/_mcmc/__init__.py
MichaelClerx: Maybe create the samplers in `run()` ?

The "needs sensitivities" should probably be a static method so we can call it without having to create the samplers?

This relates to design questions about whether or not we want controllers to be single or multi use. It might actually make sense to let them be run more than once? Continue where you left off? That kind of thing. Separate issue, but doing this might be a step towards that!

MichaelClerx: So replace

def needs_initial_phase(self):
    ...
    
with either

@staticmethod
def needs_initial_phase():
    ...
    
```
@staticmethod
def needs_initial_phase():
    ...
```

```
@classmethod
def needs_initial_phase(cls):
    ...
```

ben18785:
Closed in favour of an initialisation method presented in #1322

PR #1060, opened 2020-03-12
German credit logistic examples

ben18785:
The way the model is coded up:

- If data aren't supplied then it downloads the data from source
- If they are, then it doesn't


MichaelClerx:
Hey @ben18785 
Thanks for coding it up, but I really don't want PINTS to download stuff!
I'd expect it from a phone app, or an email client, or whatever, but if a fitting library starts connecting to the internet I'd personally get very suspicious

ben18785:
How about if we made it an explicit method, so you'd do something like:

```
model = GermanCreditLogPDF()
model.download_data()
```

That way, people have complete control and transparency over what is happening. What do you think?

MichaelClerx:
I don't see why we wouldn't just include it?
(I can have a go at that, if you like, there might be one or two things to do with installing it that we need to work out)

ben18785:
@MichaelClerx We can't really include for two reasons, I think:

- The datasets can massively increase the size of Pints (this will be especially true if we move to "Pints for big datasets" in the future
- We don't own the datasets and so shouldn't be distributing them

@martinjrobins What are your thoughts here?

martinjrobins:
I'm perfectly happy with this toy model downloading a dataset. In terms of managing user expectations re. downloads, I would just put a note in the API documentation, then they will expect it.

martinjrobins:
yea, I agree with @ben and would be hesitant about bundling it. The original author donated it to the UCI database, not to us! I'm really not sure how data licencing works.....

MichaelClerx:
> I'm perfectly happy with this toy model downloading a dataset. In terms of managing user expectations re. downloads, I would just put a note in the API documentation, then they will expect it.

Except that nobody reads that :D

Can we add a switch to this model when you create it that says something like `allow_download = False` that you explicitly need to set to true?

ben18785:
@MichaelClerx Good idea with the flag. I have updated this now. See what you think. @martinjrobins thanks!

pints/tests/test_toy_german_credit_hierarchical_logpdf.py
```diff
@@ -0,0 +1,94 @@
...
+from scipy import stats
+
+
+class TestGermanCreditHierarchicalLogPDF(unittest.TestCase):
+    """
+    Tests the logpdf toy distribution from fitting a hierarchical logistic
+    model to German credit data.
+    """
+    @classmethod
+    def setUpClass(cls):
```
MichaelClerx: Why are the tests downloading too?
You don't need to test if downloading works by downloading twice :D

MichaelClerx: In fact, you can't test if the downloaded value is correct because we don't know a-priori what it's going to be.
It should be possible for them to correct errors in the data, for example.

You can only test if it's still well-formed, e.g. can be read as data in the format we expect. So array shapes and sizes, that sort of thing.

pints/toy/_german_credit.py
```diff
@@ -0,0 +1,124 @@
...
+        vector of coefficients of length 25.
+
+    References
+    ----------
+    .. [1] "UCI machine learning repository", 2010. A. Frank and A. Asuncion.
+           https://archive.ics.uci.edu/ml/datasets/statlog+(german+credit+data)
+    .. [2] "The No-U-Turn Sampler:  Adaptively Setting Path Lengths in
+           Hamiltonian Monte Carlo", 2014, M.D. Hoffman and A. Gelman.
+    """
+    def __init__(self, x=None, y=None, download=False):
```
MichaelClerx: Why do we want to allow x and y to be overridden?

pints/toy/_german_credit.py
```diff
@@ -0,0 +1,124 @@
...
+        log_prob += -1 / (2 * self._sigma_sq) * np.dot(beta, beta)
+        return log_prob
+
+    def data(self):
+        """ Returns data used to fit model. """
+        return self._x, self._y
+
+    def download_data(self):
+        """ Downloads data from [1]. """
+        url="http://archive.ics.uci.edu/ml/machine-learning-databases/statlog/german/german.data-numeric" # noqa
```
MichaelClerx: You can write e.g.
```
url = ('https://lalalala/123'
       '/more/stuff/here')
```

pints/toy/_german_credit.py
```diff
@@ -0,0 +1,124 @@
...
+    def __call__(self, beta):
+        log_prob = sum(-np.log(1 + np.exp(-self._y * np.dot(self._x, beta))))
+        log_prob += -1 / (2 * self._sigma_sq) * np.dot(beta, beta)
+        return log_prob
+
+    def data(self):
+        """ Returns data used to fit model. """
+        return self._x, self._y
+
+    def download_data(self):
```
MichaelClerx: Would make this hidden, and only called by the constructor

pints/tests/test_toy_german_credit_hierarchical_logpdf.py
ben18785: Yep, good points; ha, was being dense before! I have now amended and added tests of shapes. The dataset and repo and old and well-used, so not anticipating any changes to it though.

pints/toy/_german_credit.py
ben18785: Not sure I do? When _both_ (x,y) and download=True are selected then I want to return an error (not overwrite them).

pints/toy/_german_credit.py
ben18785: Thanks -- been looking for a better way to handle these!

pints/toy/_german_credit.py
ben18785: Good point -- done!

ben18785:
@MichaelClerx Any idea how to handle these coverage reports? I know that those lines are covered since tests are run with both python2 and python3.

Issue #1065, opened 2020-03-14
Add eight schools example: a hierarchical model used to test MCMC convergence


ben18785:
See, for example, [here](https://arxiv.org/pdf/1903.08008.pdf) and [this](https://github.com/avehtari/rhat_ess) very useful repo.

Add centered and non-centered models as suboptions.

ben18785:
Closed by 3f39932b04757b6c14a202eac970d23793a0e067

PR #1066, opened 2020-03-15
Eight schools: centered and non-centered models

ben18785:
Began to address #1065 -- think it's definitely best to keep these as separate models but to include only a single example notebook as they only really make sense next to one another.

Need to do:

- Add gradients for half-Cauchy log prior as this is a dependency for this model
- Add non-centered model


pints/toy/_eight_schools_centered.py
```diff
@@ -0,0 +1,96 @@
...
+from __future__ import print_function, unicode_literals
+import numpy as np
+import pints
+
+from . import ToyLogPDF
+
+
+class EightSchoolsCenteredLogPDF(ToyLogPDF):
+    r"""
+    The classic Eight Schools example from [1]_. This model was used to
```
MichaelClerx: Should we be careful using the word 'model' here? To show the implementation is of a model + pdf?

ben18785: I personally think it's ok since model is used by lots of people all over the place. It also really _is_ a model (albeit a statistical model of school performance on SATs); haha.

ben18785:
@MichaelClerx @martinjrobins If you have the chance, give this a thorough code-dissecting! (My code always needs dissecting!)

MichaelClerx:
Should that stan interface notebook be in here?

ben18785:
@MichaelClerx Thanks -- yep, have moved the Stan interface stuff out into a separate branch.

pints/tests/test_toy_eight_schools_logpdf.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+#
+# Tests the cone distribution.
```
MichaelClerx: No it doesn't :D

pints/tests/test_toy_eight_schools_logpdf.py
```diff
@@ -0,0 +1,76 @@
...
+#
+# This file is part of PINTS.
```
MichaelClerx: Update to new template please

pints/toy/_eight_schools.py
```diff
@@ -0,0 +1,153 @@
...
+from __future__ import print_function, unicode_literals
+import numpy as np
+import pints
+
+from . import ToyLogPDF
+
+
+class EightSchoolsLogPDF(ToyLogPDF):
+    r"""
+    The classic Eight Schools example from [1]_. This model was used to
```
MichaelClerx: This makes me feel like I'm supposed to know about this example, which is not what we expect from pints' target audience?

pints/toy/_eight_schools.py
```diff
@@ -0,0 +1,153 @@
...
+        y_j\sim\mathcal{N}(\theta_j, \sigma_j).
+
+    Note that, in the non-centered case, the parameter samples correspond to
+    theta_tilde rather than theta.
+
+    Extends :class:`pints.toy.ToyLogPDF`.
+
+    Parameters
+    ----------
+    mu : float
```
MichaelClerx: These aren't right

MichaelClerx:
Not sure what I should be looking for!

You do seem to create a lot of priors?!

pints/toy/_eight_schools.py
ben18785: Good point. Have changed tack a bit with the wording now. See what you think.

pints/toy/_eight_schools.py
ben18785: Not sure I understand, sorry? Can you elaborate?

MichaelClerx: They're not the constructor arguments

Issue #1070, opened 2020-03-18
Convenience functions for transforming parameters


ben18785:
I have been using Stan in anger the past couple of days, which has been enlightening and also a little frustrating.

The reason for this frustration is that a lot of our methods (all sensitivity-based approaches, for example) do not play well in problems where parameters have constraints; that is, cases when there are boundaries beyond which the log-probability is infinite. Because of this, in using models with uniform priors, I have had methods throwing errors mid-sampling.

To increase the robustness and to simplify the introduction of future algorithms (because we don't need to worry about any boundary issues), I strongly suggest we do what Stan and PyMC3 do -- transform all parameters to be unconstrained and then do sampling on this transformed space. In addition to reducing potential bugs, there is theory that sampling is more efficient on the transformed space.

This is actually quite simple to do (at least statistically) and involves the following:

- Transform constrained parameters to unconstrained space (for uniform parameters, one way to do this is to transform to the unit interval then take the logit function)
- Add a Jacobian term to the log probability (and to the gradient, if necessary) that accounts for this transformation
- Sample on transformed space
- Transform samples back to their original (constrained) parameter space

I'm happy to do this and think it is a real priority as will reduce a whole host of bugs awaiting our users. My question is, how is best to do this @MichaelClerx @martinjrobins ? We could do something like:

- Before sampling occurs, all parameters are checked for whether they are bounded; this is recorded in some sort of class which has methods like: transformed(), istransformed(), untransform() and jacobian() which can be queried for each parameter
- When sampling starts, each time a function is evaluated the jacobian term is added to the log prob (and gradient if necessary)
- When sampling ends, we perform untransform() on all parameters where it is necessary

MichaelClerx:
We'd have to think very very carefully about this. For the problems I've worked on the transform made a lot of difference, but had to be made in a very problem-specific way. For me that's outside the scope of PINTS; precisely the kind of thing a user (who understands the problem) should do before passing it off to PINTS

MichaelClerx:
Had several discussions with Gary about this as well, see e.g. https://github.com/pints-team/pints/issues/31

My advice is still the same: Write a model wrapper around your forward model. See the example here https://github.com/pints-team/pints/blob/master/examples/optimisation/transformed-parameters.ipynb

This
- Gives the user full control over the transformation
- Is less work for the user than having to always implement transformations
- Stops us from having to update _all of pints_ to always expect transformations

MichaelClerx:
[...]

PR #1077, opened 2020-03-19
Stan interface

### LINKED ISSUE #1071: Stan interface
Should be easy to do and worthwhile

ben18785:
Closes #1071


ben18785:
@MichaelClerx @martinjrobins Let me know what you think. The thing I'm least sure about is the best way to define boundaries on parameters (this is one of the reasons I started the discussion in #1070). Have a look at the example notebook to see what I mean -- ideally, we wouldn't have to pick some arbitrarily large value to constrain sampling of unconstrained variables...

ben18785:
By the way, there are no tests yet because I want to see what you think about the structure.

martinjrobins:
Hi @ben18785. In general looks great, couple of thoughts from my side:
- Can the StanInterface class directly take stan code and compile it, run the first few samples etc. internally? These seem like details that would be common for every model, that the user doesn't need to worry about
- Why can't pints sample in unconstrained space? Surely you can start a sampler without providing boundaries?

MichaelClerx:
Yeah I don't get the constraints bit either! We've never used constraints in sampling so far

MichaelClerx:
Can we maybe have a separate skype call about this early next week? I don't want to look at the code before I know
- why we need a stan interface
- why this would be better for our users than just using stan directly
- what the impact will be for the "continuous integration" testing (travis).
- how maintainable any stan/pints bridge will be. Most interfaces between different pieces of software are fickle, fiddly things, that get disowned after a couple of years

I'm more interested in going down the nuts sampler reimplementation route, figuring out if it's really only the compiled-vs-interpreted thing that causes the performance difference, and then just ending it there by telling our users "if you want a nuts sampler on a non-ode problem with sensitivities available and you don't mind the software transforming your parameters etc., then please use stan, because it's much better at that than pints."

MichaelClerx:
I mean, if you've got a problem that's easy to formulate and solve in stan, but pints needs to wring itself through some hoops to accommodate it, then chances are high none of the tools in the pints "environment" will be very useful either, while the stan userbase will have lots of nice tools for your stan-type problem?

MichaelClerx:
Let's discuss this in this PINTS meeting maybe, we've increasingly been adding methods that come with their own PDF, not their own ODE model, to show why/when they're useful. I liked the ODE niche and people seemed to think it was an underdeveloped aspect of the field so a good direction to go?

ben18785:
@martinjrobins @MichaelClerx Ready for your view and review!

pints/interfaces/_interface_classes.py
```diff
@@ -0,0 +1,25 @@
...
+# This file is part of PINTS (https://github.com/pints-team/pints/) which is
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+
+
+class InterfaceLogPDF(pints.LogPDF):
```
MichaelClerx: Not sure I see the value of this base class?

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,101 @@
...
+from collections import Counter
+import pystan
+
+from . import InterfaceLogPDF
+
+
+class StanLogPDF(InterfaceLogPDF):
+    def __init__(self, stan_code, stan_data):
+
+        print("Compiling Stan model then test running...")
```
MichaelClerx: Are these staying?

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,101 @@
...
+        sm = pystan.StanModel(model_code=stan_code)
+        stanfit = sm.sampling(data=stan_data, iter=1, chains=1,
+                              verbose=False, refresh=10,
+                              control={'adapt_engaged': False})
+        print("Stan model compiled and runs ok...ignore various warnings.")
```
MichaelClerx: If we really wanted we could probably capture & hide this output?

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,101 @@
...
+        self._names, self._index = self._initialise_dict_index(names)
+        self._long_names = names
+        self._counter = Counter(self._index)
+        self._dict = {self._names[i]: [] for i in range(len(self._names))}
+
+    def __call__(self, x):
+        dict = self._dict_update(x)
+        try:
+            return self._log_prob(self._u_to_c(dict), adjust_transform=True)
+        # if Pints proposes a value outside of Stan's parameter bounds
```
MichaelClerx: Some way to check this? Or some more specific error we could catch?
RuntimeError can be a lot of things

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,101 @@
...
+        except RuntimeError:
+            return -np.inf
```
MichaelClerx: That's very negative, Ben. These are tough times

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,101 @@
...
+            if num < 0:
+                names_short.append(name)
+            else:
+                names_short.append(name[:num])
+        names_long = list(names_short)
+        names_short = list(dict.fromkeys(names_short))
+        index = [names_short.index(name) for name in names_long]
+        return names_short, index
+
+    def pickled_form(self):
```
MichaelClerx: Whence the pickling?

pints/interfaces/_stan.py
martinjrobins: yea, this caught me out recently! Need to catch the specific error

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,101 @@
...
+import pystan
+
+from . import InterfaceLogPDF
+
+
+class StanLogPDF(InterfaceLogPDF):
+    def __init__(self, stan_code, stan_data):
+
+        print("Compiling Stan model then test running...")
+        sm = pystan.StanModel(model_code=stan_code)
```
martinjrobins: It would be good if we can allow for pickling of the compiled model, something like the below (where the user optionally passes in a filename to pickle to). An even better way might be to hash the input code and generate a filename based on that?

```suggestion
        if pickle_filename:
            if os.path.isfile(pickle_filename):
                sm = pickle.load(open(pickle_filename, 'rb'))
            else:
                sm = pystan.StanModel(model_code=stan_code)
                pickle.dump(sm, open(pickle_filename, 'wb'))
        else:
            sm = pystan.StanModel(model_code=stan_code)
```

MichaelClerx:
Pickling?! You've reached peak python! :D

ben18785:
Hahaha. It wasn’t my code though sadly — Martin was the boss of these changes!

ben18785:
@MichaelClerx Any idea how I get Pystan to be installed for these tests? At the moment, it's in "extras_require" in setup.py but, for some reason, that doesn't seem to be installing.

MichaelClerx:
I guess we'll need to update the "workflows" (which replace the travis file still in this branch).
Could you "update branch" again and then I can have a look after?

MichaelClerx:
Nice!

ben18785:
@MichaelClerx This seems to fail unit testing because of a known issue of getting Pystan working on Windows. See [here](https://github.com/stan-dev/pystan/issues/680). It seems that we would need a particular compiler type (m2w64.toolchain) for this to work on Windows. Any ideas?

MichaelClerx:
> @MichaelClerx This seems to fail unit testing because of a known issue of getting Pystan working on Windows. See [here](https://github.com/stan-dev/pystan/issues/680). It seems that we would need a particular compiler type (m2w64.toolchain) for this to work on Windows. Any ideas?

Oof! We _could_ add a test that uses miniconda with mingw I suppose, but it probably makes more sense to disable the tests if pystan is not found?

So put something at the top of the test module, near the imports, like

```
try:
  import pystan
  have_stan = True
except ImportError:
  have_stan = False
```

and then annotate all tests classes with

```@unittest.skipIf(not have_stan, 'PyStan not found')```

(or however you capitalise pystan)

Examples in nelder_mead and nuts tests

**Comment by MichaelClerx** [comment]
Pointing out known issues with toolchains is a very RSE thing to do - just saying

**Comment by ben18785** [comment]
@MichaelClerx This is now ready for a going over. Cheers!

File: examples/interfaces/stan.ipynb [hunk 8]
```diff
@@ -0,0 +1,992 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Interfacing with Stan\n",
+    "This notebook explains how it is possible to interface with [Stan](https://mc-stan.org/) to make use of their awesome library for log-probability calculations, gradient calculation (via autodiff) and their large zoo of probability distributions.\n",
```
MichaelClerx, line 8: Can we reword this to reflect the actual code more? I think we can only interact with a log pdf at the moment?

examples/interfaces/stan.ipynb
```diff
@@ -0,0 +1,992 @@
...
+    "\n",
+    "One thing to be mindful of is that the interface below only allows the log probability to be accessed from Stan objects up to an additive constant. Another is that Stan operates on the unconstrained scale, so the PINTS interface to Stan passes and receives parameter draws on this scale.\n",
```
MichaelClerx, line 10: > the log probability

from the last sentence, it's not clear to me that there is going to be a log probability, so needs a bit more introduction!

examples/interfaces/stan.ipynb
```diff
@@ -0,0 +1,992 @@
...
+    "results = pints.MCMCSummary(chains=full_chains, time=(end-start),\n",
+    "                            parameter_names=stanmodel.names())\n",
+    "print(results)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Constrained $\\sigma$\n",
```
MichaelClerx, line 245: Replace \sigma with "thing it represents (\sigma)"

examples/interfaces/stan.ipynb
```diff
@@ -0,0 +1,992 @@
...
+   "metadata": {},
+   "source": [
+    "## Eight Schools"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We now illustrate how to use the Stan interface to sample from the notorious eight schools model. First, we use the PINTS [eight schools toy model](https://github.com/pints-team/pints/blob/master/examples/toy/distribution-eight-schools.ipynb) to give us the requisite data."
```
MichaelClerx, line 400: Mention it's hierarchical? 

Optional: Give some idea of why it's notorious?

examples/interfaces/stan.ipynb
```diff
@@ -0,0 +1,992 @@
...
+   "source": [
+    "model = pints.toy.EightSchoolsLogPDF()\n",
+    "model.data()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Centered model"
```
MichaelClerx, line 430: What's a centered model?

examples/interfaces/stan.ipynb
```diff
@@ -0,0 +1,992 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Interfacing with Stan\n",
+    "This notebook explains how it is possible to interface with [Stan](https://mc-stan.org/) to make use of their awesome library for log-probability calculations, gradient calculation (via autodiff) and their large zoo of probability distributions.\n",
+    "\n",
+    "One thing to be mindful of is that the interface below only allows the log probability to be accessed from Stan objects up to an additive constant. Another is that Stan operates on the unconstrained scale, so the PINTS interface to Stan passes and receives parameter draws on this scale.\n",
+    "\n",
+    "In this notebook, we use two examples:\n",
```
MichaelClerx, line 12: This is good! But maybe reword in terms of what we're gonna do/show rather than what we're going to use?

pints/interfaces/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# Provides a number of toy models and logpdfs for tests of Pints' functions.
+#
+# This file is part of PINTS (https://github.com/pints-team/pints/) which is
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+
+from ._stan import StanLogPDF                                       # noqa
```
MichaelClerx, line 12: This would render the entire `interfaces` module useless if the `stan` interface doesn't work.

Some alternative options:

- Long form: `pints.interfaces.stan.StanThing`, then later `pints.interfaces.pymc3.OtherThing` etc.
- Short form, with some hint in name: `pints.interface_stan.StanThing` or `pints.stan_interface.StanThing`
- Simple form: `pints.stan.StanThing`

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,179 @@
...
+    def __call__(self, x):
+        if self._fit is None:
+            raise RuntimeError(
+                "No data supplied to create Stan model fit object. " +
+                "Run `update_data` first.")
+        vals = self._prepare_values(x)
+        try:
+            return self._log_prob(vals, adjust_transform=True)
+        # if Pints proposes a value outside of Stan's parameter bounds
+        except (RuntimeError, ValueError):
```
MichaelClerx, line 75: Should we have some option to run without this error checking? Will a user be able to debug their stan code if we catch the errors here?

MichaelClerx: Or raise a warning anytime this happens?

pints/interfaces/_stan.py
```diff
@@ -0,0 +1,179 @@
...
+    def _prepare_values(self, x):
+        """ Flattens lists from PyStan's dictionary. """
+        dict = self._dict_update(x)
+        vals = dict.values()
+        b = []
+        for ele in vals:
+            if not isinstance(ele, list):
+                ele = [ele]
+            b.append(ele)
+        vals = [item for sublist in b for item in sublist]
```
MichaelClerx, line 178: Is there a slightly more readable way to write this? I'm struggling a bit. But no worries if not :D

examples/README.md
```diff
@@ -132,10 +132,16 @@ relevant code.
 - [Simple Egg Box](./toy/distribution-simple-egg-box.ipynb)
 - [Twisted Gaussian Banana](./toy/distribution-twisted-gaussian.ipynb)
 
+
 ## Interfaces
+### To probabilistic programming languages
```
MichaelClerx, line 137: Do we need to make this distinction at this point?

pints/tests/test_interfaces_stan.py
```diff
@@ -0,0 +1,155 @@
...
+import numpy as np
+import pints.toy
+
+# Consistent unit testing in Python 2 and 3
+try:
+    unittest.TestCase.assertRaisesRegex
+except AttributeError:
+    unittest.TestCase.assertRaisesRegex = unittest.TestCase.assertRaisesRegexp
+
+try:
```
MichaelClerx, line 20: See above. If there's an "interfaces" module that isn't just for stan, then it should be possible to import it with or without stan

MichaelClerx (commented):
Great work! Few small comments on code

pints/interfaces/_stan.py
ben18785: I've added a warning to both times we use the try/except

pints/tests/test_interfaces_stan.py
ben18785: Ok, cool. See my comment above about whether that means I need to put the Stan stuff in a separate folder

pints/interfaces/_stan.py
MichaelClerx: Is it useful to add the pystan error message?
```
except (...) as e:
  warnings.warn('Text: ' + str(e))
```

pints/interfaces/__init__.py
MichaelClerx: If we go the `pints.interface.stan.StanThing` route the `pints.interfaces` module would be pretty much just a placeholder, so it would not import the stan module in its init.
Users would do e.g.
```
import pints.interfaces.stan as stan

print(stan.StanThing)
```

pints/interfaces/_stan.py
ben18785: Great idea! Done.

pints/interfaces/__init__.py
ben18785: good idea. Done

MichaelClerx (approved):
Let's merge it in!

ben18785:
@MichaelClerx yesssssssssss. Thanks!

PR #1079, opened 2020-03-20
Copyright blurb testing and Python 3.8

MichaelClerx:
Made it a separate thing, so it doesn't look like the unit tests are failing


MichaelClerx:
Seems travis hasn't been reporting tests properly. Manually looking at travis these tests finished days ago. @martinjrobins do you think we should switch over from travis.org to the newer travis.com? Had a few weird issues in other projects that were solved by that...

PR #1082, opened 2020-03-26
Updated ask-and-tell and transformed-parameters examples.

MichaelClerx:
They weren't easy to find (both were recently missed by people who know pints reasonably well), and were written as optimisation-specific, while they're actually more generally applicable than that.

Have tried to remedy


MichaelClerx:
Pssst! @martinjrobins @ben18785 please have a look!

PR #1087, opened 2020-04-01
I590 Add NormalisedRootMeanSquareError

### LINKED ISSUE #590: Add score function 'RRMSE'?
Sometime when I compare different output where the absolute magnitude does not really matter, I prefer to have a 'relative' RMSE score, which is simply `RMSE divided by RMSD of observed data to zero`. Maybe worth adding to PINTS?

chonlei:
Closes #590.

I have named it as 'NormalisedRootMeanSquareError', as it does seem like we are normalising the RMSE.


docs/source/error_measures.rst
```diff
@@ -23,6 +23,8 @@ Example::
 
 .. autoclass:: RootMeanSquaredError
 
+.. autoclass:: NormalisedRootMeanSquaredError
```
MichaelClerx, line 26: We're doing everything alphabetical now, so this needs to go up before ProbabilityBasedError !

pints/__init__.py
```diff
@@ -140,6 +140,7 @@ def version(formatted=False):
     SumOfErrors,
     MeanSquaredError,
     RootMeanSquaredError,
+    NormalisedRootMeanSquaredError,
```
MichaelClerx, line 143: Same here

pints/__init__.py
```diff
@@ -140,6 +140,7 @@ def version(formatted=False):
     SumOfErrors,
```
MichaelClerx, line 140: This one's also wrong :-D

pints/_error_measures.py
```diff
@@ -272,6 +272,35 @@ def __call__(self, x):
             (self._problem.evaluate(x) - self._values)**2))
 
 
+class NormalisedRootMeanSquaredError(ProblemErrorMeasure):
```
MichaelClerx, line 275: Again, alphabetical!

The reason is, when there's only a few classes you can make a sensible order, but then when there's more it becomes less clear and different people have different opinions so it's easier to just do everything alphabetical. Also makes it easier to find stuff :-)

pints/_error_measures.py
```diff
@@ -272,6 +272,35 @@ def __call__(self, x):
...
+    """
+    Calculates a normalised root mean squared error (RMSE):
```
MichaelClerx, line 277: This makes me think (normalised root mean square error) = (RMSE)
Maybe clearer to write "normalised RMSE (root mean square error)" in this case?

You might even want to break it up into multiple sentences, e.g.

```
"""
Calculates a normalised root mean square error.

This is similar to the (unnormalised) root mean square error (RMSE) defined as
``...``
but
...
```
?

MichaelClerx: Also would it be clearer to define it as ``sqrt(....) / N`` ? (even though in code it's better the way you have it)

pints/_error_measures.py
```diff
@@ -272,6 +272,35 @@ def __call__(self, x):
...
+    ``f = N * sqrt( sum( (x[i] - y[i])**2 / n) )``
+    where ``N`` is a normalisation constant defined as the inverse of the RMSE
+    of the data ``y`` with respect to zero.
```
MichaelClerx, line 280: Is the error-w.r.t.-zero way the best way to describe it? Is it conceptually an error w.r.t. a zero signal? Or do you derive the measure by thinking of sqrt(sum(y**2)) as some kind of magnitude?

pints/tests/test_error_measures.py
```diff
@@ -292,6 +292,25 @@ def test_root_mean_squared_error(self):
...
+        self.assertNotEqual(np.all(e([1, 2, 3])), 0)
+        x = [0, 0, 0]
+        y = 1.0
+        self.assertEqual(e(x), y)
+        x = [1, 1, 1]
+        y = np.sqrt((4 + 1 + 4) / 3) / np.sqrt((1 + 4 + 9) / 3)
+        self.assertAlmostEqual(e(x), y)
+
+        p = MultiMiniProblem()
+        self.assertRaises(ValueError, pints.NormalisedRootMeanSquaredError, p)
```
MichaelClerx, line 312: Can we do an assertRaisesRegex here?

docs/source/error_measures.rst
chonlei: :D OK! I thought about it, but then it didn't look alpahbetically ordered, so I put it to where I thought the most sensible place. Sure, I can move it!

pints/__init__.py
chonlei: But SumofErrors is before MeanSquaredError... Which bit is alphabetically ordered?

pints/__init__.py
chonlei: Oh! OK, you just answered my question above. I will move it too then.

chonlei: Is it just for the docs/__init__.py only? Or I need to move it for the actual _error_measures.py functions as well...?!

pints/_error_measures.py
chonlei: You kept answering my questions above! :-D

pints/_error_measures.py
chonlei: Do you have any suggestion on how to word it?

pints/__init__.py
MichaelClerx: They should have been!
You can fix all of them :D

pints/__init__.py
MichaelClerx: docs, imports, and actual code!

MichaelClerx: We've done it in _most_, but not all places, so you might see some other instances where it's wrong

pints/_error_measures.py
MichaelClerx: Maybe just write out the equation in full so you don't need to use too many words?

pints/_error_measures.py
chonlei: I have moved this one up too. But there're a few classes that others inherent, they can't go in alphabetical order. So I have left the others stay unchanged.

pints/_error_measures.py
chonlei: Done!

pints/tests/test_error_measures.py
chonlei: Yup -- changed it for the RootMeanSqauredError unit test too!

PR #1088, opened 2020-04-01
I1016 kde 1d hist

### LINKED ISSUE #1016: Add kernel density estimation to pints.plot.histogram
KDE is not available in the `pints.plot.histogram` function.

Some code which manually adds KDE to a histogram was previously written for the [`plot-mcmc-pairwise-kde-plots`](https://github.com/pints-team/pints/blob/master/examples/plot-mcmc-pairwise-kde-plots.ipynb) example notebook:

```
import numpy as np
import scipy.stats as stats

def plot_kde_1d(x, ax):
    """ Creates a 1d histogram and an estimate of the PDF using KDE. """
    xmin = np.min(x)
    xmax = np.max(x)
    x1 = np.linspace(xmin, xmax, 100)
    x2 = np.linspace(xmin, xmax, 50)
    kernel = stats.gaussian_kde(x)
    f = kernel(x1)
    hist = ax.hist(x, bins=x2, normed=True)
    ax.plot(x1, f)
```

chonlei:
Closes #1016 

Add a KDE option to the 1D histogram plot.


chonlei:
Also changed the histogram to plot the normalised/density histogram -- as it's done in the histogram plots in the pairwise plot.

chonlei:
If everyone is happy with that change, I will update the plots in the examples as well.

MichaelClerx:
> If everyone is happy with that change, I will update the plots in the examples as well.

Could you do 1 so we can see what it looks like (without trying it ourselves)?

MichaelClerx:
Actually, could you create a histogram example notebook? Maybe just sample from two priors and show the histogram for each, and then do KDE on one of them? I think we're missing that in https://github.com/pints-team/pints/tree/master/examples/plotting

PR #1089, opened 2020-04-01
I620 close figure objects in unit test

### LINKED ISSUE #620: Close figure objects in unit test
Currently in unit test for `pints.plot`, it creates a lot of figure objects and leave them there. Maybe do

```
pints.plot.blablabla()
plt.close('all')
```

But that we have to `import matplotlib.pyplot` in the unit test.

(Cannot just do `fig, _ = pints.plot.blablabla(); plt.close(fig)`, this will leave only remove the stuffs in the figure but not the figure itself.

chonlei:
Close #620.


pints/tests/test_plot.py
```diff
@@ -12,6 +12,7 @@
 import unittest
 import numpy as np
 import matplotlib
+import matplotlib.pyplot as plt
```
MichaelClerx, line 15: This needs to go after `matplotlib.use`

chonlei: OK! Yes, I got flake8 complained so I moved it before... Yes, I can do #noqa and move it later!

chonlei:
> `matplotlib.use` has to be called before `pyplot` is imported (might require #noqa to stop flake8 from complaining),but otherwise looks great!

Moved it and added #noqa. If it looks good, feel free to merge it.

MichaelClerx:
Thanks! I've gone 1 further and moved the pyplot imports into the test functions (I forgot that that was best practice when I reviewed earlier :man_facepalming: )

chonlei:
Ha! I started with that and not sure why I moved it out later. 😂

PR #1090, opened 2020-04-01
Update docs for logistic model

### LINKED ISSUE #1084: Typo in Logistic model doc
https://pints.readthedocs.io/en/latest/toy/logistic_model.html

It has the 'parameter' p0 in the equation which is the initial population size, but it's never been defined it. It did mention the initial population size can be set but then is notated as f(0).

chonlei:
Closes #1084.

### DIFF
--- a/pints/toy/_logistic_model.py
+++ b/pints/toy/_logistic_model.py
@@ -28,8 +28,8 @@ class LogisticModel(pints.ForwardModelS1, ToyModel):
                                          + \frac{1}{(k/p_0 - 1)\exp(-r t) + 1}
 
     Has two model parameters: A growth rate :math:`r` and a carrying capacity
-    :math:`k`. :math:`p_0` is the initial population size which is a fixed
-    (known) parameter in the model.
+    :math:`k`. The initial population size :math:`p_0` is a fixed (known)
+    parameter in the model.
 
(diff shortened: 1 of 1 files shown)


chonlei:
Also I think the initial population size can be float?! I can't see why it must be int in the model -- because this is a model for the expected population, not for individual count I think... And because if p_0 is int, then k (the carrying capacity of the population) must be int as well -- which is not what we do in the inference. But I am happy to be convinced that I was wrong.

pints/toy/_logistic_model.py
```diff
@@ -28,14 +28,15 @@ class LogisticModel(pints.ForwardModelS1, ToyModel):
...
+    parameter in the model.
 
     Extends :class:`pints.ForwardModel`, :class:`pints.toy.ToyModel`.
 
     Parameters
     ----------
-    initial_population_size : int
-        Sets the initial population size :math:`f(0)`.
+    initial_population_size : float
+        Sets the initial population size :math:`f(0) = p_0`.
```
MichaelClerx, line 39: Should this equation maybe go in the text above? And then just use either p_0 or f(0) here?

chonlei: Yup, sure.

Issue #734, opened 2019-02-20
Look at sanmitra's concerns about AR1 model


MichaelClerx:
https://github.com/pints-team/pints/issues/607#issuecomment-461435323

The docs should probably become a bit more clear about what and why we're doing, maybe with more literature references?

sanmitraghosh:
Also the likelihood calculation of any moving average type model ARMA is non-trivial, as the hidden noise series/path needs to be handled carefully. One approach is conditional likelihood (conditioned on the first sample of the path) or for unconditional likelihood a method called Backforecast. 
There are a plenty of good literature on time series and those should be followed/referred.

ben18785:
@sanmitraghosh We're using the conditional log likelihood here and so think all is ok. The generative model matches the inferential algorithm and so am fairly confident all is well.

sanmitraghosh:
@ben18785 Can you explicitly write somewhere what the model is, I mean what your assumption is?
For example are you saying that the residual R(t) =Y(t) - F(t,theta) follows an ARMA(1,1) process ?

Issue #676, opened 2019-02-08
AR1 and ARMA1 errors need a bit more documentation


MichaelClerx:
Seems both Simon and Sanmitra don't think the stated log likelihoods are obvious, so would be good to have a bit more of the derivation in the docstrings

MichaelClerx:
This could be a good one for @ben18785 or @rcw5890 to pick up?

rcw5890:
I'd be happy to give this a shot

MichaelClerx:
Thanks!

rcw5890:
Closed in #1093

PR #1093, opened 2020-04-03
676 ar arma docs

### LINKED ISSUE #694: ARMA11 documentation contains an unexplained... nu?
https://pints.readthedocs.io/en/latest/log_likelihoods.html?highlight=arma11#pints.ARMA11LogLikelihood

rcw5890:
Here's my current work on Small Issue #676. When this is finished, I think it should close #694 and #734 as well.

@ben18785  let me know if I've managed to get this correct and understandable. One of the main issues with these models seems to be the question of whether to work out the exact likelihood or condition on some initial observations for simplicity. In Pints it looks like the conditional likelihoods have been implemented, so I included that in the docstring. 

For both AR(1) and ARMA(1,1), we are currently using a transformed variance parameter for the white noise (e.g., sigma' = sigma sqrt{1-rho^2} for AR(1)). Is there a good justification or reference for these rescalings that you think we should mention in the docstring?

### DIFF
--- a/pints/_log_likelihoods.py
+++ b/pints/_log_likelihoods.py
@@ -109,7 +109,7 @@ class ARMA11LogLikelihood(pints.ProblemLogLikelihood):
     calculates a likelihood conditioned on initial values. This topic is
     discussed further in [2]_ , chapter 5.6. Thus, for a time series defined at
     points :math:`i=1,...,N`, summation begins at :math:`i=3`, and the
-    conditional likelihood is
+    conditional log-likelihood is
 
     .. math::
         L(\theta, \sigma, \rho, \phi|\boldsymbol{x}) =
--- a/pints/_log_likelihoods.py
+++ b/pints/_log_likelihoods.py
@@ -14,26 +14,41 @@ import scipy.special
 
 class AR1LogLikelihood(pints.ProblemLogLikelihood):
     r"""
-    Calculates a log-likelihood assuming AR1 noise model
+    Calculates a log-likelihood assuming AR(1) (autoregressive order 1)
+    errors. In this error model, the ith error term
+    :math:`\epsilon_i = x_i - f_i(\theta)` is assumed to obey the following
+    relationship.
 
     .. math::
-        \log{L(\theta, \sigma'|\boldsymbol{x})} =
-            -\frac{N}{2}\log{2\pi}
-            -N\log{\sigma'}
-            -\frac{1}{2\sigma'^2}
-                \sum_{i=2}^N{(\epsilon_i x_i - \rho \epsilon_{i-1} )^2}
+        \epsilon_i = \rho \epsilon_{i-1} + \nu_i
 
-    where
+    where :math:`\nu_i` is IID Gaussian white noise with variance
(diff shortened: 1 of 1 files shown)


pints/_log_likelihoods.py
```diff
@@ -14,26 +14,41 @@
 
 class AR1LogLikelihood(pints.ProblemLogLikelihood):
     r"""
-    Calculates a log-likelihood assuming AR1 noise model
+    Calculates a log-likelihood assuming AR(1) (autoregressive order 1)
+    errors. In this error model, the ith error term
```
MichaelClerx, line 18: Can you leave a blank line after the first sentence?
The general format should be:
```
One line explanation.

More stuff if needed.
```
The one-line explanation is used by IDEs etc. to give a quick idea of what the method does

pints/_log_likelihoods.py
```diff
@@ -14,26 +14,41 @@
...
+    :math:`\epsilon_i = x_i - f_i(\theta)` is assumed to obey the following
+    relationship.
 
     .. math::
-        \log{L(\theta, \sigma'|\boldsymbol{x})} =
```
MichaelClerx, line 20: Maybe leave this in, and say that the assumption gives rise to ...

pints/_log_likelihoods.py
```diff
@@ -73,29 +92,37 @@ def __call__(self, x):
 
 class ARMA11LogLikelihood(pints.ProblemLogLikelihood):
     r"""
-    Calculates a log-likelihood assuming ARMA(1,1) noise model.
+    Calculates a log-likelihood assuming ARMA(1,1) errors. The ARMA(1,1) model
```
MichaelClerx, line 95: See above

pints/_log_likelihoods.py
```diff
@@ -105,6 +132,11 @@ class ARMA11LogLikelihood(pints.ProblemLogLikelihood):
         A :class:`SingleOutputProblem` or :class:`MultiOutputProblem`. For a
         single-output problem three parameters are added (rho, phi, sigma),
         for a multi-output problem 3 * ``n_outputs`` parameters are added.
+
+    References
+    ----------
+    .. [2] Hamilton, James D. Time series analysis. Vol. 2. New Jersey:
```
MichaelClerx, line 138: I'm not 100% sure but I think this can be [1] too? So numbering per class rather than for the whole module? Correct me if I'm wrong please!

rcw5890: I also thought calling them both [1] would be better, especially as it is the same book. But Sphinx does throw a warning if you do that and messes up the HTML. 

``pints/_log_likelihoods.py:docstring of pints.ARMA11LogLikelihood:15: WARNING: Duplicate target name, cannot be used as a unique reference: "1".``

MichaelClerx:
@rcw5890 I'm assuming you've got the right permissions to hit the merge button?

rcw5890:
> @rcw5890 I'm assuming you've got the right permissions to hit the merge button?

Yes I do, but Ben just merged before I noticed this!

MichaelClerx:
Great! Just checking :-)

PR #1104, opened 2020-04-07
Simple gradient descent optimiser

### LINKED ISSUE #930: Add simple gradient descent optimiser
See also #893 

Only for models with sensitivities

MichaelClerx:
Closes #930 

- Updates the optimiser interface to allow sensitivities
- Updates the optimiser controller to work with sensitivities
- Adds a simple fixed learning rate gradient descent method


MichaelClerx:
@ben18785 & @alisterde could you have a quick look at this PR for me?
It adds in a first gradient based optimiser, which I think would pave the way for your BFGS optimiser.

MichaelClerx:
Thanks @alisterde !

pints/_optimisers/_gradient_descent.py
```diff
@@ -0,0 +1,126 @@
...
+        """ See :meth:`Optimiser.fbest()`. """
+        return self._fbest
+
+    def learning_rate(self):
+        """ Returns this optimiser's learning rate. """
+        return self._eta
+
+    def name(self):
+        """ See :meth:`Optimiser.name()`. """
+        return 'Fixed learning rate gradient descent'
```
ben18785, line 62: I would say just return 'Gradient descent' since this matches the class name...

MichaelClerx:
Thanks @ben18785 ! I quite like how the notebook suggests improvements to this method :D

MichaelClerx:
Have added some tests and changed the name now @ben18785 !

MichaelClerx:
> This all looks good to me code-wise.
> 
> Is this the canonical "gradient descent" optimiser or, for instance, are there examples with non-constant learning rate? I.e. is the name going to need to become more specific at some point in the future when someone implements a similar algorithm?

It's a good question, we'll probably want to name it something more specific in the future. (Gradient descent is one of those things that became a class of methods, I suppose)

But to be honest I'm not super worried about that, because this won't really be used in practice

MichaelClerx:
Then again, @fcooper8472 , the more specific ones do seem to have specific names, so could be fine to leave this as is even when we add more?

See https://github.com/pints-team/pints/issues/1105

Someone will probably do these for fun at some point :D

fcooper8472:
> Then again, @fcooper8472 , the more specific ones do seem to have specific names, so could be fine to leave this as is even when we add more?

Sounds good to me.

PR #1106, opened 2020-04-07
Updated examples to use plt instead of pl

### LINKED ISSUE #1103: Update examples to use plt instead of pl
Some old ones do

MichaelClerx:
Closes #1103


MichaelClerx:
@martinjrobins @chonlei this is a silly one! Anyone who feels like clicking approve please do :D

Issue #1072, opened 2020-03-19
Add optional parameter names to various Pints plots


ben18785:
This would make some of the plots:

- traceplots
- pairwise plots

More immediately presentable / publishable...

chonlei:
This was somewhat raised previously via adding it to the model #816, where things can be automated a bit more(?) and for saved samples CSV files as well? But yeah might be (a lot) easier to just do it via the plots as an optional input...

DavAug:
I am using the existing notebooks to test whether naming the parameters works as expected. Would you like me to include the changes to those notebooks into this issue as well?

ben18785:
Cool, thanks. No need to do it in all cases, but for a few notebooks would
be great.

DavAug:
PR #1118

DavAug:
I get alert that `assertRaiseRegexp` is going to be deprecated soon, and `assertRaiseRegex` should be used instead. I kept it for now to be in accordance with the rest of the tests.

Would you like me to replace it everywhere?

DavAug:
I also changed the `exclude: env` by `exclude: venv` in the .flake8 file, as .gitignore seems to ignore a virtual environment named venv and not env, but I can revert that if that was intentional.

MichaelClerx:
We usually have a block like this near the top

```
# Unit testing in Python 2 and 3
try:
    unittest.TestCase.assertRaisesRegex
except AttributeError:
    unittest.TestCase.assertRaisesRegex = unittest.TestCase.assertRaisesRegexp
```
and then use `assertRaisesRegex` exclusively

chonlei:
Actually, thought we were now going for a customising by user like: https://github.com/pints-team/pints/blob/master/examples/plotting/customise-pints-plots.ipynb instead?

MichaelClerx:
Both are valuable!

MichaelClerx:
Closed in #1126 #1118

MichaelClerx:
Thanks @DavAug !

PR #1118, opened 2020-04-30
Optional param names plots

DavAug:
See #1072


DavAug:
Ok I thought flake8 would fail because I changed the .flake8 file. That does not seem to be the problem. The fails come, however, from places I didn't fiddle around with. Any ideas?

MichaelClerx:
Not sure! We can have `env` and `venv` in .flake8 by the way. See what happens if you add them both in?

DavAug:
That does not seem to work.

MichaelClerx:
Oh wait, I just saw that the path is `./build/lib/pints/__init__.py`. I saw something similar with Myokit a few days ago.
Anyone understand what's going on here @martinjrobins @fcooper8472 ? Did pip change something?

MichaelClerx:
Yeah seems to be a new pip issue https://github.com/pints-team/pints/issues/1119

DavAug:
Ok! I think someone else might be better suited to tackle this problem.

.flake8
```diff
@@ -4,6 +4,7 @@ exclude=
   problems,
   pints/__init__.py,
   env
```
MichaelClerx, line 6: Missing a comma here after `env`

examples/plotting/mcmc-autocorrelation.ipynb
```diff
@@ -120,23 +117,29 @@
...
-      "text/plain": [
-       "<Figure size 432x432 with 3 Axes>"
-      ]
+      "text/plain": "<Figure size 432x432 with 3 Axes>",
```
MichaelClerx, line 122: There should be a figure here

pints/plot.py
```diff
@@ -212,6 +221,18 @@ def histogram(samples, kde=False, ref_parameters=None, n_percentiles=None):
     n_list = len(samples)
     _, n_param = samples[0].shape
 
+    # Check parameter names
+    if parameter_names is not None:
+        if len(parameter_names) != n_param:
+            raise ValueError(
+                'Length of `parameter_names` must be same as number of'
+                ' parameters.')
+        for name in parameter_names:
```
MichaelClerx, line 230: This second check can go: it can be whatever `set_xlabel` accepts

pints/plot.py
```diff
@@ -212,6 +221,18 @@ def histogram(samples, kde=False, ref_parameters=None, n_percentiles=None):
     n_list = len(samples)
     _, n_param = samples[0].shape
 
+    # Check parameter names
+    if parameter_names is not None:
```
MichaelClerx, line 225: I would do an `if parameter_names is None:` here that creates the default, and then `else` this code.
Then the second `if` inside the loop can go

pints/plot.py
```diff
@@ -315,6 +347,18 @@ def trace(samples, ref_parameters=None, n_percentiles=None):
                 'All samples must have the same number of parameters.'
             )
 
+    # Check parameter names
+    if parameter_names is not None:
```
MichaelClerx, line 351: See above

pints/plot.py
DavAug: I'm note sure I understand what you mean by loop here. Do you mean something like
```
if parameter_name is None:
    pass
elif len(parameter_names) != n_params:
   raise ....
```

DavAug: But I can revert the order in the plotting loops to 
```
if parameter_name is None:
   ... enumerate params ...
else:
   ... get names from parameter_names  ...
```

DavAug:
I don't have a good reason to put the `parameter_names` in the middle of the arguments, other than that it made sense to me to have the argument before before reference parameter values. But maybe it's just me.
I can move it to the end if that is preferred.

pints/plot.py
MichaelClerx: I meant getting rid of the `if` further down, when you're using the names. If you just create a `parameter_names` here you don't need to have an `if` in the plotting loop

MichaelClerx:
```
if parameter_names is None:
    parameter_names = <generate default names>
else:
    <check user input>

...

for i in ..
  for j in ...
     plot(..., parameter_names[0]
```

DavAug: I see, that makes sense!

MichaelClerx:
@DavAug can you add "build" to the list of paths that flake8 ignores? Then I can merge this

pints/plot.py
```diff
@@ -171,7 +171,12 @@ def function_between_points(f, point_1, point_2, padding=0.25, evaluations=20):
     return fig, axes
 
 
-def histogram(samples, kde=False, ref_parameters=None, n_percentiles=None):
+def histogram(
+        samples,
+        kde=False,
+        ref_parameters=None,
```
MichaelClerx, line 177: Maybe put `ref_parameters` at the end? If we're changing stuff anyway :sweat_smile: 

@chonlei @ben18785 @martinjrobins , what do you think the param order here (and in the other functions) should be?

MichaelClerx:
> @DavAug can you add "build" to the list of paths that flake8 ignores? Then I can merge this

Once we decide on the parameter order anyway :D

pints/plot.py
chonlei: What's wrong with `parameter_names` at the end?

MichaelClerx: Just trying to see if there's any particular 'best' order? Like it makes sense to put the two "parameters" arguments next to each other. And then maybe names will be used more often than ref_parameters ?

ben18785: Not helpful but I am completely ambivalent about this! Reckon merge it as, if we can't think of an obvious order, there likely isn't one.

MichaelClerx: :D OK!

In that case I'll just decide. @DavAug would you mind moving `ref_parameters` to the final position, after `parameter_names`, in the two cases we have them? After that I'm happy to merge

DavAug: Yep no problem.

MichaelClerx:
Ugh, now code-cove isn't working for this PR?

MichaelClerx:
> Run codecov
> 
>       _____          _
>      / ____|        | |
>     | |     ___   __| | ___  ___ _____   __
>     | |    / _ \ / _  |/ _ \/ __/ _ \ \ / /
>     | |___| (_) | (_| |  __/ (_| (_) \ V /
>      \_____\___/ \____|\___|\___\___/ \_/
>                                     v2.0.22
> 
> ==> Detecting CI provider
>   -> Got branch from git/hg
>   -> Got sha from git/hg
> ==> Preparing upload
> Error: Missing repository upload token
> 

@fcooper8472 do you think this was a random error, or is it to do with this being on a fork?

MichaelClerx:
Hmmm. Seems this _should_ work https://github.com/codecov/codecov-action/issues/29

MichaelClerx:
Pfffft. Just pulling this into a local branch instead, let's see if that works :D

Issue #1120, opened 2020-05-04
comment in ToyODEmodel


alisterde:
I looked at the code for the ToyODEmodel class when I was doing sensitivities and noticed a small mistake in the comments, I'm fixing it to get get used to how issues and GitHub process work

ben18785:
Cool -- feel free to have a go @alisterde -- if you add me on the pull request once you've done it, I can add my review.

MichaelClerx:
Thanks @alisterde !

To get comments from other people, I think it's easiest to just mention them in the post. Then we can save the "assignee" feature to point to the person actually doing the implementing. (So if you're reporting an issue you don't intend to fix yourself, leave it unassigned. Only when you actually start coding you should assign yourself)

alisterde:
Thanks @MichaelClerx, I'll be sure to do that next time!

Issue #1119, opened 2020-04-30
Style tests fail due to `pip` leaving behind a `build` directory


MichaelClerx:
This has changed in the last 2 days

https://github.com/pypa/pip/issues/8165

PR #1121, opened 2020-05-04
commenting ToyODEModel class

alisterde:
@ben18785 and @MichaelClerx I've added what I wanted to for issue #1120, and pre-commit checks have all passed. All I've done is add a few comments and change the thing I spotted.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,7 +3,8 @@ exclude=
   .git,
   problems,
   pints/__init__.py,
-  env
+  env,
+  build
 ignore=
   # Block comment should start with '# '
   # Not if it's a commented out line
--- a/pints/toy/_toy_classes.py
+++ b/pints/toy/_toy_classes.py
@@ -96,7 +96,7 @@ class ToyODEModel(ToyModel):
 
         Returns
         -------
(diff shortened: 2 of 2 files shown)


MichaelClerx:
Hi @alisterde !

That was a bit quick maybe, so I'll just write it here again :D

We use "continuous intregration" (GitHub actions) to run all our tests (see`https://github.com/pints-team/pints/blob/master/run-tests.py) after every commit to master, and on all branches with open PRs.
Part of this is running flake8 (see https://github.com/pints-team/pints/blob/master/CONTRIBUTING.md for details) to test the style
But the first step in running any test is checking out and installing PINTS with `pip` (again, see CONTRIBUTING.md)
A very recent change to pip is that it now leaves behind a directory called `build` (https://github.com/pints-team/pints/issues/1119) and this directory was also being scanned by flake8, causing the tests to fail in continuous integration
Your PR and @DavAug 's one were the first to run into this issue.
What I did to fix it was to adapt the configuration file for flake8 (`.flake8`) to add `build` to the list of directories for flake8 to ignore

Issue #606, opened 2019-01-23
Add tests for different input shapes for every likelihood / error measure


MichaelClerx:
They should all be tested (and work for)
- Single-value data of shape ``(n_times,)`` (or just a list)
- A NumPy array of shape ``(n_times,1)``
- A NumPy array of shape ``(n_times, n_outputs)`` where n_outputs > 1

See #604, #605

MichaelClerx:
Might also be worth discussing if things'd be easier we had ``(n_outputs, n_times)`` and ``(n_times)`` as the required formats

DavAug:
Before I start testing I just want to clarify how the test may be best structured.

In test_error_measures.py are some error functions tested for single and for multi-output data and those are organized in separate tests. This can be straightforwardly done for all error functions and likelihoods. As far as I can see is the data for the single output problem here always `(n_times,)` or a list.

Would it be best to write yet another test which is structured exactly the same but passes data as `(n_times, 1)`? Or can this be done more elegantly?

MichaelClerx:
I would write something in one test and then just copy-paste it into the other tests, bit repetitive but makes it easier to debug and add exceptions and such?

DavAug:
Ok, so I write for example a `test_input_shape_mean_squared_error` and then I can check with `self.assertRaisesRegexp` that certain errors are thrown for wrong input shapes. But what's the best way to check that no error is thrown? Something like 
```
try:
    pints.MeanSquaredError(p)
except:
    raise ValueError(...)
```
or is it sufficient to just test the exceptions, because the functionality for lists and multi-output problems is tested in other tests.

MichaelClerx:
Yup, test the exceptions as above! But also do a quick check to see if the other tests cover all three cases. 

It might even be worth it to bring all these tests in some kind of standardised format? But not a requirement for this ticket.

DavAug:
Ok, thanks.

DavAug:
The way `pints.vector` works, any input of shape `(1,1,1,...,1, n_times, 1, ..., 1)` will also be accepted by the single output problem. Is that ok, or should there be any exceptions?

MichaelClerx:
Yeah that's fine. It's to get rid of differences like (n, 1) vs (n, ) and even (1, n) and your example :-)

DavAug:
As I am writing the tests, I'm starting to wonder whether it really makes sense to test the various input shape exceptions on the error measure level. 
[...]

PR #1123, opened 2020-05-05
I1122 fix constant model simulate s1

DavAug:


DavAug:
Not surprisingly this PR breaks tests of the error measures and likelihoods. But those are worked on in #606 and need this PR. What should we do?

DavAug:
See #606 .
Changes are need for this PR, so they will merged into master through this issue.

PR #1125, opened 2020-05-05
I606 test shapes likelihoods error measures

DavAug:
See issue #606

PR #1131, opened 2020-05-11
I606 test shapes likelihoods error measures

### LINKED ISSUE #606: Add tests for different input shapes for every likelihood / error measure

DavAug:
Closes #606


DavAug:
There has been an issue with the forked repo. See PR #1125 for the details.

DavAug:
I apparently have missed 3 lines, but codecov doesn't give me access to find out which lines. Do you know how to fix this?

DavAug:
I refactored the unit tests for likelihoods and error measures and tried to come up with a somewhat standardised testing scheme.

MichaelClerx:
Some of the test functions are quite long. Would it be easy to split them into smaller tests, or is it too much hassle creating the test model/data/problem?

DavAug:
Hmm I'm sure it's possible to split them up :D
Would you be able to just show case what you have in mind for one test, say `test_gaussian_log_likelihood_single()` for example, and then I could try to generalize this for the other tests

pints/tests/test_error_measures.py
```diff
@@ -108,25 +115,52 @@ def test_mean_squared_error_single(self):
...
+        self.assertEqual(y, e([x]))
+        self.assertEqual(dy.shape, (1,))
+        self.assertEqual(dy, 2 * (x - 1))
+        x = 3
+        y, dy = e.evaluateS1([x])
+        self.assertEqual(y, e([x]))
+        self.assertEqual(dy.shape, (1,))
+        self.assertEqual(dy, 2 * (x - 1))
+
+        # Test Case II: Input as array of shape (n_times, 1)
```
MichaelClerx, line 142: "Test Case II" is a good example of where it looks like you could split into two methods

DavAug: Ok. now that I think about it. Does it maybe make sense to split the sensitivity checks? I think this would make the tests much smaller.

MichaelClerx: Sure!

I was also thinking about trying out [this](https://github.com/pints-team/pints/issues/1172) at some point...

DavAug: Ok I'll give it a go for one likelihood and then let you have a look

DavAug: @Michael 
Whenever you have time, I refactored the ScaledLogLikelihood test. I personally think it's a lot tidier now than before, but let me know what you think. We could try to implement the other tests in this pattern too.

MichaelClerx: That looks great!

DavAug:
@MichaelClerx 
I finished to refactor the tests as we discussed. Let me know what you think.

I didn't refactor the SumOfErrors test because there are many cases, I wouldn't necessarily test for. Would you like to simplify this test to be more similar to the other tests, or rather keep the structure of the test but split it up into smaller tests?

MichaelClerx:
Looks great. Thanks!

PR #1138, opened 2020-05-15
I1137 horowitz langevin mcmc

DavAug:
The algorithm is largely the same as the existing a `pints.HamiltonianMCMC`. I made references in the ask and tell methods where the code has been adapted.

The other change was to remove the hyper parameter leapfrog_steps and introduce a new one alpha that controls the persistence of the momentum.


DavAug:
@MichaelClerx @chonlei @ben18785 
I ran the doc builds but no documentation is generated. Any ideas?

DavAug:
To clarify: 
I tried 
`$ python run-tests.py --quick`
and also 
`$ cd docs` and then `$ make clean; make html`

chonlei:
I just tried it, it seems working for me (both the master and this branch). I am using v2.4.4 for Sphinx. Have you got Sphinx installed properly? Can you try to reinstall pints with `pip install -e .[dev,docs]`?

chonlei:
Just thinking given their similarity, would it be worth to have this HorowitzLangevinMCMC inherits from the HamiltonianMCMC class...?

DavAug:
Thanks I’ll try that.

I think that would be a really good idea if all methods from HMC would be in the HorowitzLangevin sampler. But since the leapfrog steps have been removed does it still make sense?

I guess what could be done is rewriting MALA into a more obvious form of HMC with one leapfrog step and then let HMC inherit from Langevin and also Horowitz from Langevin. But I’m not sure whether those changes are desirable right now?

On 15 May 2020, at 18:34, Chon Lok Lei <notifications@github.com> wrote:

﻿

Just thinking given their similarity, would it be worth to have this HorowitzLangevinMCMC inherits from the HamiltonianMCMC class...?

DavAug:
This sampler is not the Neal Langevin Sampler yet. The updates of u are not included.
I was under the impression that we wanted to include both MCMC samplers? Horowitz-Langevin and Neal Langevin. Even though Neal reduces to Horowitz when we turn of the u updates and Horowitz reduces to Langevin when turning of the persistence of the momentum, does it not still make sense to have them as separate methods?

Get Outlook for iOS<https://aka.ms/o0ukef>

ben18785:
Sorry, I was not clear when we spoke this morning! (Was still trying to understand how the sampler worked and the best way to implement it.)

If we include Horrowitz and the Neal sampler as separate objects, that means we’ve got a load of duplicate code that all needs to be maintained.

Also, I think that Horrowitz probably isn’t worth having on its own: it’s very similar to MALA and this sort of persistent momentum sampling isn’t thought to help much. 

The Neal sampler may do though. Having a hyperparameter that switches on/off reversibility allows the best of both worlds, I reckon.

DavAug:
This is the completed NealLangevin Sampler. 

1) There does seem to be an issue with Sphinx that I can't work out. I downgraded the latest version of Sphinx to 2.4.4 to match with @chonlei, but it doesn't work. Sphinx does seem to run and it checks for the existing documentation, but there is no new documentation for the Neal Sampler generated with `make clean, make html`.

2) I chose a Gaussian noise for the updating of `u`, which can be tuned. The noise is not specified in the paper. Do we want a less stringent solution for the noise? E.g. one could pass in a noise generating function or something?

3) For u, delta and the noise of delta, no default initial values are provided. So I chose some and commented on it. Maybe those have to be adapted.

4) It is not clear to me whether `u` should be updated for divergent transitions. At the moment it is not updated, but I tend to say that it should be updated, since in HMC we also sample a new momentum when a divergent transition occurs. I wasn't sure about the behaviour of the sampler in those cases, that's why I wanted to discuss this here.

chonlei:
@DavAug Besides the documentation that you have included in the Python class, you also have to include and provide a link to the docs in order to let the Sphinx knows what to link to where!

chonlei:
@DavAug I have added those to the docs for you, so that you don't have to find where to add this time. Remember to pull and merge before doing any updates/push.

Also if you now run the make, you will see some warming to fix for the documentation that you have in the MCMC class. Can you please fix those?

And you might want to change the title to "Neal Langenvin MCMC" instead of "Horowitz Langenvin MCMC" too!

DavAug:
@chonlei Great thanks! It makes sense, but I didn't know that you had to five further instruction for Sphinx :D

DavAug:
@ben18785 Yes, I can make an example notebook.

DavAug:
I somehow don't seem to have the permission to look at codecov's analysis and see which lines where not hit. Is there something we can do about it?

MichaelClerx:
> I somehow don't seem to have the permission to look at codecov's analysis and see which lines where not hit. Is there something we can do about it?

Even after you hit "Login with github" ?

DavAug:
Yes, a Github API error is thrown, but it's apparently a Safari issue. It works with Chrome.

MichaelClerx:
@ben18785 @DavAug this is 99% done :D

PR #1139, opened 2020-05-15
Fixed i1136 in EightSchoolsLogPDF

simonmarchant:
Added a term dL1[0] += (theta - mu) / tau**2 in evaluateS1 of EightSchoolsLogPDF, and appropriate tests.


pints/tests/test_toy_eight_schools_logpdf.py
```diff
@@ -31,6 +31,15 @@ def test_basic(self):
         val = f([1, 0.5, 0.4, 1, 1, 1, 1, 1, 1, 1])
         self.assertAlmostEqual(val, -38.24061255483484, places=6)
 
+        # Default settings with non-ones input
+        f1, dp = f.evaluateS1(list(n + 1 for n in range(10)))
```
MichaelClerx, line 35: usually you'd just write `evaluateS1([n + 1 for ...])`

MichaelClerx:
While we're at it: the math block for this class doesn't render properly:

https://pints.readthedocs.io/en/latest/toy/eight_schools_logpdf.html

pints/tests/test_toy_eight_schools_logpdf.py
simonmarchant: Fixed in commit f866c8f

simonmarchant:
@ben18785 I've made the changes Michael identified here. The math block should now render correctly but I'm not sure how to check it properly renders in sphinx/readthedocs.

MichaelClerx:
Thanks!

MichaelClerx:
> The math block should now render correctly but I'm not sure how to check it properly renders in sphinx/readthedocs.

See https://github.com/pints-team/pints/blob/master/CONTRIBUTING.md#building-the-documentation

simonmarchant:
> Looks really good - thanks @simonmarchant! Can I just check -- did you test out how whether centered model sampling (in the example notebook) looks any different now? I don't suspect it would look significantly different.

You would be right, it looks very similar! I've pushed with an updated ipynb; the centred output looks like this (new version on left). One of the threads gets less stuck than before, otherwise it looks similar to me.
image

ben18785:
@simonmarchant thanks -- looks great!

Issue #416, opened 2018-07-27
Explain noise bias in 'first optimisation' notebook example


chonlei:
In this [example](https://github.com/pints-team/pints/blob/e390527fb3ade9171641c7338d68109f12c296a4/examples/optimisation-first-example.ipynb) `In [14]`, we showed the found parameters are at the minimum of the score function but the actual true values are not, but we didn't explain what it means.

@mirams suggests 
> we could have a little subheading that says "Why aren't these exactly the parameters used to generate the data?" and then show as noise reduces that they tend to right answer. Although the 1D line graph will need axes with units fixed or it will just look the same!

MichaelClerx:
Would be cool to combine this with an explanation of solver tolerance (which can create local min max, but also shift the optimum, as in gary's graphs)

MichaelClerx:
@chonlei do you want to have a look at this?

MichaelClerx:
Fixed in #1141

PR #1141, opened 2020-05-19
Added tiny noise bias section to "optimisation: first example"

MichaelClerx:
See #416 

Also started to do a bit of exploration trying to graph the error as a function of sqrt(n), but that involves lots of distractions like:

- scaling parameters
- removing outliers (because optimisers don't always work)
- running for ever and ever

so got rid of that part again!


mirams:
Can someone give me a button to click to look at the notebook, can't find what I should be looking at (not good at pull requests!)

chonlei:
@mirams Here: https://github.com/pints-team/pints/blob/4e624ba3a487c6da65b4d019e5b83e370e9a085e/examples/optimisation/first-example.ipynb (from the last commit), though you can't do in-line review (not that I am aware of!)...

MichaelClerx:
> Can someone give me a button to click to look at the notebook, can't find what I should be looking at (not good at pull requests!)

If you hit "Files changed" near the top, then there's three dots next to each filename giving you some more options, including "View file"

mirams:
I like it. Maybe first show the answer you get from optimisation with no noise on the synthetic data (probably good practice anyway), then introduce it the other way and say what made the new optimisation answer a bit wrong... obviously it was the noise we added, let's look at that (I like the evaluation of objective and 1d slice).

I'd avoid too reading too much into the 'mean of the noise' and linking this directly to SEM - this objective really works with the square of the noise, and SEM also does too (SEM=sqrt(E(X^2) - E(X)^2)/sqrt(N) and all that). It's not strictly correct to say the offset calculated at the bottom is the same.

Nice to get across the idea that this is related to the sigma on the noise you added, I agree the full exercise got a bit confusing. Maybe just end with a statement that's the case and a link to someone doing a tutorial showing it's proportional to 1/sqrt(N)?

MichaelClerx:
Thanks @mirams !
I've updated the example to start from a noise-free case, and removed the stuff about the standard error of the mean. Instead I'm just linking to the wiki article on overfitting now, and we just end the (PINTS) example there

https://github.com/pints-team/pints/blob/d7cf2f55366d05fa2b16ecc26cc40d7a48c07cd1/examples/optimisation/first-example.ipynb

If you like it, please use the approve button somewhere top right :D

Issue #400, opened 2018-07-23
Loss surface visualisation


MichaelClerx:
Hi all!

@sanmitraghosh @chonlei and I have been playing with loss surface visualisation lately.

I have some code from Myokit that does similar things in 2d. One method using a 3d plot (requires a special package), and another that uses voronoi diagrams for the plot - since points are usually not equally spaced.

https://github.com/MichaelClerx/myokit/blob/master/myokit/lib/fit.py#L482-L587
https://github.com/MichaelClerx/myokit/blob/master/myokit/lib/fit.py#L1967-L2179

Should we include this in Pints maybe? @martinjrobins ?

Maybe not the 3d one, but the voronoi stuff could be good for examples?

martinjrobins:
Yea, the voronoi one looks useful for the examples and demonstrating the optimisers/samplers. Is there anyway of making it usable for different number of dimensions? Like a plane-cut through a 3d data set? (I'm guessing not, but thought I'd ask...)

MichaelClerx:
Hmmm. I think in principle you _could_ do a multi-dimensional voronoi, and then cut a plane through. Since you know all the line equations you could calculate this. Problem is, that'd mean digging into the voronoi code instead of using scipy's!

martinjrobins:
naaa, I was hoping you could translate the problem into 2d before the voronoi. I've heard bad things about voronoi in dim > 2

PR #1142, opened 2020-05-19
2d loss surface visualisation

MichaelClerx:
See #400 

Added a voronoi based method to show errors defined on 2d surfaces with irregularly spaced points.

The notebook hopefully shows why this is useful.

Adding this much plotting code, it made sense to split `plot.py` into multiple files, so have done that here too


MichaelClerx:
This one @ben18785

MichaelClerx:
> The example is lovely! I like it! Is it just optical illusion?! Any idea why PSO would (seem to) have a lot of steps that go like a cross in the middle?!

I'm not quite sure. It doesn't always happen. There might be something in the way it's written that favours that, or it could be part of the problem surface... Maybe I'll look into that

MichaelClerx:
> > The example is lovely! I like it! Is it just optical illusion?! Any idea why PSO would (seem to) have a lot of steps that go like a cross in the middle?!
> 
> I'm not quite sure. It doesn't always happen. There might be something in the way it's written that favours that, or it could be part of the problem surface... Maybe I'll look into that

I'll make a ticket for it :D

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
+#
+# Plots a function defined on a two-dimensional space.
+#
+# This file is part of PINTS (https://github.com/pints-team/pints/) which is
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+# This module was was adapted from Myokit (see http://myokit.org) by the
```
DavAug, line 8: typo "... was was..."

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+
+    # Extract x and y points
+    x = points[:, 0]
+    y = points[:, 1]
+    del(points)
+
+    # Check boundaries
+    if boundaries is None:
+        xmin = 0.80 * np.min(x)
+        xmax = 1.02 * np.max(x)
```
DavAug, line 65: Would it not make more sense to choose boundaries relative to the difference of `np.min(x)` and `np.max(x)`.  Same for `y`.  That way we get symmetric boundaries and not some funny skewed boundaries for min << max.

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    # will bring us outside the visible area for any starting point / direction
+    #
+    center = vor.points.mean(axis=0)
+    radius2 = 2 * np.sqrt((xmax - xmin)**2 + (ymax - ymin)**2)
+
+    # Create a list containing the set of vertices defining each region
+    regions = [set() for i in range(n)]
+    for (p1, p2), (v1, v2) in izip(vor.ridge_points, vor.ridge_vertices):
+        # Ensure only v1 can every be -1
+        v1, v2 = (v2, v1) if v1 > v2 else (v1, v2)
```
DavAug, line 177: what does the comment have to do with v1 < v2?

DavAug: it's probably just me not knowing how voronoi works

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+        axes.plot(x, y, markers)
+
+    # Add colorbar
+    figure.colorbar(c, ax=axes)
+
+    return figure, axes
+
+
+def _voronoi_regions(x, y, f, xlim, ylim):
+    """
```
DavAug, line 102: Is there a reference that could be put here that explains in more detail how the voronoi regions are constructed?

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    voronoi diagram.
+
+    Returns a ``matplotlib`` figure object and axes handle.
+
+    Parameters
+    ----------
+    points
+        A list of (two-dimensional) points in parameter space.
+    values
+        The values corresponding to these points.
```
chonlei, line 35: 'values' can be a misleading word in Pints, aren't we used to call the data to be fitted to as 'values'.

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    boundaries
+        An optional :class:`pints.RectangularBoundaries` object to restrict the
+        area shown.
```
chonlei, line 38: It's an optional argument, maybe explain what'd happen if set to ``None``?

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    markers
+        The markers to use to plot the sampled points. Set to ``None`` to
+        disable.
+
```
chonlei, line 42: Delete extra line? 😄

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    # Check points and values
+    points = pints.matrix2d(points)
+    n, d = points.shape
+    if d != 2:
+        raise ValueError('Only two-dimensional parameters are supported.')
+    values = pints.vector(values)
+    if len(values) != n:
+        raise ValueError(
+            'The number of values must match the number of points.')
```
chonlei, line 55: If 'value' is changed, change here too.

pints/plot/_surface.py
chonlei: Yeah maybe something like
```
xrange = np.max(x) - np.min(x)
xmin = np.min(x) - 0.1 * xrange
xmax = np.max(x) + 0.1 * xrange
```

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    # Check boundaries
+    if boundaries is None:
+        xmin = 0.80 * np.min(x)
+        xmax = 1.02 * np.max(x)
+        ymin = 0.80 * np.min(y)
+        ymax = 1.02 * np.max(y)
+    else:
+        if boundaries.n_parameters() != 2:
+            raise ValueError(
+                'If given, the boundaries must be two-dimensional.')
```
chonlei, line 71: 'The boundaries must be two-dimensional if given.'?

chonlei: And maybe move this check earlier to line 56 or something?

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    ----------
+    points
+        A list of (two-dimensional) points in parameter space.
+    values
+        The values corresponding to these points.
+    boundaries
+        An optional :class:`pints.RectangularBoundaries` object to restrict the
+        area shown.
+    markers
+        The markers to use to plot the sampled points. Set to ``None`` to
```
chonlei, line 40: Maybe say 'matplotlib' markers? I wouldn't know what possible markers would be available.

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+    axes.set_ylim(ymin, ymax)
+
+    # Add coloured voronoi regions
+    c = PolyCollection(
+        regions, array=values, edgecolors='none', cmap='viridis_r')
+    axes.add_collection(c)
+
+    # Add markers
+    if markers:
+        axes.plot(x, y, markers)
```
chonlei, line 93: Specify the colour as well -- so that it matches (even if different version...?) the viridis_r cmap set above?

pints/plot/_surface.py
chonlei: Wiki is quite nice: https://en.wikipedia.org/wiki/Voronoi_diagram

pints/plot/_surface.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Points and voronoi regions entirely outside the specified bounds will be
+    dropped. Voronoi regions partially outside the bounds will be truncated.
+    The third array ``f`` will be filtered the same way as ``x`` and ``y`` but
+    is otherwise not used.
+
+    Parameters
+    ----------
+    x
+        A list of x-coorindates
```
chonlei, line 115: Missing full stop everywhere 😄

pints/plot/_surface.py
chonlei: It's kind of explained in line 161-165. v1, v2 are non-negative integers except the inf one being -1, so that if v1 > v2 then swap them makes sure -1 must go first.

Obviously I agree that the line does more than what the comment wants!

pints/toy/__init__.py
```diff
@@ -9,30 +9,30 @@
 from __future__ import absolute_import, division
 from __future__ import print_function, unicode_literals
 
-from ._toy_classes import ToyLogPDF, ToyModel, ToyODEModel          # noqa
```
chonlei, line 12: Just curious, why are these now removed the 'noqa'? (Or why was 'noqa' added before?)

pints/plot/_surface.py
MichaelClerx: What else would you call them? They are the values that go with some points. Could be errors or likelihoods. Don't have to :man_shrugging: 

I'm quite happy for pints methods to have an argument `value` of `values` where the value isn't a measured data point.

MichaelClerx: I've updated the docstring to make this more clear

pints/plot/_surface.py
MichaelClerx: OK I've removed the 'given' thing as we've already checked they're given at this point.
Not moving the check as then I have to say if boundaries is not None and bound...

pints/toy/__init__.py
MichaelClerx: Because every single line in the file had a #noqa on it, I just removed this file from the style checking (like we already did with another similar page)

MichaelClerx:
Thanks both! Have made several fixes and improvements!

pints/plot/_surface.py
chonlei: Much better! Thanks! Yeah, I guess just explaining it in the doc makes it clear.

pints/plot/_surface.py
```diff
@@ -61,14 +67,15 @@ def surface(points, values, boundaries=None, markers='+', figsize=None):
 
     # Check boundaries
     if boundaries is None:
-        xmin = 0.80 * np.min(x)
-        xmax = 1.02 * np.max(x)
-        ymin = 0.80 * np.min(y)
-        ymax = 1.02 * np.max(y)
+        xmin, xmax = np.min(x), np.max(x)
+        xmin -= 0.1 * (xmax - xmin)
+        xmax += 0.1 * (xmax - xmin)
```
chonlei, line 72: This will be the shifted xmin... I would prefer `(xmax - xmin)` being a fixed value (much easier to interprete)...

MichaelClerx: Thanks!

PR #1143, opened 2020-05-19
Fixed PSO issue where points returned by ask were later modified.

MichaelClerx:
With this fix you can store the points that were passed back from ask(), and they won't later be modified


MichaelClerx:
@martinjrobins could you have a look at this one? Think this is OK performance wise

Issue #1114, opened 2020-04-24
Table of content in each (sub)page of the docs


chonlei:
As we expand Pints' functions (which is good!), it becomes harder to navigate the docs. For example, here: https://pints.readthedocs.io/en/latest/log_priors.html, it's really difficult to see what's available and what's not. So I think it'd be nice to have something like a table of content on this kind of subpages!

chonlei:
For example, I wouldn't know there's a LogNormalLogPrior in the middle of the page: https://pints.readthedocs.io/en/latest/log_priors.html#pints.LogNormalLogPrior. Unless there's some smarter way of navigating the docs page that I don't know...

PR #1147, opened 2020-05-20
I1114 table of content subpage

DavAug:
See #1114.

A table of contents exists now on every subpage in the dos, as well as a more detailed list in the side bar. Here and there the docstrings were improved by teXing the fomulas.

### DIFF
--- a/docs/source/boundaries.rst
+++ b/docs/source/boundaries.rst
@@ -16,17 +16,8 @@ Overview:
 - :class:`RectangularBoundaries`
 
 
-Customisable boundaries
-***********************
-
 .. autoclass:: Boundaries
 
-Boundaries based on log-PDF
-***************************
-
 .. autoclass:: LogPDFBoundaries
-
-Rectangular boundaries
-**********************
 
 .. autoclass:: RectangularBoundaries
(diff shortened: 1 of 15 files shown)


docs/source/boundaries.rst
```diff
@@ -9,8 +9,24 @@ Simple boundaries for an optimisation can be created using
...
 implementation of the :class:`Boundaries` interface.
 
+Overview:
+
+- :class:`Boundaries`
+- :class:`LogPDFBoundaries`
+- :class:`RectangularBoundaries`
+
+
+Customisable boundaries
```
MichaelClerx, line 19: I don't think we should have a sub-header for each class. If the distinction needs to be clearer we could probably modify the CSS instead?

chonlei: Agree with Michael for the sub-header for each class -- looks a bit confusing in fact...

chonlei: Except if we do this for _every_ autoclass, than we could generate the toctree for all without manually typing in in the summary...?

MichaelClerx:
Maybe this works? https://stackoverflow.com/questions/24129481/how-to-include-a-local-table-of-contents-into-sphinx-doc

chonlei:
> Maybe this works? https://stackoverflow.com/questions/24129481/how-to-include-a-local-table-of-contents-into-sphinx-doc

Unfortunately I think(!) this works only for sub-headers... :-\

PR #1153, opened 2020-05-28
Better tests for parallelisation.

MichaelClerx:
Investigating some issues with multiprocessing: the issues I'm seeing may not be PINTS related, but did notice these tests could be more stringent


MichaelClerx:
@martinjrobins or @fcooper8472 could you have a quick look? No changes other than explicitly making sure we test with multiple processes when parallelising

MichaelClerx:
Nah just increases the chance that they interfere with each other, or are slow to load or any of that :-)

MichaelClerx:
Thanks @fcooper8472 !

PR #1156, opened 2020-06-02
improved doc rhat

### LINKED ISSUE #1144: Improve docstrings of rhat and rhat_all_params
Currently the docstrings are as follows:

- pints.rhat(samples)
Calculates 𝑅̂ =𝑠𝑞𝑟𝑡(((𝑛−1)/𝑛∗𝑊+(1/𝑛)∗𝐵)/𝑊) as per [1] for a single parameter. It does this after splitting each chain into two.

- pints.rhat_all_params(chains)
Calculates 𝑅̂ =𝑠𝑞𝑟𝑡(((𝑛−1)/𝑛∗𝑊+(1/𝑛)∗𝐵)/𝑊) as per [1] for all parameters. It does this after splitting each chain into two.

Maybe it is expected to know what W and B are in those cases, but would it not be nicer to at least say that W is the within chain variance and B the between chain variance, and maybe even define those quantities more explicitly?

### LINKED ISSUE #1157: Input checks for rhat computation
There is currently nothing that checks the input to the rhat diagnostics

### LINKED ISSUE #1158: rhat computation warm up and split into sub chains
Currently Rhat is not computed according to the provided reference, where it is suggested to ignore the first half of each chain to reduce the bias of the initial conditions. Then the truncated chains are split into two so the stationarity of each chain can be assessed. 

Neither of those steps are currently performed in the pints.rhat method. Is this on purpose, so these manipulation are expected to be done by the user, or should we integrate them into the rhat computation.

I personally was confused that these operations were not automatically performed and other users might be too.

DavAug:
Closes #1144 
Closes #1158 
Closes #1157 

I extended the doc string for Rhat according to the reference. I hope it makes it clearer what happens.


pints/_diagnostics.py
```diff
@@ -100,15 +100,56 @@ def reorder_all_params(chains):
...
-    """
-    W = within(samples)
-    B = between(samples)
-    t = len(samples[0])
+    Returns the convergence measure :math:`\hat{R}` according to [1]_.
+
+    :math:`\hat{R}` diagnoses convergence by checking mixing and stationarity
+    of :math:`M` chains (at least two, :math:`M\geq 2`). To diminish the
+    influence of starting values, the first half of each chain is considered
+    as warm up and does not enter the calculation. Subsequently, the truncated
```
martinjrobins, line 110: looking at the code, it looks like it doesn't discard warm-up. Did I miss something?

DavAug:
That is correct. It also doesn’t split the chains into 2. But the previous doc had said that it computes the Rhat according to the reference and that is what the reference says. This is why we had the discussion in the meeting just now.

I will update the code in a bit according to the definition, while introducing an argument warm_up which determines which fraction is supposed to be thrown away and set it’s default to 0.

On Jun 2, 2020, at 1:33 PM, Martin Robinson <notifications@github.com<mailto:notifications@github.com>> wrote:



@martinjrobins commented on this pull request.

In pints/_diagnostics.py<https://github.com/pints-team/pints/pull/1156#discussion_r433808363>:

>      r"""
-    Calculates :math:`\hat{R} = \sqrt{(W(n - 1) / n + (1 / n) B) / W}` as per
-    [1]_ for a single parameter. It does this after splitting each chain into
-    two.
-    """
-    W = within(samples)
-    B = between(samples)
-    t = len(samples[0])
+    Returns the convergence measure :math:`\hat{R}` according to [1]_.
+
+    :math:`\hat{R}` diagnoses convergence by checking mixing and stationarity
+    of :math:`M` chains (at least two, :math:`M\geq 2`). To diminish the
+    influence of starting values, the first half of each chain is considered
+    as warm up and does not enter the calculation. Subsequently, the truncated


looking at the code, it looks like it doesn't discard warm-up. Did I miss something?

DavAug:
@MichaelClerx @ben18785 @martinjrobins @chonlei 

I refactored the rhat computation a little bit and adjusted the documentation. 
I used some numpy magic to combine `pints.rhat()` and `pints.rhat_all_params` into a single function `pints.rhat()`. This function takes multiple chains in form of numpy.ndarray as before, but now both dimensions (m, n) and (m, n, p) are accepted, where m is the number of chains, n is the number of samples per chain, and p is the number of parameters.

DavAug:
I also added some asserts which should be satisfied for the rhat computation to make sense. One of the assertions "the number of chains has to be at least 2, m>=2" breaks a test for the `pints.summary()`, because here only a single chain is passed to rhat. I am tempted to adjust the test as mixing of chains can only be tested if you have at least 2. 

Or is the rhat computation for a single chain useful in certain circumstances? In that case we could just remove the assert.

What do you think?

docs/source/diagnostics.rst
```diff
@@ -9,14 +9,11 @@ Pints provides a number of functions to diagnose MCMC progress and convergence.
 Overview:
 
 - :func:`rhat`
-- :func:`rhat_all_params`
 - :func:`effective_sample_size`
```
ben18785, line 12: Sorry, realise they weren't before, but can we make these alphabetical?

examples/sampling/adaptive-covariance-haario.ipynb
```diff
@@ -139,7 +139,7 @@
     "\n",
     "# Check convergence using rhat criterion\n",
     "print('R-hat:')\n",
-    "print(pints.rhat_all_params(chains))\n",
+    "print(pints.rhat(chains))\n",
```
ben18785, line 142: Thanks for changing these: what'd I'd prefer though would be to change all of these over to using the sampler results object (the one that prints a table of parameter quantiles). I think these are much neater and display everything that's needed. See the ``adaptive-covariance-haario-bardenet.ipynb`` for an example of this.

examples/sampling/adaptive-covariance-rao-blackwell.ipynb
```diff
@@ -138,7 +138,7 @@
     "\n",
     "# Check convergence using rhat criterion\n",
     "print('R-hat:')\n",
-    "print(pints.rhat_all_params(chains))\n",
+    "print(pints.rhat(chains))\n",
```
ben18785, line 141: Same as above. For all MCMC notebooks, I think this should be the case.

pints/_diagnostics.py
```diff
@@ -55,72 +55,185 @@ def effective_sample_size(samples):
...
+
+    Here, :math:`\psi _{ij}` is the jth sample of the ith
+    chain, :math:`\bar{\psi _j}=\sum _{i=1}^{n'}\psi _{ij}/n'` is the within
+    chain mean of the parameter :math:`\psi` and
+    :math:`\bar{\psi } = \sum _{j=1}^{m'}\bar{\psi} _{j}/m'` is the between
+    chain mean of the within chain means.
 
     References
     ----------
     ..  [1] "Bayesian data analysis", 3rd edition, Gelman et al., 2014.
```
ben18785, line 173: Can we point to the relevant chapter in this? Think it's chapter 11. Also, I think we should cite Stan here, since the splitting of chains in two is what they did (I don't think it's mentioned in Gelman).

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
-        self.assertEqual(y.shape, (3, 8, 5))
+        # Pass bad warm-up arguments
+        chains = np.empty(shape=(2, 4))
 
-    def test_rhat(self):
-        # Tests that rhat works
+        # warm-up greater than 100%
+        warm_up = 1.1
+        message = (
+            'warum_up is set to 1.1. warm_up only takes values in [0,1].')
```
ben18785, line 156: ''warm-up''

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
+        self.assertRaisesRegex(
+            ValueError, message[0], pints.rhat, chains, warm_up)
 
-        x = np.array([[1.0, 1.1, 1.4, 1.3, 1.3],
```
ben18785, line 109: Is the reason this test has a different result this time because the chains are being split? As in Rhat=1.43115 before, now it's 2.33...

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
-        # Tests that reorder_all function reshapes correctly
+    def test_bad_rhat_inputs(self):
+        # Tests whether exceptions are thrown should the input to rhat not be
+        # valid
+
+        # Pass chain of dimension 1
+        chains = np.empty(shape=1)
+        message = (
+            'Dimension of chains is 1. '
+            + 'Method computes R^hat for one '
```
ben18785, line 134: Rhat rather than R^hat

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
-    def test_reorder_all(self):
-        # Tests that reorder_all function reshapes correctly
+    def test_bad_rhat_inputs(self):
+        # Tests whether exceptions are thrown should the input to rhat not be
+        # valid
+
+        # Pass chain of dimension 1
+        chains = np.empty(shape=1)
+        message = (
+            'Dimension of chains is 1. '
```
ben18785, line 133: Can I just check -- does this method still work for models with only a single parameter?

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
+            'or multiple parameters and therefore only accepts 2 or 3 '
+            'dimensional arrays.')
+        self.assertRaisesRegex(
+            ValueError, message[0], pints.rhat, chains)
+
+        # Pass chain of dimension 4
+        chains = np.empty(shape=(1, 1, 1, 1))
+        message = (
+            'Dimension of chains is 4. '
+            + 'Method computes R^hat for one '
```
ben18785, line 144: Rhat

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
+        # Tests that rhat works
 
-        test = np.random.normal(loc=0, scale=1, size=(4, 10, 3))
-        y = np.array(pints._diagnostics.reorder(0, test))
-        self.assertEqual(y.shape, (8, 5))
+        # Tests Rhat computation for one parameter, chains.shape=(2, 4)
+        chains = np.array([[1.0, 1.1, 1.4, 1.3],
+                          [1.0, 2.0, 3.0, 4.0]])
+        self.assertAlmostEqual(
+            pints._diagnostics.rhat(chains), 2.3303847470550716, 6)
```
ben18785, line 99: This looks like a new value of Rhat versus the old test value...

pints/tests/test_diagnostics.py
```diff
@@ -79,36 +79,99 @@ def test_within(self):
...
+        self.assertRaisesRegex(
+            ValueError, message[0], pints.rhat, chains, warm_up)
 
-        x = np.array([[1.0, 1.1, 1.4, 1.3, 1.3],
-                      [1.0, 2.0, 3.0, 4.0, 5.0]])
-        self.assertAlmostEqual(pints._diagnostics.rhat(x), 1.433115, 6)
+        # Negative warm-up
+        warm_up = -0.1
+        message = (
+            'warum_up is set to -0.1. warm_up only takes values in [0,1].')
```
ben18785, line 163: warm-up

docs/source/diagnostics.rst
DavAug: Done

pints/_diagnostics.py
DavAug: Yes chapter 11, page 285 equation 11.4. I implemented the method strictly according to this definition. Do you maybe mean the rank-based computation that Stan does, and is not mentioned in this book?

pints/tests/test_diagnostics.py
DavAug: I referring here to the input variable `warm_p`. I would therefore leave it like that. But I can add some `.` marks if that make sit clearer.

pints/tests/test_diagnostics.py
DavAug: Yes the old rhat did not test for stationarity. This rhat always splits the chains in half and that's why the old and new rhat do not agree. But I implemented the methods to work for one parameter only, and afterwards realised that broadcasting also allows for the computation of multiple parameters with the same expressions. And for multiple params the rhat agrees with the old results. So I'm pretty confident that rhat is correct.

pints/tests/test_diagnostics.py
DavAug: Done

pints/tests/test_diagnostics.py
DavAug: Yes it works for input dimensions (m, n) and (m, n, p) where m is the number of chains, n the number of samples per chain and p the number of parameters.

pints/tests/test_diagnostics.py
DavAug: This is because the old Rhat did not split the chains in half.

pints/_diagnostics.py
ben18785: Nah, I was just meaning about splitting the chains in two. Does Gelman suggest this? If so, fine to keep as is. I think we can save doing any rank-based stuff for now.

pints/tests/test_diagnostics.py
ben18785: My comment was just about spelling warm_up rather than warum_up

pints/tests/test_diagnostics.py
ben18785: Ok, cool

pints/tests/test_diagnostics.py
ben18785: Cool

pints/tests/test_diagnostics.py
ben18785: Again, just warum_up vs warm_up

pints/tests/test_diagnostics.py
DavAug: Ahh I see :D

Issue #939, opened 2019-09-24
Add a "before you begin" notebook


MichaelClerx:
Some chapter 0 to the PINTS story, that doesn't explain things, but links to our favourite resources on:

- Python (see also contributing.md)
- Something about modelling ?
- Pages that explain inference (Ben's book and videos?)
- Points people towards CONTRIBUTING.md for everything else?

MichaelClerx:
@simonmarchant I guess your notebook goes somewhere between this one and the "first examples" ones?

simonmarchant:
@MichaelClerx right now it does, but I'm happy to add this stuff if you'd like? I do an ultra-short explanation of Bayes' rule so could add links for further reading there, could add a section at the start with links to python resources, and a bit at the end for "where next?". Or if you'd prefer that "before you begin" is entirely separate from any explanations then I'm happy to do it as a separate markdown file - shouldn't take long.

MichaelClerx:
I'm not sure! I was hoping we could just link to stuff for the basic basic introduction...
(e.g. Ben has done quite a bit of work explaining bayesian inference already, elsewhere)
Probably best to have a chat about this at some point, with. e.g. @ben18785 and @martinjrobins

MichaelClerx:
@ben18785 wrote

> @MichaelClerx I like this idea -- how about if we put these tutorials in a separate repo?

Not sure why we'd need to move it?

I would very much like to have stuff like this in the PINTS repo, but not sure about the best way to structure it, and how much detail we need to provide VS where we should link to external resources!

MichaelClerx:
If that means a separate repo where the goal is explaining stats instead of showing how PINTS works, that's fine by me too :D

We could do a version of your book with PINTS examples ;-)

MichaelClerx:
Done in  #1159

PR #1159, opened 2020-06-02
Tutorial for complete beginners

simonmarchant:
This is a jupyter notebook tutorial that works through:-
* optimisation with simple code mirroring the maths
* optimisation using pints
* sampling mirroring the maths
* sampling using pints

The intended audience is complete newcomers to bayesian inference, who may need an extra walkthrough before trying the first-example notebook.


MichaelClerx:
Looks nice, and potentially very useful! 

But I'd like to have a think (and chat) about how much intro we want/need, and at which point it goes too far off focus. Shall we discuss in https://github.com/pints-team/pints/issues/939 ?

chonlei:
Can we also link this to the existing toy model? This is part of the stochastic degradation toy model (the mean function of the model) in Pints.

chonlei:
Or in fact maybe have another degradation model model that's not stochastic...?

ben18785:
@MichaelClerx I like this idea -- how about if we put these tutorials in a separate repo?

MichaelClerx:
> @MichaelClerx I like this idea -- how about if we put these tutorials in a separate repo?

I'm coming round to this idea actually! Should we set up e.g. `pints-team/bayesian-inference-tutorial` and then everyone who's interested can just add tutorials like these there? And then eventually you could clean it up, turn it into some educational resource :D

We could then link to it from pints/examples too

MichaelClerx:
A different set-up that might be cool would be to have a tiny section at the top of the examples page with these three sections:

**Before you begin**
- Installing PINTS (links to the main installation instructions, not in notebook format!)
- Using Python (an MD file with curated links to e.g. Python for non-programmers, Python for matlab fans, Python for R users)
- An introduction to optimisation and Bayesian inference (this notebook, with a bit of the overlap with the above removed)

Does that sound reasonable @simonmarchant @ben18785 @chonlei @martinjrobins ?

Just trying to work out how to include this in PINTS without having too much overlap / disjoint bits of documentation!

martinjrobins:
yea, this sounds like a nice way to organise it

README.md
```diff
@@ -44,7 +44,7 @@ PINTS comes with a number of [detailed examples](examples/README.md), hosted her
 In addition, there is a [full API documentation](http://pints.readthedocs.io/en/latest/), hosted on readthedocs.io.
 
 
-## Installing PINTS
+## Installing PINTS
```
MichaelClerx, line 47: Is this a crucial space? Because my editor will cut this out next time I open the file :D

simonmarchant:
I've added changes suggested in https://github.com/pints-team/pints/pull/1159#issuecomment-645494077 @MichaelClerx - what do you think?

I'm sure there are plenty of resources that I haven't thought of, by the way, so the more resource suggestions the merrier!

README.md
simonmarchant: That was quick! No it's not, I just undid a change I made and didn't notice the space :-|

simonmarchant: Fixed.

MichaelClerx:
Cheers!

Maybe this instead of the PDF for matlab ? https://numpy.org/devdocs/user/numpy-for-matlab-users.html

MichaelClerx:
> For pharmacokinetics, the system of interest is the distribution of a drug in the body.

This bit's a bit out of the blue for me. Maybe introduce it (one sentence) as one of many possible applications?

MichaelClerx:
@ben18785 have you been able to have a look at the notebook itself? Or maybe @chonlei wants to have a go? I'm happy with the readme changes and the stuff around it :-)

MichaelClerx:
Hi @simonmarchant !

I've had a detailed look at the text now (although I had to stop somewhere in the sampling section). It looks good! I think it will be really useful to have this, but we might need more information (or links to where it can be found) in this example. I made some direct changes, but also made some notes for you. Posting them here as PR comments on notebooks aren't easy :-)

------------------------------------------------------------------------------------

When writing text in git, it's helpful to put line breaks after each line. (instead of writing a paragraph on each line.)

## Bayesian probability revision

> For our example, we are using a Gaussian probability distribution (i.e. normal distribution) to compare to measured data, so $P(X) = \frac{1}{\sqrt{2\pi\sigma^2}}e^{\frac{-(X-data)^2}{2\sigma^2}}$.
This seems reasonable because random noise often behaves in a roughly Gaussian way.
We decide on a uniform prior between 0 and 10, to show that before analysis we believe that $k$ could equally be any value between 0 and 10.

"we are using" Guessing this was moved here at some point?
I'd either (1) make this slower and provide some links e.g. to Ben's lecture pages, or;
(2) make this an intermezzo somewhere in the text - but make sure that it's clear from the main intro of this page that we're going to explain opt & inference, _not_ bayesian stats itself!

Don't call it P(X) if it's called P(y|theta) in the line before :D
I would go a bit slower here and discuss all three parts of Bayes equation (you almost do this already)

> This is what we'll use throughout to calculate $P(\theta|y)$, the probability of any given parameter set $\theta$ given the observations $y$.

I don't think we actually do this! Better make it more accurate :-)
For example, in optimisation we know that, since the denominator is positive, the peak of numerator = peak of what we're looking for, so we can ignore it
In sampling, we're trying to get a sample such that SomeMeasure(our sample) is approximately equal to SomeMeasure(true distribution)

## Bayesian optimisation

Careful with terminology here!
Many systems, including yours, have an _input_ (time-dependent, not part of the model), an _output_ (some function of the model's state), and _parameters_ (used in model equations, or to formulate the output function).
Don't call the parameters input, when discussing this model! (It's fine for the forwardmodel, which abstracts away/fixes the input)

Code: `times` is not a parameter!

Code:

> In
> y: y-axis value
> t: x-axis value
> k: gradient parameter
>
> Out: gradient at the given point

Just stick to the PINTS style conventions here instead of introducing a new one :D

> # You can solve ODEs with the Forward Euler method

Instead of having these comments above methods, I'd break out of code here and into a markdown block!
You've explained above that we've got a model, then given the model, so then switch back to markdown and explain we'll need an integration method, etc.
(`times` and the default parameters can move to one of the later blocks then too)

> While we can estimate concentration (y) at a given time (t) numerically by plotting from y(t=0), we have to guess a value of k. We are trying to find the true value of k.

Maybe turn this around? _if_ we know $k$, and a given $y(t=0)$, then we can simulate the time course y(t), but we have to guess $k$. (Also you kind of said all this above already, so maybe re-read and streamline a bit? Next line again is "We're trying to find which value of k makes our model output...")

> # Optimise it with scipy

Again, break out into markdown first, and explain what we're going to do

## Using Pints

It seems we've settled on old-school "PINTS" as the way of writing Pints, so best stick to that throughout.

> our estimator estimates $k$ pretty

It's not one of these so be careful with terminology: https://en.wikipedia.org/wiki/Estimator


> So how would we do what we've done above with PINTS?

This makes it seem a bit like we're crowbarring PINTS into a situation where it isn't needed :D
Maybe rephrase a bit?
You could say something about how PINTS formalises the steps above, and easily lets you change out different error measure / likelihoods, different optimisation methods, lets you re-use them in different settings.

> problem = pints.SingleOutputProblem(PintsOneComp(), times, noisy_data)
> error_measure = pints.SumOfSquaresError(problem)
> optimisation = pints.OptimisationController(error_measure, [1], method=pints.XNES)
> optimisation.set_log_to_screen(False)
> parameters, error = optimisation.run()

I know you're trying to show that it's fast here, but might still be better to have a comment above each line, saying what the steps are! ("Create a model instance", "Define an statistical problem")

You've made an important switch here too, from a model of the system of interest (with its input y(t), output z(state, parameters, t), and parameters), to a statistical model (which takes parameters and times as input, and returns values as output).
This is a big conceptual switch, and it's easy to get the two confused, so this needs attention / a paragraph!
(Either before or after the example, maybe best after)

## Model section?

I think it would be a good idea to define the model (ODEs) in a separate section, before the optimisation.
You can also use this to show how the mechanistic model gets turned into a statistical one (and input becomes fixed)
Then when the optimisation section starts, say something along the lines of "we start with the problem of finding a single k that makes the model output match the measurements", or "we ask the question which k will...".
And then in the sampling section say "Now we tackle a different problem, and try to find a _distribution_ of $k$ values that blah blah blah"

## Bayesian sampling

> The method we'll use here is Metropolis Hastings.

Add link to wiki, or _emphasise_ it's a new term (that you'll explain below).

> In our exponential decay model,

Why call it this, when we went to the trouble of introducing PK earlier?

> Metropolis Hastings creates

"The Metropolis-Hasting algorithm"?

> "with SD given" --> standard deviation

Is it a "covariate matrix" ???

> prior x likelihood

it took me forever to realise you meant prior times likelihood, just use the symbols you had before, or write `$\text{prior} \cdot \text{likelihood}$`

> This approximates your posterior.

Don't say that like it's obvious. It really isn't! Apologise to the reader :D Give links, say they can read up somewhere else to see why it works.

Add some markdown between the code below

## Using PINTS

capitals, emphasise that this is the same problem set up as before!

Add a recap/summary at the end

simonmarchant:
https://github.com/pints-team/pints/pull/1159#issuecomment-651707197 This is brilliant, thankyou @MichaelClerx! And apologies that there was so many simple mistakes to correct - I've looked at it so much that I can't see those things anymore (also some things - especially terminology - I actually didn't know about...). 

I think I can fix these things over the next week.

I have two questions. 
> When writing text in git, it's helpful to put line breaks after each line. 

Do you mean "writing text in markdown" here? Or in my git commit messages?

> Add link to wiki, or emphasise it's a new term (that you'll explain below).

Would a short glossary at the bottom be useful, or overkill?

MichaelClerx:
> Do you mean "writing text in markdown" here? Or in my git commit messages?

:facepalm: oh sorry! I meant "writing text in markdown that gets stored in git". Because git was designed for code it counts the number of changed lines, so if you have a paragraph on a line and someone adds a comma it marks the whole thing as changed which is not ideal.
(same holds for latex stored in git!)

MichaelClerx:
> Would a short glossary at the bottom be useful, or overkill?

Up to you! It can be hard to explain these terms out of context, so I'd say give it a go and then if you don't like the result just remove the glossary again :D

simonmarchant:
> > Do you mean "writing text in markdown" here? Or in my git commit messages?
> 
> facepalm oh sorry! I meant "writing text in markdown that gets stored in git". Because git was designed for code it works with _lines_ as the fundamental units of data, so if you have a paragraph on a line and someone adds a comma it marks the whole thing as changed which is not ideal.
> (same holds for latex stored in git!)

TIL: this.  Thanks! :smile:

simonmarchant:
Hi @MichaelClerx this is taking a little longer than I expected, but I haven't forgotten! I'm trying to write the short section at the end of the optimisation section, but I'm not sure I've written it well / understood it properly. Could you take a look?

Your comment:-
> You've made an important switch here too, from a model of the system of interest (with its input y(t), output z(state, parameters, t), and parameters), to a statistical model (which takes parameters and times as input, and returns values as output).
This is a big conceptual switch, and it's easy to get the two confused, so this needs attention / a paragraph!

What I've written now:-
> There is an important conceptual change in the code above. Previously, we have dealt with a _physical_ model of the real-life system of interest (the function `onecomp`). This model takes $y(t_n)$ as input, and for a given parameter outputs $y(t_{n+1})$. PINTS abstracts this to a _statistical_ model, which takes different parameters and times as input, and outputs values $y(t)$ for all time values. The same things are happening, but PINTS does them 'under the hood'.

MichaelClerx:
> Hi @MichaelClerx this is taking a little longer than I expected, but I haven't forgotten! I'm trying to write the short section at the end of the optimisation section, but I'm not sure I've written it well / understood it properly. Could you take a look?
> 
> Your comment:-
> 
> > You've made an important switch here too, from a model of the system of interest (with its input y(t), output z(state, parameters, t), and parameters), to a statistical model (which takes parameters and times as input, and returns values as output).
> > This is a big conceptual switch, and it's easy to get the two confused, so this needs attention / a paragraph!
> 
> What I've written now:-
> 
> > There is an important conceptual change in the code above. Previously, we have dealt with a _physical_ model of the real-life system of interest (the function `onecomp`). This model takes $y(t_n)$ as input, and for a given parameter outputs $y(t_{n+1})$. PINTS abstracts this to a _statistical_ model, which takes different parameters and times as input, and outputs values $y(t)$ for all time values. The same things are happening, but PINTS does them 'under the hood'.

Great start! But in many cases we have a model that takes $y(t_n, p)$ as input and produces $\frac{d}{dt}y(t_n, p)$. Part of wrapping this into a "statistical model" is that we add an entire solution method (plus often a $y(t_0)$), so that it can take $p$ as input to produce $y(t_0, p), y(t_1, p), y(t_n, p)$. All of this is done by the user so not really PINTS that does the abstraction in this case. It's up to the user to write the "ForwardModel" problem that wraps the user problem in a statistical model

simonmarchant:
Thanks very much for all the advice on this, @MichaelClerx !

PR #1165, opened 2020-06-20
Transformation objects

### LINKED ISSUE #1070: Convenience functions for transforming parameters

chonlei:
Closes #1070.

Got a simple working framework for parameter transformation. It's not ready to merge yet, but I think it should be a good start -- it's quite clean both in the code and in terms of the user API.

Here's a few things need to be done/discussed before merging:
- [x] Work on S1 for transformation
  - [x] Add `log_jacobian_det_S1()` method
- [x] Add a few more transformation (will do this _after_ we all happy with the structure!)
  - [x] general logit transformation (interval [a, b] to all real number; instead of [0, 1])
  - [x] simple scaling transformation
- [x] Need a "ComposeTransform" class in some way
- [x] Add lot of tests...
  - [x] Test all Transform classes
  - [x] Update OptimisationController tests
  - [x] Update MCMCController tests
- [x] Add docs
- [x] Change `Transform` to `Transformation`
- [x] Add a `many_to_model()` method with slow for loop in the base class
- [x] Maybe add an notebook example (with a simple toy PDF) to illustrate the effect of the sampling (incorrectly) missing out Jacobian
- [x] Explain Jacobian and transformation in the docstring...


examples/README.md
```diff
@@ -16,7 +16,9 @@ relevant code.
 
 
 ## Digging deeper
-- [Searching in a transformed parameter space](./optimisation/transformed-parameters.ipynb)
+- [Optimisation in a transformed parameter space with a model wrapper](./optimisation/transformed-parameters-model-wrapper.ipynb)
```
MichaelClerx, line 19: The do-it-yourself example can go, if this works

chonlei: We could maybe have one folder that shows some "hack around Pints", then we could put this in, as well as some examples showing using Pints for non-time series problems etc.

chonlei:
@mirams might be interested in this too? Pints is now trying to include some convenience functions for transforming parameters. Particularly the new example notebooks, it's a bit like the one we have in the WIREs review.

chonlei:
@ben18785 would you be able to check that I am doing the sensitivity transformation correct? I am pretty sure it should be correct for the error measure case, just hope there's no extra rules that I've missed/forgotten for log-PDFs... thanks!

chonlei:
I will add the rest of the transforms once we all happy with the structure.

MichaelClerx:
I think we're missing the bit where we detransform the parameters before logging them (to CSV). Best to read the whole controller run() methods and see if they're used anywhere else, e.g. for error output

chonlei:
> I think we're missing the bit where we detransform the parameters before logging them (to CSV). Best to read the whole controller run() methods and see if they're used anywhere else, e.g. for error output

Good spot! I have updated that now -- detransform the parameters before logging them to CSV. Also checked that the stored in memory samples (self._samples) are detransformed too. Can't see it's using `samples` in error output.

MichaelClerx:
So uhm, "transform" or "transformation"?

I'd go with transformation myself

chonlei:
> So uhm, "transform" or "transformation"?
> 
> I'd go with transformation myself.
> 
> @ben18785 @martinjrobins @chonlei what do you think? I'm a bit unclear on when mathematicians (it's definitely only mathematicians) use "transform" as a noun

Add it as a TODO to change "Transform" to "Transformation" now.

pints/_mcmc/__init__.py
```diff
@@ -293,13 +293,60 @@ class MCMCController(object):
...
         The class of :class:`MCMCSampler` to use. If no method is specified,
         :class:`HaarioBardenetACMC` is used.
+    transform : pints.Transform
+        A :class:`pints.Transform` that transform the model parameter space to
+        search space. If no transform is specified, search will be performed
+        in the model space.
     """
 
-    def __init__(self, log_pdf, chains, x0, sigma0=None, method=None):
+    def __init__(
```
MichaelClerx, line 302: I think I'd prefer the keep `method` as the final argument

pints/_mcmc/__init__.py
```diff
@@ -293,13 +293,60 @@ class MCMCController(object):
     method : class
         The class of :class:`MCMCSampler` to use. If no method is specified,
         :class:`HaarioBardenetACMC` is used.
+    transform : pints.Transform
```
MichaelClerx, line 296: I think I'd prefer the keep method as the final argument

pints/_mcmc/__init__.py
```diff
@@ -293,13 +293,60 @@ class MCMCController(object):
...
+
+        # Transform everything
+        # From this point onward the MCMC sampler will see only the
+        # transformed search space and will know nothing about the model
+        # parameter space.
+        if transform:
+            log_pdf = transform.apply_log_pdf(log_pdf)
+            x0 = [transform.to_search(x) for x in x0]
+            #
+            # Following Eq. 5 in
```
MichaelClerx, line 318: This is important information that should be in the main docstring for the controller

pints/_mcmc/__init__.py
```diff
@@ -293,13 +293,60 @@ class MCMCController(object):
...
+                sigma0 = np.asarray(sigma0)
+                n_parameters = log_pdf.n_parameters()
+                # Make sure sigma0 is a (covariance) matrix
+                if np.product(sigma0.shape) == n_parameters:
+                    # Convert from 1d array
+                    sigma0 = sigma0.reshape((n_parameters,))
+                    sigma0 = np.diag(sigma0)
+                else:
+                    # Check if 2d matrix of correct size
+                    sigma0 = sigma0.reshape((n_parameters, n_parameters))
```
MichaelClerx, line 343: This should be a check with a user friendly error that we test for!

pints/_mcmc/__init__.py
```diff
@@ -293,13 +293,60 @@ class MCMCController(object):
...
+            # C(x) = (d/dp g(E(p))) C(p) (d/dp g(E(p)))^T
+            #        = J^{-1}(E(x)) C(p) J^{-1}(E(x))^T
+            #
+            # Using the property that J^{-1} = dg/dp, from the inverse function
+            # theorem, i.e. the matrix inverse of the Jacobian matrix of an
+            # invertible function is the Jacobian matrix of the inverse
+            # function.
+            #
+            if sigma0 is not None:
+                # Transform sigma0 if provided
```
MichaelClerx, line 333: Around line 370 it says

>         # Don't check initial standard deviation: done by samplers!

That will need to go, if we _do_ check it here!
In that case we should separate the checking of sigma0 from the transforming

pints/_mcmc/__init__.py
```diff
@@ -293,13 +293,60 @@ class MCMCController(object):
...
+                n_parameters = log_pdf.n_parameters()
+                # Make sure sigma0 is a (covariance) matrix
+                if np.product(sigma0.shape) == n_parameters:
+                    # Convert from 1d array
+                    sigma0 = sigma0.reshape((n_parameters,))
+                    sigma0 = np.diag(sigma0)
+                else:
+                    # Check if 2d matrix of correct size
+                    sigma0 = sigma0.reshape((n_parameters, n_parameters))
+                # Transform sigma0
```
MichaelClerx, line 344: This is the only part of the "if sigma" block that's part of the transformation handling

pints/_mcmc/__init__.py
```diff
@@ -399,6 +446,8 @@ def chains(self):
         """
         # Note: Not copying this, for efficiency. At this point we're done with
         # the chains, so nothing will go wrong if the user messes the array up.
+        # Note: Any inverse transform for parameters should be performed before
+        # stored to memory.
```
MichaelClerx, line 450: - stored --> storing
- add "or logging to screen"

pints/_mcmc/__init__.py
MichaelClerx: If users give 4x4x4x4 instead of 16x16 I think that's an error here. Same as given 9x1x1x1 instead of 3x3. Bit stricter than in the other places :-)

pints/_mcmc/__init__.py
MichaelClerx: And probably move x0 checking up here too then (or move transform checking down)

pints/_mcmc/__init__.py
```diff
@@ -399,6 +446,8 @@ def chains(self):
         """
```
MichaelClerx, line 446: If was expecting something near line 398 "self._needs_sensitivities = ..." that checks that the transform provides them too. I imagine this is not always the case?

pints/_mcmc/__init__.py
```diff
@@ -735,10 +784,17 @@ def run(self):
             # Write samples to disk
```
MichaelClerx, line 784: - [ ] Check near line 686 ``# Store sample in memory``
- [ ] Add tests that do something like transform from a space [0, 1] to a space [10, 11], and make sure that all logged points are in [10, 11], whether that's logging to disk or to memory

pints/_mcmc/__init__.py
```diff
@@ -774,16 +830,30 @@ def run(self):
...
 
-        # Store generated chains in memory
-        if self._chains_in_memory:
-            self._samples = samples
-
         # Store evaluations in memory
         if self._evaluations_in_memory:
             self._evaluations = evaluations
 
+        # Inverse transform to model space if transform is provided
```
MichaelClerx, line 837: I think I'd prefer to do this on-the-fly, instead of after we're done! But maybe you can convince me otherwise :D

pints/_mcmc/__init__.py
```diff
@@ -774,16 +830,30 @@ def run(self):
...
+        if self._chains_in_memory:
+            if self._transform:
+                n_c, n_s, n_p = samples.shape
+                samples_user = np.zeros((n_c, n_s, n_p))
+                for c in range(n_c):
+                    for s in range(n_s):
+                        samples_user[c, s, :] = \
+                            self._transform.to_model(samples[c, s, :])
```
MichaelClerx, line 845: If this is going to be the standard use we should probably make it so that transformations can do this with 1 numpy command instead of in a nested for loop (much slower!)

pints/_mcmc/__init__.py
```diff
@@ -774,16 +830,30 @@ def run(self):
...
+            else:
```
MichaelClerx, line 846: I'd get rid of this and write `samples = samples_user` in the above, followed by del(samples_user)

pints/_optimisers/__init__.py
```diff
@@ -324,6 +329,46 @@ def __init__(
...
+        # From this point onward the optimisation will see only the
+        # transformed search space and will know nothing about the model
+        # parameter space.
+        if transform:
+            if self._minimising:
+                function = transform.apply_error_measure(function)
+            else:
+                function = transform.apply_log_pdf(function)
+            x0 = transform.to_search(x0)
+            #
```
MichaelClerx, line 342: This should go in a docstring somewhere. Or do we need a method `transform.apply_covariance_matrix` ??

pints/_optimisers/__init__.py
```diff
@@ -598,7 +643,12 @@ def run(self):
             print('Unexpected termination.')
             print('Current best score: ' + str(fbest))
             print('Current best position:')
-            for p in self._optimiser.xbest():
+            # Inverse transform search parameters
```
MichaelClerx, line 646: You can stick some newlines around this, and do it outside of this if block (and only once!)

pints/_optimisers/__init__.py
```diff
@@ -753,7 +809,8 @@ class Optimisation(OptimisationController):
     """ Deprecated alias for :class:`OptimisationController`. """
 
     def __init__(
-            self, function, x0, sigma0=None, boundaries=None, method=None):
+            self, function, x0, sigma0=None, boundaries=None, method=None,
```
MichaelClerx, line 812: Again think method is better off staying the last argument

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
+#
+# Transformation functions
```
MichaelClerx, line 2: "Classes that automate parameter transformation" or something?

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+from scipy.special import logit, expit
+
+
+class Transform(object):
+    """
+    Abstract base class for objects that provide some convenience parameter
```
MichaelClerx, line 17: less emphasis on the convenience here, they do tricky things :-)

"objects that provide transformations between two parameter spaces" or something

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+    transformation from the model parameter space to a search space.
+
+    If ``t`` is an instance of a ``Transform`` class, you can apply
```
MichaelClerx, line 20: Maybe not `t` but e.g. `trans` ?
(Also maybe `q` instead of `p`? Or are we using `$\theta$` everywhere?)

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+    the transformation of a parameter vector from the model space ``p`` to the
+    search space ``x`` by using ``x = t.to_search(p)`` and the inverse by using
+    ``p = t.to_model(x)``.
+    """
+    def apply_log_pdf(self, log_pdf):
```
MichaelClerx, line 25: I'm not sure "apply_" is the right term, so maybe replace with just "def log_pdf(..)" ?

MichaelClerx: Or "convert_log_pdf" etc?

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+        """
+        Returns a transformed boundaries class.
+        """
+        return TransformedBoundaries(boundaries, self)
+
+    def jacobian(self, x):
+        """
+        Returns the Jacobian for a parameter vector ``x`` in the search space.
+
+        *This is an optional method. It is needed when `sigma0` is provided in
```
MichaelClerx, line 47: As stated earlier, might be better to have a general covariance matrix transformation method!

(Even if we don't would update this to say it can be used for any old sigma, not just sigma0 in a controller!)

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+        :meth:`pints.OptimisationController` or :meth:`pints.MCMCController`,
+        or when the method requires ``evaluateS1()``.*
+        """
+        raise NotImplementedError
+
+    def log_jacobian_det(self, x):
+        """
+        Returns the logarithm of the absolute value of the Jacobian
```
MichaelClerx, line 55: What's a "jacobian determinant"? Can write the same without introducing new terms?

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+
+    def apply_boundaries(self, boundaries):
+        """
+        Returns a transformed boundaries class.
+        """
+        return TransformedBoundaries(boundaries, self)
+
+    def jacobian(self, x):
+        """
+        Returns the Jacobian for a parameter vector ``x`` in the search space.
```
MichaelClerx, line 45: Probably worth adding the maths to say _which_ Jacobian, of the many involved in time series inference :-)

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+        """
+        Returns the dimension of the parameter space this transformation is
+        defined over.
+        """
+        raise NotImplementedError
+
+    def to_model(self, x):
+        """
+        Returns the inverse of transformation of a vector from the search space
+        ``x`` to the model parameter space ``p``.
```
MichaelClerx, line 75: Can we use `q` here? There's too many x's in the world :p

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+    def n_parameters(self):
+        """
+        Returns the dimension of the parameter space this transformation is
+        defined over.
+        """
+        raise NotImplementedError
+
+    def to_model(self, x):
+        """
+        Returns the inverse of transformation of a vector from the search space
```
MichaelClerx, line 74: Maybe "Transforms a parameter vector x from search space to model space" ?
(and the other way round for the one below)

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+        """
+        Returns the forward transformation of a vector from the model parameter
+        space ``p`` to the search space ``x``.
+        """
+        raise NotImplementedError
+
+
+class TransformedLogPDF(pints.LogPDF):
+    """
+    A log-PDF that is transformed from the model space to the search space.
```
MichaelClerx, line 89: Something like 

``A :class:`pints.LogPDF` that accepts parameters in a transformed search space`` ? (same for others)

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+                             'must match.')
+
+    def __call__(self, x):
+        # Get parameters in the model space
+        p = self._transform.to_model(x)
+
+        # Compute LogPDF in the model space
+        logpdf_nojac = self._log_pdf(p)
+
+        # Calculate the PDF change of variable. From Wikipedia:
```
MichaelClerx, line 115: This kind of stuff goes in the main docstrings. Users will want to know! (Or will want to be assured that whoever wrote the class did some fancy maths so that they don't have to)

pints/_mcmc/__init__.py
chonlei: OK, good point, maybe just do some check for `sigma0.shape == (n_parameters, n_parameters)`?

pints/_mcmc/__init__.py
chonlei: Interesting one, I thought I would like the run keep in the search space to avoid any potential issue. If one decided to use the stored samples to look at the next tell() point in the future or such like. Not 100% convinced that one way is the best.

pints/_mcmc/__init__.py
chonlei: Ha!!! I did that and removed it... Because that'd involve in the ComposeTransforms checking whether it's a single parameter or a set of parameters... And that in the Transform docs, it'd become slightly tricky, because we have to tell users whoever try to implement their own that the `Transform.to_xxx()` can take either `(n_parameters,)` 1D array or `(n_sets, n_parameters)` 2D array...

pints/_transformation.py
chonlei: Cool, either "convert_log_pdf" or just "log_pdf" sounds good. So which one? 😄

pints/_transformation.py
chonlei: Which of the many involved in time series inference?

pints/_transformation.py
chonlei: What do you mean by without introducing new terms? It's the determinant of the Jacobian (if this helps...).

pints/_transformation.py
chonlei: so `p` in model space and `q` in search space?

pints/_transformation.py
chonlei: Haha! OK, I thought this was quite boring, so didn't want to bother users in the main docstrings.

pints/_mcmc/__init__.py
MichaelClerx: Yeah and a `ValueError` if it isn't

pints/_transformation.py
MichaelClerx: Like, dx/dwhat?

pints/_transformation.py
MichaelClerx: Yup. But not if you don't want to :D

pints/_transformation.py
MichaelClerx: I can imagine users having some vague memory that you need to do something special, and then being happy that this method does it.

Or alternatively users who know exactly what you do, being worried that the docs don't mention that this is done!

pints/_transformation.py
chonlei: Do you want something like "Jacobian of the transformation"?

pints/_mcmc/__init__.py
chonlei: What do you mean here?

pints/_mcmc/__init__.py
chonlei: Hmm yeah, but not quite doing the full check at the same time, it just need to handle those that's provided. If it's not provided, we can leave this to be handled later.

pints/_mcmc/__init__.py
chonlei: Hmm?

pints/_mcmc/__init__.py
MichaelClerx: If we're going to do it in the controller instead of the optimiser, I think it'd be less confusing for everybody if we did it all the time. Otherwise we're creating room for weird bugs and edge cases :-)

pints/_transformation.py
MichaelClerx: No I just want to docstring to define the jacobian. What are you taking the derivative of, and with respect to what? What are x and y in dx/dy?

pints/_mcmc/__init__.py
MichaelClerx: The controller checks if sensitivities are available when the method requires them.
This means the transformed objects should also have sensitivities available.
Didn't read the code in detail, so this might be automatic? But otherwise we'd have to update that check a bit to test that the provided transformation is able to do sensitivities

pints/_mcmc/__init__.py
MichaelClerx: OK, could you maybe run an O(10000) samples MCMC chain and do a log transform both ways, just to see if there's an appreciable difference?
We could always do a `many_to_model()` method or something in the transformation, with a slow default implementation in the base class?

pints/_transformation.py
MichaelClerx: Then just write that :-)

MichaelClerx: Then just write that :-)

pints/_transformation.py
MichaelClerx: Maybe "wrap_" :thinking:

MichaelClerx: Oh no this one's not wrapped :D

How about "convert_x" for all these methods?

MichaelClerx: Or "apply_to_", so it becomes logtransform.apply_to_log_pdf()". But I still favour convert_

pints/_transformation.py
chonlei: Ha, OK, well, Jacobian is _the_ matrix :smile:

pints/_mcmc/__init__.py
chonlei: >We could always do a many_to_model() method or something in the transformation, with a slow default implementation in the base class?
Yup, don't mind doing this one, if you think that makes sense too!

pints/_mcmc/__init__.py
chonlei: I think I tried doing that, but that was a bit annoying for multi- and single-chain methods... but I guess I can give it another go and move all the checks etc. from the optimiser/sampler to the controller.

chonlei: In fact, what was the reason of having that in the optimiser/sampler instead of the controller in the first place?!

pints/_transformation.py
```diff
@@ -0,0 +1,574 @@
...
+        logpdf = logpdf_nojac + log_jacobian_det
+
+        # Calculate the S1 change of variable. From Chain Rule Wikipedia:
+        # https://w.wiki/Us8
+        #
+        # For some transformation x = h(p) and its inverse p = g(x) applied
+        # to some function E
+        #
+        # dE/dx_i = \sum_l dE/dp_l * dp_l/dx_i  (d denotes partial derivative)
+        #
```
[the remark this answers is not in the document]
chonlei: Ha, good spot, thanks. That means we need to provide `log_jacobian_det_S1()` :roll_eyes:

pints/_transformation.py
DavAug: Haha yes "Transformation Jacobian" should probably be clear enough. Although I personally get sometimes confused whether the Jacobian is defined as dx/dy or dy/dx, because in theory both work equally fine. :D

pints/_transformation.py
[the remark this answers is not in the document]
DavAug: Looks like it. Do you think a `log_jacobian_det_S1()` could be generalised though?

pints/_transformation.py
chonlei: Yup, in fact, they just hold a inverse relation, according to the inverse function theorem (awesome!). But yes, we need _one_ of them but not the other(!) -- we need the dg^{-1}/y in your y = g(x) notation! I will add this in the docstring.

pints/_transformation.py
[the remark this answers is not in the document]
chonlei: What do you mean by "generalised" here? For each of the transformation function, we need to provide it manually, one by one (unfortunately). There's no general way to calculate it unless we have auto diff...!

examples/sampling/transformed-parameters.ipynb
```diff
@@ -0,0 +1,335 @@
...
+    "time_taken = timer.time()\n",
+    "\n",
+    "# Discard warm up\n",
+    "chains = chains[:, 2000:, :]\n",
+    "\n",
+    "# Look at distribution across all chains\n",
+    "pints.plot.pairwise(np.vstack(chains), kde=False)\n",
+    "\n",
+    "# Show graphs\n",
+    "plt.show()"
```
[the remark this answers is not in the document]
MichaelClerx: A lot of plot methods have an argument to pass parameter names in too

chonlei:
@MichaelClerx Re:Nested sampling, I am thinking to have this in a separate PR. As it seems to be a bit more trickier.

@ben18785 I probably have to talk to you about this...

Because we cannot sample from the transformed prior; we can sample only from the original prior and then transform the samples. Then the evidence Z = \int L(x) p(x) dx, where L(.) is the likelihood, p(.) is the prior, and x is the model parameter. Then if we apply a transformation of y = g(x), then Z = \int L(x(y)) p(x(y)) (dx/dy) dy.

My understanding of nested sampling, from a quick read [here](https://projecteuclid.org/download/pdf_1/euclid.ba/1340370944), we need to first sample from the prior. But we when do the transformation, we can no longer sample parameters y from the transformed prior p(y) = p(x(y)) dx/dy. What we can do is sample y from the _original_ prior p(x(y)), and the calculate the _transformed_ likelihood L(y) = L(x(y)) dx/dy...

Do you think there's an easy way of implementing this?

pints/_optimisers/__init__.py
chonlei: Doing `transform.convert_standard_deviation` and `transform.convert_covariance_matrix` now.

pints/_mcmc/__init__.py
chonlei: @MichaelClerx It seems to be a bit annoying to move everything down to the controller, because we have `MultiChainMCMC` and `SingleChainMCMC` that do slightly different things... And wouldn't that lose the checks if someone decided to use the ask and tell with `MCMCSampler` -- as well as the handling of sigma0 if not provided!

pints/_mcmc/__init__.py
chonlei: Hmm we don't check if sensitivities are available or not (by far I can see/search!). It only asks if the method requires sensitivity
```
        # Check if sensitivities are required
        self._needs_sensitivities = self._samplers[0].needs_sensitivities()
```
Then it went straight to ask for it
```
        # Choose method to evaluate                                             
        f = self._log_pdf                                                       
        if self._needs_sensitivities:                                           
            f = f.evaluateS1
```
So the `TransformLogPDF` that has `evaluateS1` should automatically deal with this.

pints/_mcmc/__init__.py
chonlei: I have got that tests already!! See somewhere [here](https://github.com/pints-team/pints/pull/1165/files#diff-cf96e760e788f41c818773f7d42e7f15R763). The assert for all larger than zero ensure that is transformed back to the model space. Now I have added this explanation somewhere in the test...
```
# log-transform of the parameter in [0.01, 0.02] will always be         
# negative values, so checking it larger than zero make sure it's       
# transformed back to the model space
```

chonlei: Not sure what do you want me to check near line 686, that just assign the parameters from `tell()` to `samples`, and it is 'detransformed' before it's saved to disk and assign to `self._samples` (in memory).

chonlei: I will mark this as resolved. If I've missed your point, you can reopen this again.

pints/_mcmc/__init__.py
chonlei: I will mark this as resolved too. Repoen this if you think I've got this wrong.

pints/_mcmc/__init__.py
chonlei: Done

pints/_mcmc/__init__.py
chonlei: Done

pints/_mcmc/__init__.py
chonlei: It's moved to Transform now as `convert_covariance_matrix`.

pints/_mcmc/__init__.py
chonlei: Done

pints/_mcmc/__init__.py
chonlei: Done

pints/_mcmc/__init__.py
chonlei: Done

pints/_optimisers/__init__.py
chonlei: Done

pints/_transformation.py
chonlei: Use `trans`. But keep `p` for model parameter and `q` as transformed search space parameter (adapting your comment below).

pints/_transformation.py
chonlei: Just write that now :-)

pints/_transformation.py
chonlei: Don't mind, so changed.

pints/_transformation.py
chonlei: Done

pints/_transformation.py
chonlei: Done

pints/_transformation.py
chonlei: Done.

pints/_transformation.py
chonlei: Updated.

pints/_transformation.py
chonlei: Updated docstring.

pints/_mcmc/__init__.py
chonlei: Done. It's called `multiple_to_model()` (instead of "many", so it's just anything more than one...).

pints/_optimisers/__init__.py
chonlei: Added some newlines around this. But I don't see how this can be done outside the except block (which includes KeyboardInterrupt!)...

pints/_transformation.py
chonlei: Changed to "convert_".

examples/sampling/transformed-parameters.ipynb
```diff
@@ -0,0 +1,335 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The MCMC samples in the above plot looks very similar to the one without parameter transformation when we had some good starting points in [here](./adaptive-covariance-haario-bardenet.ipynb), which is a good sign!\n",
```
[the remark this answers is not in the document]
chonlei: The reference works for me. It should work on GitHub too!

Ha, not sure if it was me. @MichaelClerx did you add that one? Anyway, I have slightly reworded it to make it a bit clearer that we are comparing with another example notebook.

examples/sampling/transformed-parameters.ipynb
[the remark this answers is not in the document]
chonlei: Thanks, added the parameter names to the plots.

pints/_mcmc/__init__.py
chonlei: We will keep those checks in the samplers as they are for now, before we come up a better way of doing this such as moving them into a utility method...

chonlei:
@ben18785, thanks for the suggestions -- happy to have the review in several parts, no worries!

Re:
>In the examples/optimisation/transformed-parameters-model-wrapper.ipynb notebook, can I just check that we do use a Jacobian here? I know we say that we should but I couldn't find where it was. If we don't use a Jacobian (because I know we want to encourage people into using our own transform functionality) then I think we need to really spell out that this is (technically) not right...potentially even not having the notebook without it;

Yeah, you are right that the Jacobian wasn't there -- so it's not correct if used as a log-likelihood/posterior/any PDF. The example uses the SumOfSquareError measure, so it's correct for that purpose (I think I need to add a sentence in the intro to highlight this).

@MichaelClerx suggested to remove this at some point too. When I was making the other one, in my mind, this example was kind of a "hacking with Pints" notebook... I am not sure about this now!

Should I take this out?!

docs/source/transformations.rst
```diff
@@ -0,0 +1,61 @@
...
+performance improvements for many methods.
+
+Example::
+
+    transform = pints.LogTransformation(n_parameters)
+    mcmc = pints.MCMCController(log_posterior, n_chains, x0, transform=transform)
+
+Overview:
+
+- :class:`Transformation`
```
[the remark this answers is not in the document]
chonlei: This is alphabetical, except the abstract class Transformation which I thought would make sense to go first...

pints/_mcmc/__init__.py
```diff
@@ -288,18 +288,51 @@ class MCMCController(object):
         in ``logpdf`` around the points in ``x0`` (the same ``sigma0`` is used
         for each point in ``x0``).
         Can be specified as a ``(d, d)`` matrix (where ``d`` is the dimension
-        of the parameterspace) or as a ``(d, )`` vector, in which case
+        of the parameter space) or as a ``(d, )`` vector, in which case
         ``diag(sigma0)`` will be used.
+    transform : pints.Transformation
+        A :class:`pints.Transformation` that transform the model parameter
```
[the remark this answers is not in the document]
chonlei: Done.

pints/_optimisers/__init__.py
```diff
@@ -301,13 +301,18 @@ class OptimisationController(object):
         this information.
     boundaries
         An optional set of boundaries on the parameter space.
+    transform
+        A :class:`pints.Transformation` that transform the model parameter
```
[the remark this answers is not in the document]
chonlei: Done

docs/source/transformations.rst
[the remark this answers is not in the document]
chonlei: OK have made it alphabetical now.

docs/source/transformations.rst
```diff
@@ -0,0 +1,61 @@
...
+- :class:`LogTransformation`
+- :class:`LogitTransformation`
+- :class:`RectangularBoundariesTransformation`
+- :class:`TransformedLogPDF`
+- :class:`TransformedLogPrior`
+- :class:`TransformedErrorMeasure`
+- :class:`TransformedBoundaries`
+
+
+.. autoclass:: Transformation
```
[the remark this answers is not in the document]
chonlei: Done.

pints/__init__.py
```diff
@@ -231,6 +233,23 @@ def version(formatted=False):
 from ._nested._ellipsoid import NestedEllipsoidSampler
 
 
+#
+# Transformation
+#
+from ._transformation import (
```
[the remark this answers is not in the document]
chonlei: Done

chonlei:
@MichaelClerx I think you wrote this bit in the Transformation docs
>To perform optimisation or sampling in a transformed space, users can choose to write their pints.ForwardModel in “search space” directly. But an alternative is to write the ForwardModel in model parameters, and pass a Transformation object to e.g. an OptimisationController or MCMCController.

If we say "users can write the model in search space (whatever it means to be 'nonphysical') directly", I think we also have to warm users that the posterior (or even the likelihood!) obtained in this form will be (strictly speaking) *incorrect* in the "model space" (those 'physical' parameter) because of the missing Jacobian!

examples/README.md
```diff
@@ -16,7 +16,9 @@ relevant code.
 
 
 ## Digging deeper
-- [Searching in a transformed parameter space](./optimisation/transformed-parameters.ipynb)
+- [Optimisation in a transformed parameter space - using a Transform object](./optimisation/transformed-parameters.ipynb)
```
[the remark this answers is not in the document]
chonlei: Glad that you like them! Even moving the "Full control with the ask-and-tell interface" up?! Thought we left that one later on purpose...

chonlei:
Just added an example showing why Jacobian adjustment is important [here](https://github.com/pints-team/pints/blob/4d74bd378ce124d5b1521c622d7fa5ea90be4bd7/examples/sampling/transformation-with-and-without-jacobian.ipynb), @ben18785 @MichaelClerx can you have a look? You might like it!

examples/README.md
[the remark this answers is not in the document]
ben18785: Errm, I'd just keep it alphabetical to make it more searchable. I suspect it's unlikely people will go through these sequentially.

chonlei:
Thanks @ben18785, it's a big relief that you've been through the Jacobian and their derivatives in Mathematica!

And sure, happy to add the scaling transformation, should be fairly straight forward as you said. Thought log or 'rectangular boundaries' transforms would be a bit more appropriate?

Thanks for the comments, I will work through them!

pints/_transformation.py
```diff
@@ -0,0 +1,846 @@
...
+        """
+        return TransformedErrorMeasure(error_measure, self)
+
+    def convert_boundaries(self, boundaries):
+        """
+        Returns a transformed boundaries class.
+        """
+        return TransformedBoundaries(boundaries, self)
+
+    def convert_covariance_matrix(self, C, q):
```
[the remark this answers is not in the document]
chonlei: These are used in the `_mcmc/__init__.py`. ``MCMCController`` takes a convariance matrix as an input argument, we need some way to transform that into the search space, so I made these approximations for that.

ben18785:
@chonlei Cool, thanks.

Re: the scaling transform, I guess it's slightly different to the log or rect boundaries transforms since it doesn't allow an edge. All it does is make sure the variables are on similar (both unconstrained) scales.

docs/source/transformations.rst
```diff
@@ -0,0 +1,61 @@
+***************
+Transformations
+***************
+
+.. currentmodule:: pints
+
+:class:`Transformation` objects provide methods to transform between different
+representations of a parameter space; for example from a "model space" where
+parameters have units and some physical counterpart to a "search space" where
+parameters are non-dimensionalised and less-recognisable to the modeller but
```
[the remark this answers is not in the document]
chonlei: @MichaelClerx any strong opinion/rewording suggestion on this? :smile:

pints/_transformation.py
```diff
@@ -0,0 +1,846 @@
...
+
+
+class RectangularBoundariesTransformation(Transformation):
+    r"""
+    A generalised version of logit transformation for the model parameters,
+    which transform an interval or ractangular boundaries :math:`[a, b)` to
+    all real number:
+
+    .. math::
+        q = f(p) = \log(p - a) - \log(b - p),
```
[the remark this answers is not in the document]
chonlei: Agreed, they are equivalent, maybe I can add that and make it something like `q = f(p) = logit((p - a) / (b - a)) = ...`?

pints/_transformation.py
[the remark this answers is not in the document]
ben18785: Gotcha

pints/_transformation.py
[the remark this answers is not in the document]
ben18785: Yep, sounds good!

chonlei:
Have added an abstract class `ElementWiseTransformation` that effectively separates out the transformation of the form `y_i = f_i(x_i)` and some fancier(...?!) 'general' transformation `y_i = f_i(x_i, x_j, x_k, ...)`. So far all of our transformation classes are element-wise, so when we compose them, we can save a lot of matrix calculation(!); whereas for general transformation, when we compose them, we need to do quite a few matrix calculations... So in order to keep the advantage of element-wise transformations when composing them, I created `ComposedElementWiseTransformation` that particularly deals with element-wise ones (i.e. those with simple diagonal matrices for Jacobians), along with ``ComposedTransformation` that can take any general transformation but need to do with a few matrix calculations (for any non-diagonal Jacobians).

@DavAug @MichaelClerx @martinjrobins Can you have a look? Thanks!

examples/README.md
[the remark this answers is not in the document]
chonlei: OK, done.

docs/source/transformations.rst
```diff
@@ -0,0 +1,61 @@
...
+easier to deal with mathematically.
```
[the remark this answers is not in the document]
chonlei: Done.

docs/source/transformations.rst
```diff
@@ -0,0 +1,61 @@
...
+
+To perform optimisation or sampling in a transformed space, users can choose to
```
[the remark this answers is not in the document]
chonlei: Agreed! Added an explanation to that.

pints/_transformation.py
```diff
@@ -0,0 +1,846 @@
...
+    """
+    def __init__(self, log_prior, transform):
+        super(TransformedLogPrior, self).__init__(log_prior, transform)
+
+    def sample(self, n):
+        """
+        See :meth:`pints.LogPrior.sample()`.
+
+        *Note that this does not sample from the transformed log-prior but
+        simply transform the samples from the original log-prior.*
```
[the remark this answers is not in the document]
chonlei: Fixed.

pints/_transformation.py
```diff
@@ -0,0 +1,846 @@
...
+    The Jacobian adjustment of the logit transformation is given by
+
+    .. math::
+        |\frac{d}{dq} \text{logit}^{-1}(q)| = \text{logit}^{-1}(q) \times
+        (1 - \text{logit}^{-1}(q)).
+
+    The first order derivative of the log determinant of the Jacobian is
+
+    .. math::
+        \frac{d}{dq} \log(|J(q)|) = 2 \times \exp(-q) \times
```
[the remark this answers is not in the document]
chonlei: Thanks! Think I will keep this for now, done the maths and more derivatives from this. We could change it if we find tanh is better performance-wise.

pints/_transformation.py
```diff
@@ -0,0 +1,846 @@
...
+    def to_search(self, p):
+        """ See :meth:`Transformation.to_search()`. """
+        p = pints.vector(p)
+        return logit(p)
+
+
+class RectangularBoundariesTransformation(Transformation):
+    r"""
+    A generalised version of logit transformation for the model parameters,
+    which transform an interval or ractangular boundaries :math:`[a, b)` to
```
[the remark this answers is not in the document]
chonlei: Fixed.

pints/_transformation.py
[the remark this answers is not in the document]
chonlei: Done.

pints/_transformation.py
```diff
@@ -0,0 +1,846 @@
...
+    The log-determinant of the Jacobian matrix is given by
+
+    .. math::
+        \log|\frac{d}{dq} f^{-1}(q)| = \sum_i \left( \log(b_i - a_i) -
+            2 \times \log(1 + \exp(-q_i)) - q_i \right)
+
+    The first order derivative of the log determinant of the Jacobian is
+
+    .. math::
+        \frac{d}{dq} \log(|J(q)|) = 2 \times \exp(-q) \times
```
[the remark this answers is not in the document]
chonlei: Again, I will keep it as it is for now...

pints/tests/test_transformation.py
```diff
@@ -0,0 +1,418 @@
...
+
+        p = [0.1, 1., 10., 999.]
+        x = [-2.3025850929940455, 0., 2.3025850929940459, 6.9067547786485539]
+        j = np.diag(p)
+        log_j_det = np.sum(x)
+
+        # Test inverse transform
+        self.assertTrue(np.allclose(t.to_model(x), p))
+
+        # Test many invesre transform
```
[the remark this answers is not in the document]
chonlei: Fixed.

chonlei:
@ben18785 I have added `ScalingTransformation` as requested. I thought 'scaling * parameter' may be better than 'parameter / scaling'... Hope that's OK.

So instead of inputting with some typical values val1 and val2 of the parameters, users input 1/val1 and 1/val2 to move the parameters to the unit scale with this `ScalingTransformation`.

Maybe have a quick look, see if that's what you want.

pints/_transformation.py
```diff
@@ -0,0 +1,787 @@
...
+        Using the property that :math:`J^{-1} = \frac{dg}{dp}`, from the
+        inverse function theorem, i.e. the matrix inverse of the Jacobian
+        matrix of an invertible function is the Jacobian matrix of the inverse
+        function.
+
+        If :math:`C(p) = var(p) \cdot I`, where :math:`var(.)` is the variance
+        and :math:`I` is the identity matrix, then
+
+        .. math::
+            C(q) = var(q) \cdot I = diag(J^{-1}(q))^2 var(p) I
```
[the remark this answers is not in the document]
chonlei: This expression is indeed only intended to be used for standard deviation.

pints/_transformation.py
```diff
@@ -0,0 +1,787 @@
...
+
+        Therefore the standard deviation ``s`` can be transformed using
+
+        s(q) = diag(J^{-1}(q)) \times s(p)
```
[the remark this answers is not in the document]
chonlei: Yeah, have assumed element-wise transformation when I was doing this... Forgot to change it back when we decided to have Transformation and ElementWiseTransformation. Thanks.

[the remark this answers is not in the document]
chonlei: Done.

pints/_transformation.py
```diff
@@ -0,0 +1,787 @@
...
+        Converts a convariance matrix ``C`` from the model space to the search
+        space around a parameter vector ``q`` provided in the search space.
+
+        The transformation is performed using a linear approximation [1]_ with
+        the Jacobian :math:`J`:
+
+        .. math::
+
+            C(q) &= \frac{dg(p)}{dp} C(p) (\frac{dg(p)}{dp})^T \\
+                 &= J^{-1}(q) C(p) (J^{-1}(q))^T.
```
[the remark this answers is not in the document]
chonlei: Added.

pints/_transformation.py
```diff
@@ -0,0 +1,787 @@
...
+        \log(\pi(q)) = \log(\pi(f^{-1}(q))) + \log(|det(J(f^{-1}(q)))|).
+
+    For the first order sensitivity, the transformation is done using
+
+    .. math::
+        \frac{\partial \log(\pi(q))}{\partial q_i} =
+            \frac{\partial \log(|det(J)|)}{\partial q_i}
+            + \frac{\partial \log(\pi(f^{-1}(q)))}{\partial q_i}.
+
+    The second term can be calculated using the calculated using the chain rule
```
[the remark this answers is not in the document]
chonlei: Fixed.

pints/_transformation.py
```diff
@@ -0,0 +1,787 @@
...
+        the model space to the search space around a parameter vector ``q``
+        provided in the search space.
+
+        The transformation is performed using a linear approximation [1]_ with
+        the Jacobian :math:`J`:
+
+        .. math::
+
+            C(q) &= \frac{dg(p)}{dp} C(p) (\frac{dg(p)}{dp})^T \\
+                 &= J^{-1}(q) C(p) (J^{-1}(q))^T.
```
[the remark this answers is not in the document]
chonlei: Added.

pints/_transformation.py
```diff
@@ -0,0 +1,787 @@
...
+    Hence
+
+    .. math::
+        \log(\pi(q)) = \log(\pi(f^{-1}(q))) + \log(|det(J(f^{-1}(q)))|).
+
+    For the first order sensitivity, the transformation is done using
+
+    .. math::
+        \frac{\partial \log(\pi(q))}{\partial q_i} =
+            \frac{\partial \log(|det(J)|)}{\partial q_i}
```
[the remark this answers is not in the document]
chonlei: Fair enough, have swapped them. Thanks!

pints/tests/test_transformation.py
```diff
@@ -0,0 +1,610 @@
...
+        # Test log-Jacobian determinant
+        self.assertAlmostEqual(t.log_jacobian_det(x),
+                               t_elem.log_jacobian_det(x))
+        # Test log-Jacobian determinant derivatives
+        _, t_deriv = t.log_jacobian_det_S1(x)
+        _, t_elem_deriv = t_elem.log_jacobian_det_S1(x)
+        self.assertTrue(np.allclose(t_deriv, t_elem_deriv))
+
+    def test_composed_elementwise_transform(self):
+        # Test ComposedElementWiseTransformation class
```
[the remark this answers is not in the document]
chonlei: Split all.

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+        p = pints.vector(p)
+        """ See :meth:`Transformation.to_search()`. """
+        return np.log(p - self._a) - np.log(self._b - p)
+
+    def _softplus(self, q):
+        """ Returns the softplus function. """
+        return np.log(1. + np.exp(q))
+
+
+class ScalingTransformation(ElementWiseTransformation):
```
ben18785, line 821: Thanks @chonlei -- looks great! This is exactly the transformation I was wanting to have in there...

chonlei: Great!

docs/source/transformations.rst
```diff
@@ -0,0 +1,61 @@
+***************
+Transformations
+***************
+
+.. currentmodule:: pints
+
+:class:`Transformation` objects provide methods to transform between different
+representations of a parameter space; for example from a "model space" where
```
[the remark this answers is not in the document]
chonlei: Done.

chonlei:
I will leave a few more potential useful transformation function as a separate PR to work on #1187.

pints/_mcmc/__init__.py
```diff
@@ -774,14 +816,25 @@ def run(self):
...
         # Store evaluations in memory
         if self._evaluations_in_memory:
             self._evaluations = evaluations
 
+        if self._chains_in_memory:
+
+            # Inverse transform to model space if transform is provided
+            if self._transform:
+                n_c, n_s, n_p = samples.shape
+                samples_user = np.zeros((n_c, n_s, n_p))
```
MichaelClerx, line 828: I _think_ that the code in the constructor could actually cope with the situation where n_p isn't the same in the world and search space. At least, I was about to re-order some checks until I figured this was the case, and that you might have done it on purpose.

If so... or just if we want to have the option of having an n_p change, then this code should use `np.zeros((n_c, n_s, original-log-pdf.n_parameters()))` or something similar.

If we think changing `n_p` is a terrible idea, then maybe this code can be optimised by modifying the samples array during the loop, instead of creating a full copy? (so e.g. `samples[c, :, :] = self._transform...`)

MichaelClerx: Most memory-efficient solution would still be to just transform before sticking them in `samples` in the first place!
Then the transformation objects become simpler too? (No need for `multiple_to_model`)

MichaelClerx: > What was the reasoning for transforming the whole chain at the end of the sampling routine, rather than when each sample comes in?
> I know we discussed this before at some point...

> chon  2:33 PM
> Yes haha! Because I want “whatever happens within the sampling route stay in the transformed/search space”. More for in case e.g. later if someone needs a method that takes the previous two steps within the sampled chain, and wonder if it’s transformed or detransformed — suspect quite error prone.  (edited) 

> michael  2:35 PM
> Hmmmm, you can't take 2 samples without losing the markov property
> But also we'll never have two-way communication between sampler and controller! sampler doesn't know about controller
> So if you did need 2 previous samples in a sampler, you'd have to store them yourself
> Same way that samplers already maintain a current state and logpdf, instead of requesting that from the controller

MichaelClerx: (with respect to the possibility of changing n_p, that would mean the preallocation would have to use n_p of "world" samples, I guess)

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+        """
+        raise NotImplementedError
+
+    def log_jacobian_det(self, q):
+        """
+        Returns the logarithm of the absolute value of the determinant of the
+        Jacobian matrix of the transformation :meth:`Transformation.jacobian`
+        calculated at the parameter vector ``q`` in the search space.
+
+        *This is an optional method.* It is needed when transformation is
```
MichaelClerx, line 177: Is the implementation below always suitable? Maybe the "optional" text should mention its implemented?
Minor detail :D

MichaelClerx: Something like *This method only works if the optional method X() is implemented"

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+        # The second term is given by Eq. 46 in the Matrix Cookbook (2012)
+        # http://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf
+        #
+        # d/dq det(J(q)) = det(J) Tr(J^{-1} d/dq J)
+        #
+        # Therefore
+        #
+        # d/dq log(|det(J(q))|) = Tr(J^{-1} d/dq J)
+        #
+        q = pints.vector(q)
```
MichaelClerx, line 213: Same here. And the default implementation could be explained in the docstring?

pints/_transformation.py
MichaelClerx: And the default implementation could be explained in the docstring?

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+        out_S1 = np.zeros(q.shape)
+        for i, jac_S1_i in enumerate(jac_S1):
+            out_S1[i] = np.trace(np.matmul(np.linalg.pinv(jac), jac_S1_i))
+        return self.log_jacobian_det(q), out_S1
+
+    def multiple_to_model(self, qs):
+        """
+        Transforms a series of parameter vectors ``qs`` from the search space
+        to the model space. ``qs`` must be provided in the shape
+        (``n_vectors``, ``n_parameters``).
```
MichaelClerx, line 224: the shape is a tuple, so 
```
``(n_vectors, n_parameters)``
```

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+
+    where :math:`x_i` denotes the :math:`i^{\text{th}}` element of the vector
+    :math:`\boldsymbol{x}`.
+    """
+
+
+class ComposedTransformation(Transformation):
+    r"""
+    N-dimensional :class:`Transformation` composed of one or more other
+    :math:`N_i`-dimensional ``Transformation``, such that
```
MichaelClerx, line 274: ``Transformations``

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+        """
+        Transforms a parameter vector ``p`` from the model space to the search
+        space.
+        """
+        raise NotImplementedError
+
+
+class ElementWiseTransformation(Transformation):
+    r"""
+    Abstract base class for objects that provide *element-wise* transformations
```
MichaelClerx, line 256: Maybe mention in docs that this is a subclass of Transformation? (Not a separate thing)

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+    and a parameter vector :math:`\boldsymbol{q}` in the search space, the
+    transformation :math:`\boldsymbol{f}` relating the two vectors takes the
+    form
+
+    .. math::
+        q_i = f_i(p_i),
+
+    where :math:`x_i` denotes the :math:`i^{\text{th}}` element of the vector
+    :math:`\boldsymbol{x}`.
+    """
```
MichaelClerx, line 268: Guess this can implement n_parameters() then?

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+            out_S1[i] = np.trace(np.matmul(np.linalg.pinv(jac), jac_S1_i))
+        return self.log_jacobian_det(q), out_S1
+
+    def multiple_to_model(self, qs):
+        """
+        Transforms a series of parameter vectors ``qs`` from the search space
+        to the model space. ``qs`` must be provided in the shape
+        (``n_vectors``, ``n_parameters``).
+        """
+        qs = np.asarray(qs)
```
MichaelClerx, line 226: If we're going to use this a lot (and it doesn't disappear, see comments above), then maybe do a quick benchmark to see if it's faster as a list comprehension (with an array wrapped around that)

pints/_transformation.py
MichaelClerx: And can any of the other implemented methods be implemented faster/easier in the 1d case?

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+                #     [ 0   0   0 ]
+                #
+                o = np.zeros(matrix_shape)
+                o[lo:hi, lo:hi] = jac_S1_i[:, :]
+                output_S1[lo + i, :, :] = o
+        return self.jacobian(q), output_S1
+
+    def multiple_to_model(self, qs):
+        """ See :meth:`Transformation.multiple_to_model()`. """
+        qs = np.asarray(qs)
```
MichaelClerx, line 361: I don't think we can do better but I'm not loving this implementation!
You've got your original array, then a copy filled with zeros, and then copies of each part are made by the submethods!
Another reason I'd rather avoid this method 😅

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+            hi += transform.n_parameters()
+            output[lo:hi] = np.asarray(transform.to_search(p[lo:hi]))
+        return output
+
+    def n_parameters(self):
+        """ See :meth:`Transformation.n_parameters()`. """
+        return self._n_parameters
+
+
+class ComposedElementWiseTransformation(ElementWiseTransformation,
```
MichaelClerx, line 398: Great stuff!

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+
+    def to_search(self, p):
+        """ See :meth:`Transformation.to_search()`. """
+        p = pints.vector(p)
+        return np.log(p)
+
+
+class RectangularBoundariesTransformation(ElementWiseTransformation):
+    r"""
+    A generalised version of logit transformation for the model parameters,
```
MichaelClerx, line 697: of the logit transformation

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+            hi += transform.n_parameters()
+            j, j_S1 = transform.log_jacobian_det_S1(q[lo:hi])
+            output += j
+            output_S1[lo:hi] = np.asarray(j_S1)
+        return output, output_S1
+
+
+class IdentityTransformation(ElementWiseTransformation):
+    """
+    Identity transformation does nothing to the input parameters, i.e. the
```
MichaelClerx, line 480: Maybe something like ":class`Transformation` that returns the untransformed parameters" ?

Don't think the jacobian line is needed

pints/_transformation.py
```diff
@@ -0,0 +1,1120 @@
...
+        return pints.vector(q)
+
+    def to_search(self, p):
+        """ See :meth:`Transformation.to_search()`. """
+        return pints.vector(p)
+
+
+class LogitTransformation(ElementWiseTransformation):
+    r"""
+    Logit (or log-odds) transformation of the model parameters:
```
MichaelClerx, line 530: Wondering if this, and all the following, should be rewritten to have the standard 1-sentence, space, 2nd paragraph format for docstrings?
The "math" part won't show up in hints in IDEs etc. So it might look like "Log (or log-odds) transformation of the model parameters:" there? Maybe just a full stop is better, and then the next paragraph can be "The transformation is given by" or something?

pints/tests/test_opt_optimisation_controller.py
```diff
@@ -44,14 +44,47 @@ def test_optimise(self):
         self.assertEqual(x.shape, (2, ))
         self.assertTrue(f < 1e-6)
 
+    def test_transform(self):
+        # Test optimisation with parameter transformation.
+
+        # Test with LogPDF
```
MichaelClerx, line 50: Would these fail without a transformation? That'd be quite good, from a functional testing point of view :D 
(For example nelder-mead on logistic fails without a transform, and is very fast so good for testing)

pints/tests/test_transformation.py
```diff
@@ -0,0 +1,734 @@
...
+# copyright notice and full license details.
+#
+from __future__ import division
+import unittest
+import pints
+import pints.toy
+import numpy as np
+
+
+class TestTransformation(pints.Transformation):
```
MichaelClerx, line 16: An obvious extra test in all these methods could be `assertEqual(p, detransform(transform(p)))` ?
It _almost_ follows from what we have now, but I think having this too would catch any issues where the shape slightly changes or anything like that?

pints/_mcmc/__init__.py
chonlei: Moved transformation to before assigning values to `samples` now. And added a note to samples saying something like "we store only the detransformed parameters if transform is provided", so if anyone later have a look at `samples`, hopefully they will see that message and know that they are dealing with detransformed parameters.

I will also remove `multiple_to_model` soon.

pints/_transformation.py
chonlei: Done

pints/_transformation.py
chonlei: Fixed

pints/_transformation.py
chonlei: Fixed

pints/_mcmc/__init__.py
chonlei: Removed `multiple_to_model` methods.

pints/_transformation.py
chonlei: Removed this method.

pints/_transformation.py
chonlei: Changed it to
```
        The default implementation numerically calculates the determinant of    
        the full matrix. If there is an analytic expression for the specific    
        transformation, a reimplementation of this method may be preferred.     
                                                                                
        *This is an optional method that only works if the optional method      
        :meth:`Transformation.jacobian` is implemented.* It is needed when      
        transformation is performed on :class:`LogPDF` and/or that requires     
        ``evaluateS1()``; e.g. not necessary if it's used for                   
        :class:`ErrorMeasure` without :meth:`ErrorMeasure.evaluateS1()`.
```
Does it explain better? Any suggestion?

pints/_transformation.py
chonlei: Added explanation
```
        The absolute value of the determinant of the Jacobian is provided by    
        :meth:`Transformation.log_jacobian_det`. The default implementation     
        calculates the derivatives of the log-determinant using [1]_            
                                                                                
        .. math::                                                               
            \frac{d}{dq} \log(|det(\mathbf{J})|) =                              
                trace(\mathbf{J}^{-1} \frac{d}{dq} \mathbf{J}),                 
                                                                                
        where the derivative of the Jacobian matrix is provided by              
        :meth:`Transformation.jacobian_S1` and the matrix inversion is          
        numerically calculated. If there is an analytic expression for the      
        specific transformation, a reimplementation of this method may be       
        preferred.
```
Any suggestion?

pints/_transformation.py
chonlei: Done

pints/_transformation.py
chonlei: What do you mean by "can implement n_parameters()"? n_parameters is specified in each class that extends this. This is not necessary to be 1d.

pints/_transformation.py
chonlei: Removed this method.

pints/_transformation.py
chonlei: Updated docs. But I kept the jacobian line, thought in case anyone wonders...

pints/_transformation.py
chonlei: Updated.

pints/tests/test_opt_optimisation_controller.py
chonlei: I would leave these out for another PR! :sweat_smile:

chonlei: Created a ticket for that #1200

pints/tests/test_transformation.py
chonlei: Added.

pints/_transformation.py
MichaelClerx: What do you mean by "element-wise" if this is not for a single parameter (element) ?

MichaelClerx: If this can work on any number of parameters, then I don't understand the need to define it. Surely you can compose transformations of any other transformations, including already composed ones (nesting) ?

MichaelClerx (changes requested):
Just a question about the element-wise transformation class and then we're done!

pints/_transformation.py
chonlei: Hmm, I didn't explain that very well. I will try make it a bit more explicit (longer) here, and see what you think first.

Element-wise means the transformation applies to each _element_ of the parameter vector _independently_ (hence the name element-wise). So if we have a transformation `g`, then it does `q_i = g(p_i)` for every element `i` in the parameter vector. It is different from the general transformation in the sense that the transformation could do something like q1 = p1 + p2 and q2 = p1 * p2, which is a valid transformation but not element-wise. That is the whole idea of separating element-wise and the general one. This is done similarly in PyMC3!

Why not just for one element? Because often we would say just log transform the full parameter vector. Then it can be done simply with `LogTransformation(n_parameters=123)` rather than composing one transformation at a time...!

MichaelClerx: Right. It applies _mathematically_ to each element independently, but in practice it does n of them in one go :D 

Where do we use the independent/mixed difference in our code?

MichaelClerx: Is it between the ComposedTransformation and the ComposedElementWiseTransformations ? Probably best to cross-reference those then in their docstrings! So the composed one should say, if you're using all element-wise ones then using the other class if faster.

(Looking at them, they should also use the 1-line+space+paragraph docstring format)

Alternatively, you could have element_wise() as a method (or a private property), and a single ComposedTransformation that decides at construction time whether it is element wise or not, and then uses the appropriate calculations?

chonlei: Ha, if "mathematically" helps, yeah. :sweat_smile: 

No mixed one so far, all element-wise. But in case if anyone or later we want to have a mixed one.

chonlei: OK, after some more discussion with @MichaelClerx, I will move "element-wise" from subclass to a class property. So each Transformation class need one of
```
    def elementwise(self):
        """ Returns True if it is an element-wise transformation. """
        return NotImplementedError
```
And for the ComposedTransformatioin, which needs both, I will do something like
```
    def jacobian(self, q):                                                      
        """ See :meth:`Transformation.jacobian()`. """                          
        if self._elementwise:                                                   
            return self._elementwise_jacobian(q)                                
        else:                                                                   
            return self._general_jacobian(q)
```

pints/_transformation.py
```diff
@@ -264,24 +264,22 @@ def to_search(self, p):
...
-    Abstract base class for objects that provide *element-wise* transformations
-    between two parameter spaces: the model parameter space and a search space.
-    That is, for a parameter vector :math:`\boldsymbol{p}` in the model space
-    and a parameter vector :math:`\boldsymbol{q}` in the search space, the
-    transformation :math:`\boldsymbol{f}` relating the two vectors takes the
-    form
-
-    .. math::
-        q_i = f_i(p_i),
+        Element-wise transformation is a transformation :math:`\boldsymbol{f}`
```
MichaelClerx, line 271: how about "An element-wise transformation is a transformation f that can be carried out element by element: EQUATION (as opposed to a transformation in which multiple elements are combined to create the transformed elements"

MichaelClerx (approved):
Love it!

Suggestion for a docstring fix. Happy for this to go in!

MichaelClerx:
@chonlei Once you've hit the merge button, can you add a few lines to the changelog in https://github.com/pints-team/pints/pull/1204 please?

pints/_transformation.py
chonlei: Sounds good, updated!

PR #1166, opened 2020-06-22
Testing to see if something has changed with coverage of __init__

MichaelClerx:
See #1165


chonlei:
You probably meant #1165, but yeah, not sure why!

PR #1175, opened 2020-07-05
statsmodels: ARIMAX and state space models interface notebook

ben18785:
Shows how to wrap statsmodels ARIMAX and state space models in Pints log-likelihoods, then perform inference on them. The "local level" state space model I think is particularly interesting: I can imagine this being applicable to quite a few ODE models.


ben18785:
@chonlei @martinjrobins Can you have a look at the notebook and let me know your thoughts? Also, not sure I've done the yaml file correctly...

martinjrobins:
Thanks @ben18785, yml file looks ok, just have to make sure it runs ok once this is merged. Couple of things regarding the notebook:
- I had to squint at the first __call__ function to work out what was going on with all the parameter vector manipulations, as it is not 100% clear. Why are you converting it to a list? Can you add a few more explanatory comments?
- You state that the sigma values don't match due to a different definition, can we instead provide a consistent definition of sigma?
- you seem to swap between epsilon and mu for the noise term in the local level state-space model

chonlei:
Thanks @ben18785 looks really cool! This is definitely a very useful example for other users -- in fact we've used statsmodel this way (wrapping around for Pints) in our previous [discrepancy paper for hERG model](https://doi.org/10.1098/rsta.2019.0349)!

ARIMA
---------
Agree with Martin that it might be easier to just use `np.append()` to stack parameters as numpy array instead of list?

Would have thought it'd be cleaner to have the ARIMA pre-initialised in the `__init__` and then change the `exog` to the new model simulation (i.e. your `sol`) by `self.arima_model.exog = sol[:, None]` -- that's just how we did it...

And like Martin said, can we explain the difference in the sigma definition? Or better, if we could convert that some way so they are comparable. Although they are not very different...

Another thing probably need explaining is the difference between what you show in the example and their `ARIMA(...).fit()` function. Theirs is a simple optimisation and yours is a Bayesian approach, and that they could use a larger class of optimisation methods e.g. CMA-ES, etc.

Can you change the parameter 1, 2, 3, ... in the plot to the parameter name using the `parameter_names` argument in the `pints.plot` functions?

State space
---------------
The symbol for epsilon = mu = nu!

Maybe like the "ARIMA error" title, do one for the "State space"? And make those "Fitting ..." title to a lower level (with one more #)? So that make it very clear that there's (only!) two parts, one is ARIMA and one is state space.

Similar to the ARIMA, probably needs a bit more explanation in the likelihood wrapper...

Maybe plot the true parameters in the trace plot? Had to look into the code to find where those true parameters were defined...

ben18785:
@martinjrobins @chonlei thanks very much. I have incorporated most of your comments now:

- I've split the ARIMA and state space examples into separate notebooks;
- Added more documentation to the call functions;
- Changed the sigma value in the ARMA example, so that it's consistent and comparable across the notebook;
- Added param names and actual values to plots.

Things I was less sure about were the mu \ nu \ eta comment? Can you point me to exactly (in the new state space notebook) where the confusion arises?

@chonlei One thing I didn't do was your instantiation of the ARIMA model. Since we wouldn't be calling the model itself, I wasn't sure on the necessity to do so?

chonlei:
Hi @ben18785, thanks! Ha interesting, looking at the raw file, seems like you've got the symbols right:
https://github.com/pints-team/pints/pull/1175/files#diff-87bfa7d0ed48e0786b8bbfee6d39a5ceR31-R44.
But looking at the rendered notebook on GitHub, it seems like `\nu` is not defined anywhere...
image
Missing the first equation...!

And if you want, you could use `parameter_names=[r'$r$', r'$k$', r'$\sigma^2_{\nu}$', r'\sigma^2_{\eta}'],` and `parameter_names=[r'$r$', r'$k$', r'$\rho$', r'$\phi$', r'$\sigma$']` as the argument for the plots, so that they could render the symbols (I think it'd be a bit easier to read...).

And I think the `m` in
```
        # p AR params; q MA params
        m = (self._p + self._q + 1) * self._no
```
is a constant, so could move to `__init__`?

Re: instantiation, I just didn't want to 're-instantiate' the statemodel every time the likelihood is called (which is called many times!). Maybe we don't want to make it looks like we are hacking around other's packages, so I don't mind leave it as it is.

.github/workflows/notebook-interfaces-test.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          python --version
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install .
+          python -m pip install .[dev]
+
+      - name: run jupyter notebooks
+        run: |
+          python -m pip install statsmodels
+          python run-tests.py --books
```
martinjrobins, line 33: this will run all the notebooks right? I think you want to just run the interface ones

ben18785: @chonlei Thanks -- have made those changes.

ben18785: I didn't make the change to the parameter names in tables because the latex didn't render...

ben18785:
@chonlei Thanks -- have made those changes (apart from to the parameter names in tables because the latex didn't render for me).

ben18785:
@martinjrobins Happy to do so, but no idea how to do this, I'm afraid.

chonlei:
Thanks @ben18785 it looks great. I aligned the equations in the state-space example which makes them reappear in the GitHub rendered notebook.

.github/workflows/notebook-interfaces-test.yml
chonlei: Hmm, also, shouldn't we take out the interface ones from the --books? Afraid this will break the `.github/workflows/notebook-test.yml` test...

chonlei:
@ben18785 I have changed the yml tests for you, hope you don't mind! So now the `--books` ignores the interfaces as they require extra dependence. And added a new option `--interfaces` used in the new yml file that you created, which test only the interfaces notebooks.

@martinjrobins Can you have a look see if it looks about right?

chonlei:
Ha, random failing of the log test in `test_opt_optimisation_controller.TestOptimisationController`
```
======================================================================
FAIL: test_logging (test_opt_optimisation_controller.TestOptimisationController)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/pints/pints/pints/tests/test_opt_optimisation_controller.py", line 94, in test_logging
    self.assertEqual(log_should_be, c.text())
AssertionError: 'Maxi[333 chars]0:00.0\n10    60    -4.140462   0:00.0\nHaltin[43 chars]d.\n' != 'Maxi[333 chars]0:00.1\n10    60    -4.140462   0:00.1\nHaltin[43 chars]d.\n'
Diff is 642 characters long. Set self.maxDiff to None to see it.

----------------------------------------------------------------------
```
Definitely not happening on my local tests (and didn't change anything in the unit test)! @MichaelClerx Is this a known issue?!

.github/workflows/notebook-interfaces-test.yml
chonlei: Should be fixed now.

PR #1180, opened 2020-07-14
I1179 Improve multiplicative gaussian loglikelihood code and docs

### LINKED ISSUE #1179: Bug: MultiplicativeGaussianLogLikelihood
I believe that the computation of the log-likelihood is not quite correct for MultiOutputProblems. The constant term is missing a factor of `n_outputs`, see GaussianLogLikelihood for details.

Fortunately it doesn't make a difference for the inference, but may be nice to correct regardless.

DavAug:
Closes #1179 .

I fixed the issue reported in the issue above and changed the docstring accordingly.

### DIFF
--- a/pints/_log_likelihoods.py
+++ b/pints/_log_likelihoods.py
@@ -564,8 +564,7 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
     r"""
     Calculates the log-likelihood for a time-series model assuming a
     heteroscedastic Gaussian error of the model predictions
-    :math:`f(t, \theta )`. The standard deviation of a heteroscedascic
-    Gaussian error scales with the model predictions.
+    :math:`f(t, \theta )`.
 
     This likelihood introduces two new scalar parameters for each dimension of
     the model output: an exponential power :math:`\eta` and a scale
--- a/pints/_log_likelihoods.py
+++ b/pints/_log_likelihoods.py
@@ -562,34 +562,37 @@ class KnownNoiseLogLikelihood(GaussianKnownSigmaLogLikelihood):
 
 class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
     r"""
-    Calculates a log-likelihood assuming heteroscedastic Gaussian errors, with
-    the magnitude of the error variance scaling with the problem output.
+    Calculates the log-likelihood for a time-series model assuming a
+    heteroscedastic Gaussian error of the model predictions
+    :math:`f(t, \theta )`. The standard deviation of a heteroscedascic
+    Gaussian error scales with the model predictions.
 
-    For each output in the problem, this likelihood introduces two new scalar
-    parameters: an exponential power :math:`\eta` and a scale :math:`\sigma`.
+    This likelihood introduces two new scalar parameters for each dimension of
+    the model output: an exponential power :math:`\eta` and a scale
+    :math:`\sigma`.
 
-    This likelihood is applicable to a model given by
+    A heteroscedascic Gaussian noise model assumes that the observable
+    :math:`X` is Gaussian distributed around the model predictions
+    :math:`f(t, \theta )` with a standard deviation that scales with
+    :math:`f(t, \theta )`
 
     .. math::
-        X(t) = f(t; \theta) + f(t; \theta)^\eta v(t)
+        X(t) = f(t, \theta) + \sigma f(t, \theta)^\eta v(t)
 
-    where v(t) is iid Gaussian:
+    where :math:`v(t)` is a standard i.i.d. Gaussian random variable
 
     .. math::
-        v(t) \sim \text{ iid } N(0, \sigma)
+        v(t) \sim \mathcal{N}(0, 1).
 
-    Note that the scalar parameter :math:`\eta` controls the exponential
-    dependence of the noise on the function output, while the scalar parameter
-    :math:`\sigma` provides a baseline level of the noise standard deviation.
-    This model leads to a log likelihood of
+    This model leads to a log likelihood of the model parameters of
 
     .. math::
         \log{L(\theta, \eta , \sigma | X^{\text{obs}})} =
             -\frac{n_t}{2} \log{2 \pi}
             -\sum_{i=1}^{n_t}{\log{f(t_i, \theta)^\eta \sigma}}
-            -\frac{1}{2}\sum_{i=1}^{n_t}
-                \frac{(X^{\text{obs}}_{i} - f(t_i, \theta))^2}
-                {(f(t_i, \theta)^\eta \sigma)^2}
+            -\frac{1}{2}\sum_{i=1}^{n_t}\left(
+                \frac{X^{\text{obs}}_{i} - f(t_i, \theta)}
+                {f(t_i, \theta)^\eta \sigma}\right) ^2,
 
     where :math:`n_t` is the number of time points in the series, and
     :math:`X^{\text{obs}}_{i}` the measurement at time :math:`t_i`.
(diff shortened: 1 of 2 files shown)


rcw5890:
The docstring update ("For a system with :math:`n_o` outputs, this becomes...") is definitely very helpful.

I may be wrong here. But in the code, wasn't numpy broadcasting already giving us the correct number of prefactors (one for each output)? In any case, the old behaviour of MultiplicativeLogLikelihood looks the same as GaussianLogLikelihood, when you set eta to zero:

```>>> import numpy as np
>>> import pints
>>> import pints.toy
>>> np.random.seed(1000)
>>> t = np.linspace(0, 10, 100)
>>> m = pints.toy.ConstantModel(3)
>>> values = m.simulate([3, 5, 7], t)
>>> data = values + np.random.normal(0, 1, values.shape)
>>> problem = pints.MultiOutputProblem(m, t, data)
>>> l_iid = pints.GaussianLogLikelihood(problem)
>>> l_mult = pints.MultiplicativeGaussianLogLikelihood(problem)
>>> print(l_iid([2, 6, 7, 1, 1, 1]))
-712.2134008238952
>>> print(l_mult([2, 6, 7, 0, 1, 0, 1, 0, 1]))
-712.2134008238952
```

DavAug:
That is correct. I didn't realise that the factor `self._logn` is summed for every output. So it the value is computed correctly afterall. I'll just add the tests that you suggested to `TestMulplicativeGaussianLogLikelihood`.

DavAug:
So no bug afterall, just my confusion.

DavAug:
The contribution of this PR is then not to fix a non-existent bug (:D), but to slightly improve the docstring and add agreement tests between a GaussianLogLikelihood and a MultiplicativeGaussianLogLikelihood with eta = 0.

pints/_log_likelihoods.py
```diff
@@ -578,57 +578,78 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
...
+        no = problem.n_outputs()
+        self._np = 2 * no  # 2 parameters added per output
 
         # Add parameters to problem
-        self._n_parameters = problem.n_parameters() + 2 * self._no
+        self._n_parameters = problem.n_parameters() + self._np
 
         # Pre-calculate the constant part of the likelihood
-        self._logn = 0.5 * self._nt * np.log(2 * np.pi)
+        self._logn = 0.5 * self._nt * no * np.log(2 * np.pi)
```
MichaelClerx, line 635: This is still a change, so will let @rcw5890 review this!

DavAug: True. I moved `self._logn` outside the sum in the likelihood computation. So we don't sum `self._logn` `no` times but just precompute the constant factor.

MichaelClerx: Cool! Sorry @rcw5890 I got mixed up with another PR that you were already reviewing :sweat:

pints/_log_likelihoods.py
```diff
@@ -578,57 +578,78 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
     .. math::
         v(t) \sim \text{ iid } N(0, \sigma)
 
-    Note that the scalar parameter ``eta`` controls the exponential dependence
-    of the noise on the function output, while the scalar parameter ``sigma``
-    provides a baseline level of the noise standard deviation. This model leads
-    to a log likelihood of
+    Note that the scalar parameter :math:`\eta` controls the exponential
```
MichaelClerx, line 581: I think f and X need to be introduced here

MichaelClerx: Also f(t, \theta) instead of f(t; \theta) (twice)

pints/_log_likelihoods.py
```diff
@@ -578,57 +578,78 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
...
             -\frac{n_t}{2} \log{2 \pi}
             -\sum_{i=1}^{n_t}{\log{f(t_i, \theta)^\eta \sigma}}
             -\frac{1}{2}\sum_{i=1}^{n_t}
-                \frac{(X(t_i) - f(t_i, \theta))^2}
+                \frac{(X^{\text{obs}}_{i} - f(t_i, \theta))^2}
                 {(f(t_i, \theta)^\eta \sigma)^2}
 
-    where ``n_t`` is the number of time points in the series.
+    where :math:`n_t` is the number of time points in the series, and
+    :math:`X^{\text{obs}}_{i}` the measurement at time :math:`t_i`.
```
MichaelClerx, line 595: Not sure about calling it X^obs here, is it not the same X as above?

pints/_log_likelihoods.py
```diff
@@ -578,57 +578,78 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
...
+    This model leads to a log likelihood of
 
     .. math::
-        \log{L(\theta, \sigma, \eta | \boldsymbol{x})} =
+        \log{L(\theta, \eta , \sigma | X^{\text{obs}})} =
             -\frac{n_t}{2} \log{2 \pi}
             -\sum_{i=1}^{n_t}{\log{f(t_i, \theta)^\eta \sigma}}
             -\frac{1}{2}\sum_{i=1}^{n_t}
-                \frac{(X(t_i) - f(t_i, \theta))^2}
+                \frac{(X^{\text{obs}}_{i} - f(t_i, \theta))^2}
```
MichaelClerx, line 591: Put the square around the entire frac maybe? Or is that uglier?

pints/_log_likelihoods.py
MichaelClerx: I mean, the model says that our observations should be distributed a certain way, so same X in that sense

pints/_log_likelihoods.py
```diff
@@ -578,57 +578,78 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
...
 
-    where ``n_t`` is the number of time points in the series.
+    where :math:`n_t` is the number of time points in the series, and
+    :math:`X^{\text{obs}}_{i}` the measurement at time :math:`t_i`.
+
+    For a system with :math:`n_o` outputs, this becomes
+
+    .. math::
+        \log{L(\theta, \eta , \sigma | \boldsymbol{x})} =
+            -\frac{n_tn_o}{2} \log{2 \pi}
```
MichaelClerx, line 601: n_t n_o please, just for clarity :D

pints/_log_likelihoods.py
```diff
@@ -578,57 +578,78 @@ class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
...
+            -\sum ^{n_o}_{j=1}\sum_{i=1}^{n_t}{\log{f_j(t_i, \theta)^\eta
+            \sigma _j}}
+            -\sum ^{n_o}_{j=1}\sum_{i=1}^{n_t}
```
MichaelClerx, line 604: Please put the 1/2 in front of the sum again, to match the notation above!

pints/_log_likelihoods.py
DavAug: I like to distinguish between the random variable X, and the realisations of that random variable (observations) X^obs. So strictly speaking X and X^obs are not the same. But I can remove the 'obs' if you'd prefer that.

MichaelClerx: No fair enough, leave it in!

pints/_log_likelihoods.py
MichaelClerx: (with a space, I mean)

pints/_log_likelihoods.py
DavAug: Done

pints/_log_likelihoods.py
DavAug: Agreed. I put the square around the fraction. It looks a little better.

pints/_log_likelihoods.py
DavAug: done

pints/_log_likelihoods.py
DavAug: done

pints/_log_likelihoods.py
```diff
@@ -562,48 +562,51 @@ def __init__(self, problem, sigma):
 
 class MultiplicativeGaussianLogLikelihood(pints.ProblemLogLikelihood):
     r"""
-    Calculates a log-likelihood assuming heteroscedastic Gaussian errors, with
-    the magnitude of the error variance scaling with the problem output.
+    Calculates the log-likelihood for a time-series model assuming a
+    heteroscedastic Gaussian error of the model predictions
+    :math:`f(t, \theta )`. The standard deviation of a heteroscedascic
```
MichaelClerx, line 567: The first paragraph of a docstring should usually be a single sentence.

Maybe we could write "...assuming a heteroscedastic (see below) Gaussian error of..." and leave out the second sentence?

DavAug: No problem

DavAug:
Thanks for reviewing!

Issue #1168, opened 2020-06-22
Autocorrelation plot needs some documentation fixes


MichaelClerx:
- docs state it returns `(fig, ax)` when in fact it returns `(fig, array-of-axes)`
- some info about the interpretation should be given in notebook: https://github.com/pints-team/pints/blob/master/examples/plotting/mcmc-autocorrelation.ipynb
- notebook ends in empty cell

Issue #1163, opened 2020-06-08
Python 3.8.2 Ubuntu 20.04 installation error


chonlei:
Trying out Python 3.8.2 on the new Ubuntu 20.04 and the installation throws me this error messages
```
Building wheels for collected packages: pandocfilters, tornado, backcall, pyrsistent
  Building wheel for pandocfilters (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: pints-dev/venv/bin/python3 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-hbyhwwff/pandocfilters/setup.py'"'"'; __file__='"'"'/tmp/pip-install-hbyhwwff/pandocfilters/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-cy4koyqv
       cwd: /tmp/pip-install-hbyhwwff/pandocfilters/
  Complete output (6 lines):
  usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: setup.py --help [cmd1 cmd2 ...]
     or: setup.py --help-commands
     or: setup.py cmd --help
  
  error: invalid command 'bdist_wheel'
  ----------------------------------------
  ERROR: Failed building wheel for pandocfilters
  Running setup.py clean for pandocfilters

  ...

Failed to build pandocfilters tornado backcall pyrsistent
```
It was simply missing the `wheel` package --- are we expecting this to be installed as part of the python?!

MichaelClerx:
I'd say that's an issue with the `pandocfilters` package then ?

chonlei:
Sounds good -- if that's the case, then I will close this issue!

PR #1190, opened 2020-08-11
I1170 combined gaussian loglikelihood

DavAug:
See PR #1163 

This is probably not the right way to do it, but...

I checked out a previous version of the branch in #1163 and added some changes. When trying do merge this branch into the branch in #1168 I ran into some annoying merge conflicts, so I thought it might be easier to just open a new PR.


pints/_log_likelihoods.py
```diff
@@ -237,6 +237,228 @@ def __call__(self, x):
         )
 
 
+class CombinedGaussianLogLikelihood(pints.ProblemLogLikelihood):
+    r"""
+    Calculates the log-likelihood for either a
```
MichaelClerx, line 242: Would try shortening this a bit and leaving the `:math:` to the second paragraph.
E.g. "Log-likelihood that assumes a mixture of Gaussian noise and Gaussian heteroscedastic noise.". And then the 2nd paragraph gives math, explanation of heteroscedastic, single and multi, etc.

pints/_log_likelihoods.py
```diff
@@ -237,6 +237,228 @@ def __call__(self, x):
         )
 
 
+class CombinedGaussianLogLikelihood(pints.ProblemLogLikelihood):
```
MichaelClerx, line 240: "Combined gaussian" sounds much more broad than what this actually is. For example it doesn't combine two N(mu, sigma) log likelihoods

pints/_log_likelihoods.py
MichaelClerx: For example, does this generalise to some kind of polynomial normal + normal * model + normal * model^2 + ... etc.?
Needs more specific name!

DavAug: Done

MichaelClerx:
I quite like what it does to the logistic example!

DavAug:
> Needs a more specific name. Other than that I don't really know what this is about so happy for others to review :D

Thank @MichaelClerx !

MichaelClerx:
Is it maybe a linear heteroscedastic gaussian log likelihood?

MichaelClerx:
> Is it maybe a linear heteroscedastic gaussian log likelihood?

There's a certain "a + bx"-ishness about its heteroscedasticity :D

DavAug:
> Is it maybe a linear heteroscedastic gaussian log likelihood?

I'm fine with any name really. I thought because the heteroscedastic noise is called MultiplicativeGaussian, it would be most consistent to call it ConstantAndMultiplicativeGaussian?

DavAug:
> > Is it maybe a linear heteroscedastic gaussian log likelihood?
> 
> There's a certain "a + bx"-ishness about its heteroscedasticity :D

Agreed! Let's see whether other people have string feelings about this. Otherwise I'll go with LinearHeteroscedasticGaussianLogLikelihood

MichaelClerx:
ConstantAndMultiplicative is fine by me too! You decide 😄

docs/source/log_likelihoods.rst
```diff
@@ -19,6 +19,7 @@ Overview:
 - :class:`AR1LogLikelihood`
 - :class:`ARMA11LogLikelihood`
 - :class:`CauchyLogLikelihood`
+- :class:`LinearHeteroscedasticGaussianLogLikelihood`
```
ben18785, line 22: Sorry, can we make this alphabetical?

docs/source/log_likelihoods.rst
```diff
@@ -35,6 +36,8 @@ Overview:
 
 .. autoclass:: CauchyLogLikelihood
 
+.. autoclass:: LinearHeteroscedasticGaussianLogLikelihood
```
ben18785, line 39: Same.

examples/README.md
```diff
@@ -82,6 +82,7 @@ relevant code.
 ## Statistical modelling
 - [Autoregressive moving average errors](./stats/autoregressive-moving-average-errors.ipynb)
 - [Cauchy sampling error](./stats/cauchy-sampling-error.ipynb)
+- [Linear heteroscedastic Gaussian error](./stats/linear_heteroscedastic_gaussian_error.ipynb)
```
ben18785, line 85: Alphabetical too.

pints/__init__.py
```diff
@@ -113,6 +113,7 @@ def version(formatted=False):
     AR1LogLikelihood,
     ARMA11LogLikelihood,
     CauchyLogLikelihood,
+    LinearHeteroscedasticGaussianLogLikelihood,
```
ben18785, line 116: Alphabetical please

pints/_log_likelihoods.py
```diff
@@ -237,6 +237,226 @@ def __call__(self, x):
         )
 
 
+class LinearHeteroscedasticGaussianLogLikelihood(
```
ben18785, line 240: Sorry, being dense -- why is it linear? Doesn't eta mean it can be non-linear?

DavAug:
> Looks great -- apart from the few annoying comments about language (sorry!), think we're good to go. For the notebook, because it's hard to comment inline here are some comments here:
> 
> * Should be "logistic growth" not "logistic Growth" in the description
> * The sentence "where the model parameters..." doesn't make sense at the moment
> * I'm not sure we need both the adaptive covariance inference and HMC: I'd just use one of these across your examples. At the moment, the notebook is quite long and, because of that, a bit hard to read.
> * I'd turn off printing to screen in your MCMC running as you have so many of these examples
> * You often put comments using "#" within the blocks rather than write them outside them using Markdown (for example, you have somewhere "Compute diagnostics and show diagnostic plots." within a cell). I'd write them outside as that makes it easier to read.

Ok, cool thanks @ben18785! I'll make the changes.

pints/_log_likelihoods.py
DavAug: That is correct! Hmm I don't know why I thought it's linear. It somehow sounded need :D. Anyway it's not correct. Well nonlinear is probably to general. Let's just go back to ConstantAndMultiplicativeGaussianError then?

ben18785:
Yep!

docs/source/log_likelihoods.rst
DavAug: done

docs/source/log_likelihoods.rst
DavAug: done

pints/__init__.py
DavAug: done

pints/_log_likelihoods.py
DavAug: done

examples/README.md
DavAug: done

DavAug:
@MichaelClerx @ben18785 
I've incorporated the suggested changes. I also added a very brief discussion in the notebook how one may deduce that the MCMC run is not successful at first.

Let me know what you think. :)

ben18785 (changes requested):
Looks good @DavAug -- thanks for working through these changes. A couple more things then good to go:

- Can you change the notebook title to "Inference using a two compartment error model incorporating constant and multiplicative errors". Only the first letter "I" should be capitalised.
- There are quite a few comments that start with #s in the cells themselves. Some of these are repetitive: for example, "Infer model parameters with Haario ACMCMC." repeats the text above it. Also, many of these comments at the top of code blocks should be moved outside of them: that'll make them look neater. Can you clean these up?

DavAug:
> Looks good @DavAug -- thanks for working through these changes. A couple more things then good to go:
> 
> * Can you change the notebook title to "Inference using a two compartment error model incorporating constant and multiplicative errors". Only the first letter "I" should be capitalised.
> * There are quite a few comments that start with #s in the cells themselves. Some of these are repetitive: for example, "Infer model parameters with Haario ACMCMC." repeats the text above it. Also, many of these comments at the top of code blocks should be moved outside of them: that'll make them look neater. Can you clean these up?

Sure! Is this how those models are called "two compartment error model"?

ben18785:
@DavAug no (I don't know how those models are called) but perhaps that's a better title?

DavAug:
@ben18785 
Thanks again for the review. I tried to incorporate your change requests. Would you prefer if I removed all comments on the top of the cells?

ben18785:
No need to remove them -- just make the text markdown (i.e. move it above them) as, when they're in the cells, the comments look a bit messy.

DavAug:
> No need to remove them -- just make the text markdown (i.e. move it above them) as, when they're in the cells, the comments look a bit messy.

Sorry I think I'm a little slow right now.. Would you prefer to see
[ ] the requirements for each cell in markdown above the cell, 
[ ] generally when there is a stream of multiple cells a markdown above each cell that contains the short comment that is currently at the top of each cell, or
[ ] both of the above?

ben18785:
Ha, sorry, this is quite hard to communicate and I'm not doing a good job of this. It's easier to explain with a few examples. In general, my point is, you aren't currently making sufficient use of markdown text. The comments written at the top of most cells (but within them) would look much nicer if they were in markdown.

Take the cell immediately below, "Show quantitative and visual diagnostics of MCMC runs". Within the cell, you have a comment, ``This cell needs the HaarioACMC chains'', which is written using #s as it's a comment in Python. This would look better if this comment were moved above the cell (and perhaps explained a little) and written in markdown (immediately below "Show quantitative and visual diagnostics of MCMC runs").

Similarly, take the cell with "Construct posterior" in it. This cell has no markdown comments above it currently. The stuff that's written from "Construct posterior. This cell needs the above defined true parameters, the generated data, and the model wrapper to be able to fix eta" could be moved up to markdown text. It could also perhaps be explained a bit better.

A separate thing that I've just noticed is that you seem to be reloading packages: for example, you import pints a number of times. Can we just have these once at the top (or inline if that makes it easier to understand)?

DavAug:
> Ha, sorry, this is quite hard to communicate and I'm not doing a good job of this. It's easier to explain with a few examples. In general, my point is, you aren't currently making sufficient use of markdown text. The comments written at the top of most cells (but within them) would look much nicer if they were in markdown.
> 
> Take the cell immediately below, "Show quantitative and visual diagnostics of MCMC runs". Within the cell, you have a comment, ``This cell needs the HaarioACMC chains'', which is written using #s as it's a comment in Python. This would look better if this comment were moved above the cell (and perhaps explained a little) and written in markdown (immediately below "Show quantitative and visual diagnostics of MCMC runs").
> 
> Similarly, take the cell with "Construct posterior" in it. This cell has no markdown comments above it currently. The stuff that's written from "Construct posterior. This cell needs the above defined true parameters, the generated data, and the model wrapper to be able to fix eta" could be moved up to markdown text. It could also perhaps be explained a bit better.
> 
> A separate thing that I've just noticed is that you seem to be reloading packages: for example, you import pints a number of times. Can we just have these once at the top (or inline if that makes it easier to understand)?

Ah ok! I'm happy to do those changes :)

The reason why I include the requirements, i.e. "this cell needs...", and the imports in each cell is that I would like to avoid the problem of notebooks that the order of execution matters. Most notebooks only work from top to bottom. This way I highlight which cells need to be executed first, because they compute some variables that are needed. By included the relevant imports in each cell, I avoid that dependence between cells. (that is also why I'd prefer to have the "comment" dependencies ("this cell needs...") within the cell and not as markdown.

I hope that makes sense. But if you have strong feelings about this, I am happy to change it.

ben18785:
Makes sense but I think we, in general, try to have the notebooks so that they run from top to bottom and only import packages once (I'm not sure we always manage to do this though!). So can we change to make it like this?

DavAug:
> Makes sense but I think we, in general, try to have the notebooks so that they run from top to bottom and only import packages once (I'm not sure we always manage to do this though!). So can we change to make it like this?

Yep :)

ben18785 (approved):
Looks great! Thanks @DavAug

ben18785:
@MichaelClerx Not sure if there were any changes outstanding with this? I'm happy for this to be merged.

MichaelClerx (approved):

MichaelClerx:
I've merged it in. Thanks all!

ben18785:
Great, thanks Michael!

DavAug:
Thank you Ben and Michael.


I've merged it in. Thanks all!

PR #1199, opened 2020-08-27
Fix README typo

chonlei:
Fix typo in README after updating for pip install


MichaelClerx:
Already handled here 1195

chonlei:
@MichaelClerx Sorry, missed that, tried looking for an issue but couldn't find one. :sweat_smile:

PR #1201, opened 2020-09-02
Notebook fixes

MichaelClerx:
Three(!) notebooks were failing.

The nuts one failed because `rhat_all_params` had been removed from the API (not deprecated), so fixed that as well


rcw5890 (commented):
In residuals-autocorrelation.ipynb, the Lotka-Volterra example was randomly failing on some runs due to a bad initial condition far away from the true parameter values (it would still have worked eventually, but with more MCMC iterations than we were doing in the notebooks). I moved the initial condition slightly closer to the true values, which seems to give a consistent correct result with the same number of iterations.

DavAug (approved):
Sorry my bad for removing the `rhat_all_params` function instead of deprecating it. 

Looks good to me, I have difficulties to figure out what has changed in the constant_and_multiplicative notebook though :D

MichaelClerx:
> Sorry my bad for removing the `rhat_all_params` function instead of deprecating it.

No worries! 

> Looks good to me, I have difficulties to figure out what has changed in the constant_and_multiplicative notebook though :D

Yeah github doens't like notebooks much (they contain embedded images, which makes github think there's a massive change).
You can check it out with a local client (I use meld) by doing `git difftool master`. You need to tell it which tool though, which is something I always google :D

Changes to the notebook are small: just moved the imports around and made sure `pints.plot` was imported. It's important to always restart and re-run the notebook before submitting the final one: sometimes you've deleted a crucial import or variable from the notebook but it still works because it's remembered in the state

martinjrobins (approved):
thanks @MichaelClerx for catching this

Issue #1205, opened 2020-09-03
Unit testing fails on OS/X


MichaelClerx:
Getting this on all os/x builds, in all PRs:

```
======================================================================
FAIL: test_logging (test_opt_optimisation_controller.TestOptimisationController)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/pints/pints/pints/tests/test_opt_optimisation_controller.py", line 127, in test_logging
    self.assertEqual(log_should_be, c.text())
AssertionError: 'Maxi[333 chars]0:00.0\n10    60    -4.140462   0:00.0\nHaltin[43 chars]d.\n' != 'Maxi[333 chars]0:00.1\n10    60    -4.140462   0:00.1\nHaltin[43 chars]d.\n'
Diff is 642 characters long. Set self.maxDiff to None to see it.

----------------------------------------------------------------------
```

Anyone have a mac they can debug this with?

chonlei:
This is fixed in #1206

MichaelClerx:
Not quite fixed! But should be easier to debug now when it re-appears!

chonlei:
Ha, yeah, not a fix, it appears again
```
======================================================================
FAIL: test_logging (test_opt_optimisation_controller.TestOptimisationController)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/runner/work/pints/pints/pints/tests/test_opt_optimisation_controller.py", line 130, in test_logging
    self.assertEqual(log_should_be, c.text())
AssertionError: 'Maxi[333 chars]0:00.0\n10    60    -4.140462   0:00.0\nHaltin[43 chars]d.\n' != 'Maxi[333 chars]0:00.1\n10    60    -4.140462   0:00.1\nHaltin[43 chars]d.\n'
  Maximising LogPDF
  Using Exponential Natural Evolution Strategy (xNES)
  Running in sequential mode.
  Population size: 6
  Iter. Eval. Best      Time m:s
  0     6     -4.140462   0:00.0
  1     12    -4.140462   0:00.0
  2     18    -4.140462   0:00.0
  3     24    -4.140462   0:00.0
  6     42    -4.140462   0:00.0
- 9     60    -4.140462   0:00.0
?                              ^
+ 9     60    -4.140462   0:00.1
?                              ^
- 10    60    -4.140462   0:00.0
?                              ^
+ 10    60    -4.140462   0:00.1
?                              ^
[...]

PR #1206, opened 2020-09-03
Broken test on OS/X

MichaelClerx:
See #1205.


MichaelClerx:
Mystery error has gone away again. Suggest we still merge this in in case it happens again

chonlei (approved):
Good idea to have this fixed -- I think I have seen this a few times (and disappears suddenly).

PR #1207, opened 2020-09-04
Removed future imports from residuals autocor diag notebook.

MichaelClerx:
Was falling over on travis

Re: `__future__` imports in notebooks. I _think_ it's OK if they appear as the very first lines, but if they re-appear after that they will break in testing.
The reason is that the notebook code gets extracted and stuck into a single file. This file, like all python files, must obey the rule that imports are all at the very top. In notebooks, this rule is relaxed, so you won't get an error if you test locally.

I'd feel quite comfortable leaving future imports out of notebooks altogether. I want main PINTS to stay 2.7 compatible for a while so that it can run on big systems with ancient configurations, set up by university admins. But I think the examples will probably be run on nice shiny machines within the users control...


rcw5890 (approved):
Thanks for fixing this!

Your explanation makes sense to me. There are quite a few other ``from __future__`` ... in the other notebooks, but always in the first line.

PR #1212, opened 2020-09-09
I1211 pooled log pdf

### LINKED ISSUE #1211: PooledLogPDF
A pints.LogPDF that takes a list of pints.LogPDFs with the same number of parameters, and allows to pool a subset of those parameters across LogPDFs. 

This is useful in situations where you have measurements of a number of biologically distinct entities whose overall behaviour is nevertheless expected to be captured by the same structural model. The structural mode, say a system of ODEs, is the same across individuals, but some parameters are allowed to vary. 

So far we can only infer the model parameters for each individual separately, but if you have only a very limited number of observations, it useful to be able to link e.g. the noise parameters across individuals.

DavAug:
Closes #1211 .

I implemented a PooledLogPDF that allows pooling of parameters. Please have a look at the notebook for its benefits for some inference problems.

I also replaced the test_log_posterior.py by test_log_pdfs.py and moved the test for the SumOfIndependentLogPDFs into that file, to follow more closely the pints structure (I find it easier to find the tests that way, but I can revert it).


DavAug:
@MichaelClerx 
Hey checkout the notebook! I didn't even expect such a huge influence :D

MichaelClerx (commented):
Cool! Will try to have a detailed look before the next meeting

pints/_log_pdfs.py
```diff
@@ -115,6 +115,177 @@ def sample(self, n=1):
         raise NotImplementedError
 
 
+class PooledLogPDF(LogPDF):
+    r"""
+    Calculates a sum of :class:`LogPDF` objects, while pooling selected
```
ben18785, line 120: Creates an object of :class:`LogPDF`, with a log-pdf which is the sum of the individual log-pdfs.

pints/_log_pdfs.py
```diff
@@ -115,6 +115,177 @@ def sample(self, n=1):
...
+    parameters across log-pdfs.
+
+    This :class:`LogPDF` requires that all :class:`LogPDF` objects have the
+    same number of parameters.
+
+    This is useful for e.g. modelling the time-series of multiple individuals
+    (each individual defines a separate :class:`LogPDF`), and some parameters
+    are expected to be the same across individuals. This may be a realistic
```
ben18785, line 128: "are expected to be the same across individuals (for example, the noise parameter across different individuals within the same experiment)."

pints/_log_pdfs.py
```diff
@@ -115,6 +115,177 @@ def sample(self, n=1):
...
+
+    In general, the search of a :class:`PooledLogPDF` has dimensionality
+    :math:`nd_{\text{up}} + d_{\text{p}}`, where :math:`n` is the number of
+    individuals, :math:`d_{\text{up}}` is the dimension of the unpooled
+    :math:`psi` parameters, and :math:`d_{\text{up}}` is the dimension of the
+    pooled :math:`\theta` parameter.
+
+    The order of parameters for evaluation is largely kept, and the parameters
+    of individual log-pdfs are concatenated. However, pooled parameters are
+    always prepended. Consider for example two log-pdf with parameters
```
ben18785, line 152: two log-pdfs

ben18785 (changes requested):
Thanks @DavAug -- looks really good. Apart from my minor inline comments, can we do the following:

- Reduce the number of individuals you show results for in the example notebook (at the moment, the notebook is a bit long): perhaps, just show the individual results for 2 individuals.
- Try running HMC or NUTS on the model in the notebook. I find this is a useful way to diagnose whether the sensitivities are all fine.

pints/_log_pdfs.py
DavAug: Done

pints/_log_pdfs.py
DavAug: Done

DavAug:
Thanks @ben18785 ! Good spots! I'll do the requested changes in the notebook.

DavAug:
@ben18785 
I made the requested changes. Let me know if there is anything else. 

@ben18785 @martinjrobins 
I tried to use NUTS (with vanilla settings) first but it was terribly slow and just diverged after 200 something iterations. I didn't leave it in the notebook, but is this something that should happen with NUTS?

ben18785 (approved):
@DavAug Thanks for those. Don't worry about NUTS -- I suspect we need to do variable transformations to get it to work... @martinjrobins -- do you think it's worth adding a notebook where we show how to get NUTS working for various toy problems? I'm just concerned that people will try NUTS -- find it doesn't work -- and abandon it, just because they haven't got the parameter transforms right.

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
         raise NotImplementedError
 
 
+class PooledLogPDF(LogPDF):
+    r"""
+    Creates an object of :class:LogPDF, with a log-pdf which is the sum of the
```
MichaelClerx, line 120: ``` :class:`LogPdf` ```

I also wouldn't start with "Creates an object of" when describing the class itself

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+    individual log-pdfs, while pooling selected parameters.
+
+    This :class:`LogPDF` requires that all :class:`LogPDF` objects have the
```
MichaelClerx, line 123: I'd put this lower down after it has become clearer that there are several logpdfs involved

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+    where :math:`\theta := \theta _1 = \theta _2`, and $D_i$ is the measured
+    time-series of individual :math:`i`.
+
+    In general, the search of a :class:`PooledLogPDF` has dimensionality
+    :math:`nd_{\text{up}} + d_{\text{p}}`, where :math:`n` is the number of
+    individuals, :math:`d_{\text{up}}` is the dimension of the unpooled
+    :math:`psi` parameters, and :math:`d_{\text{up}}` is the dimension of the
+    pooled :math:`\theta` parameter.
+
+    The order of parameters for evaluation is largely kept, and the parameters
```
MichaelClerx, line 149: "is largely kept" sounds to vague for me! Just write a precise description (like below) :-)

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+    of individual log-pdfs are concatenated. However, pooled parameters are
+    always prepended. Consider for example two log-pdf with parameters
+    :math:`(\psi _{1,1}, \psi _{1, 2}, \psi _{1,3}, \psi _{1,4})` and
+    :math:`(\psi _{2,1}, \psi _{2, 2}, \psi _{2,3}, \psi _{2,4})`. Pooling the
+    first and the last parameter, i.e.
+    :math:`\theta _1 := \psi _{1,1} = \psi _{2, 1}` and
```
MichaelClerx, line 155: did you check that these render correctly (i.e. by building the docs locally) ?

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+    \theta _4)`.
+
+    Extends :class:`LogPDF`.
+
+    Parameters
+    ----------
+    log_pdfs
+        A sequence of :class:`LogPDF` objects.
+    pooled
+        An array-like object of dtype bool, indicating which parameters across
```
MichaelClerx, line 168: If we're going to say they need a `dtype`, then just say it _must_ be a `numpy.ndarray` (because they have an argument called `dtype`, so that's highly numpy-specific!)

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+        the likelihoods are pooled (`True`) or remain unpooled (`False`).
```
MichaelClerx, line 169: ``` `True` ``` --> ``` ``True`` ```

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+
+        # Check that pooled matches number of parameters
+        self._pooled = np.asarray(pooled)
+        if len(self._pooled) != n_parameters:
+            raise ValueError(
+                'The array-like input `pooled` needs to have the same length '
+                'as the number of parameters of the individual log-pdfs.')
+
+        # Check that pooled contains only booleans
+        for p in self._pooled:
```
MichaelClerx, line 212: Shouldn't we be checked the dtype of `_pooled` instead?

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
 
+class PooledLogPDF(LogPDF):
+    r"""
+    Creates an object of :class:LogPDF, with a log-pdf which is the sum of the
+    individual log-pdfs, while pooling selected parameters.
+
+    This :class:`LogPDF` requires that all :class:`LogPDF` objects have the
+    same number of parameters.
+
+    This is useful for e.g. modelling the time-series of multiple individuals
```
MichaelClerx, line 126: Either the first sentence needs to be clearer, or there needs to be more explanation here (before going into the use for this class).

I think we might have to use numbers/equations in the initial explanation, just to get things clear. Because I can't think of standard terminology that will let us explain this in words?
E.g. something like 

```
    Combines ``m`` :class:`LogPDFs<pints.LogPDF>`, each with ``n`` parameters,
    into a single ``LogPDF`` where one or more parameters are "pooled" (i.e.
    have the same value for each ``LogPDF``), so that the resulting combined 
    ``LogPDF`` has ``m * n_unpooled + n_pooled`` parameters.
```

Or maybe "where k parameters are pooled" so you can write ``m * (n - k) + k`` ? 

(note 1: You can do :class:`Hello<RealClassName>` to make the link say hello)

(note 2: No need to do a :class:`X` twice in the same sentence, just use ``X`` for the second one)

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+        implement the optional method :meth:`LogPDF.evaluateS1()`!*
+        """
+        # Get parameters of pooled log-pdf
+        parameters = np.asarray(parameters)
+
+        # Create container for parameters of individuals log-pdf and fill with
+        # pooled parameters
+        params_ind = np.empty(shape=self._n_unpooled + self._n_pooled)
+        if self._n_pooled > 0:
+            params_ind[self._pooled] = parameters[
```
MichaelClerx, line 261: Is this supposed to set just a single parameter?

pints/_log_pdfs.py
```diff
@@ -115,6 +115,178 @@ def sample(self, n=1):
...
+        for p in self._pooled:
+            if not isinstance(p, np.bool_):
+                raise ValueError(
+                    'The array-like input `pooled` passed to PooledLogPDFs '
+                    'has to contain booleans exclusively.')
+
+        # Get dimension of search space
+        self._n_pooled = np.sum(self._pooled)
+        n_individuals = len(self._log_pdfs)
+        self._n_unpooled = np.sum(~self._pooled)
```
MichaelClerx, line 221: Easier to write n_parameters - n_pooled?

MichaelClerx:
Why the change to LogPosterior's test?

MichaelClerx (commented):
Bunch of comments on the docstring, but otherwise happy!

MichaelClerx:
The named figures in the notebooks are nice, but we haven't done it in the other (O(80)!) notebooks, so best not to and remain consistent

MichaelClerx:
Can we start the notebook with some 1-liner to help people decide if they want to read on? I.E. "This notebook describes the use of the PooledLogPDF class."

DavAug:
> Why the change to LogPosterior's test?

Nothing about the LogPosterior test, I just changed the location from the test_log_likelihoods.py to test_log_pdfs.py, because it reflects the directory structure of log_likelihoods.py and log_pdfs.py (plus a posterior is not a likelihood, so I was a little confused at some point when I was looking for the posterior tests). But I can move the test back to the test_log_likelihoods.py file.

pints/_log_pdfs.py
DavAug: That's much clearer!

pints/_log_pdfs.py
DavAug: Yes, but I get a couple warnings from the transformations docs @chonlei , e.g.
```
/Users/david/Workspace/sabs-r3/pints/pints/_transformation.py:docstring of pints.ComposedTransformation.convert_covariance_matrix:4: WARNING: Duplicate target name, cannot be used as a unique reference: "1".
```

I don't think this has to do with my docs though.

pints/_log_pdfs.py
DavAug: Would it be ok to say "array-like object of Booleans ... " I'd just like to clarify that this is a boolean mask to specify the pooling.

pints/_log_pdfs.py
DavAug: Yes makes sense!

pints/_log_pdfs.py
DavAug: True, but I don't know n_parameters at this point yet. I know the n_parameters of each individual log-likelihood, say p, and with the number of individuals n I can compute 
```
n_parameters = p * n - (n-1) * n_pooled
```
but I find that more complicated. What do you think?

DavAug:
@MichaelClerx Thanks for the review.

I incorporated your change requests, except the ones where I had questions or comments.

pints/_log_pdfs.py
MichaelClerx: I guess a "sequence of booleans" then?

MichaelClerx (approved):
Thanks! I'm happy to merge

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
         raise NotImplementedError
 
 
+class PooledLogPDF(LogPDF):
+    r"""
+    Combines :math:`m` :class:`LogPDFs<pints.LogPDF>`, each with :math:`n`
```
chonlei, line 120: Wonder if having n_1, n_2, ..., n_m parameters for the m models with k parameters of that being in common and pooled, where n_i does not necessary equal to n_j, would be useful? Unlike hierarchical models, I suppose there's no rule or such like that the parameters in the m models apart from the k parameters have to be the same?

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+    (i.e. have the same value for each LogPDF), so that the resulting combined
+    LogPDF has :math:`m (n - k) + k` parameters.
+
+    This is useful for e.g. modelling the time-series of multiple individuals
+    (each individual defines a separate :class:`LogPDF`), and some parameters
+    are expected to be the same across individuals (for example, the noise
+    parameter across different individuals within the same experiment).
+
+    For two :class:`LogPDFs<pints.LogPDF>` :math:`L _1`  and
+    :math:`L _2` with parameters :math:`(\psi _1, \theta _1`) and
```
chonlei, line 131: :math:\`(\psi _1, \theta _1)\` instead of :math:\`(\psi _1, \theta _1\`)?

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+    :math:`(\psi _2, \theta _2`) respectively, a pooling of the
```
chonlei, line 132: Same as above.

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+    :math:`\theta _i` results in a pooled log-pdfs of the form
+
+    .. math::
+        L(\psi _1, \psi _2, \theta | D_1, D_2) =
+            L(\psi _1, \theta | D_1) + L(\psi _2, \theta | D_2),
+
+    where :math:`\theta := \theta _1 = \theta _2`, and :math:`D_i` is the
+    measured time-series of individual :math:`i`.
```
chonlei, line 140: Maybe add something like "In this case, we have m=2, k is the size of \theta, and n is the size of psi + k"? The notation of n and k does not directly indicate the notation of the parameters \psi and \theta here.

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+                raise ValueError(
+                    'All log-pdfs passed to PooledLogPDFs must have '
+                    'same dimension.')
+
+        # Check that pooled matches number of parameters
+        self._pooled = np.asarray(pooled)
+        if len(self._pooled) != n_parameters:
+            raise ValueError(
+                'The array-like input `pooled` needs to have the same length '
+                'as the number of parameters of the individual log-pdfs.')
```
chonlei, line 193: Be good to check the Falses in `pooled` comes at the end? Since your docs ([here](https://github.com/pints-team/pints/pull/1212/files#diff-b121974bb553002da6fe16f041d9eaa6R142-R143)) say "The parameters in a :class:`PooledLogPDF` are ordered such that unpooled parameters come before pooled parameters." -- sounds important and easy to miss or implemented incorrectly. If it's not a strict requirement, then take that out.

chonlei (changes requested):
Hi David, thanks, looking good, adding some comments.

pints/_log_pdfs.py
chonlei: In particular we calculate the unpooled with `np.sum(~self._pooled)`, which does not care where the Falses are located!

pints/_log_pdfs.py
chonlei: See my comment in #1217

MichaelClerx: Have you tried with this branch checked out? https://github.com/pints-team/pints/issues/1217#issuecomment-693231083

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+
+    The parameters in a :class:`PooledLogPDF` are ordered such that unpooled
+    parameters come before pooled parameters.
+
+    Extends :class:`LogPDF`.
+
+    Parameters
+    ----------
+    log_pdfs
+        A sequence of :class:`LogPDF` objects.
```
chonlei, line 150: I think we have to say somewhere explicitly that the ordering of the resulting m (n - k) + k parameters! It'd be intuitive to think what you've done, and guess the final parameters ordered in the same way as the order of the input PDFs. And you kind of show that in the example above. However, since we are not giving names, dict, or such like to the parameters, their ordering is crucial! So it'd be best if they are explicitly explained/stated.

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+        for idx, pdf in enumerate(self._log_pdfs):
+            # Get unpooled parameters for individual
+            params_ind[~self._pooled] = parameters[
+                idx * self._n_unpooled: (idx + 1) * self._n_unpooled]
+
+            # Compute pdf score and partials for individual
+            score, partials = pdf.evaluateS1(params_ind)
+
+            # Add contributions to score and partials. Note that partials
+            # w.r.t. unpooled parameters receive only one contribution
```
chonlei, line 259: Can you explain or give ref to what you did here?

pints/_log_pdfs.py
DavAug: yes good spot!

pints/_log_pdfs.py
chonlei: Hmm, looks fine to me as well! Would that be version issue?
```
$ git status
On branch i1211-PooledLogPDF
Your branch is up to date with 'origin/i1211-PooledLogPDF'.

nothing to commit, working tree clean
$  make clean; make html
Removing everything under 'build'...
Running Sphinx v3.1.0
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 81 source files that are out of date
updating environment: [new config] 81 added, 0 changed, 0 removed
reading sources... [100%] utilities                                                                                                                                                                        
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] utilities                                                                                                                                                                         
generating indices...  genindex py-modindexdone
highlighting module code... [100%] pints.toy._twisted_gaussian_banana                                                                                                                                      
writing additional pages...  searchdone
copying static files... ... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in build/html.
```

pints/_log_pdfs.py
```diff
@@ -115,6 +115,161 @@ def sample(self, n=1):
...
+    r"""
+    Combines :math:`m` :class:`LogPDFs<pints.LogPDF>`, each with :math:`n`
+    parameters, into a single LogPDF where :math:`k` parameters are "pooled"
+    (i.e. have the same value for each LogPDF), so that the resulting combined
+    LogPDF has :math:`m (n - k) + k` parameters.
+
+    This is useful for e.g. modelling the time-series of multiple individuals
+    (each individual defines a separate :class:`LogPDF`), and some parameters
+    are expected to be the same across individuals (for example, the noise
+    parameter across different individuals within the same experiment).
```
chonlei, line 128: Can we get some textbook reference for this in general?

chonlei: So that users can learn more about this if they are unsure whether they are using the right thing/approach for their problem!

pints/_log_pdfs.py
DavAug: The pooled mask is wrt the parameters of the individual likelihoods, and you are free to pool any of the parameters, i.e. pooled = [True, False] is allowed. It means that the first parameter is pooled across all individuals. However for the input parameters of the pooled log-pdf the unpooled parameters of the individuals would come before the pooled parameters. Does that make sense?

pints/_log_pdfs.py
DavAug: Ok agreed. I'll add that to the docs.

pints/_log_pdfs.py
DavAug: I am not sure whether there is a text book reference. @ben18785 Would you know of one?

ben18785:
@DavAug, I'm not sure there is one to be honest! Think it's fine as long as
things are clearly defined.

pints/_log_pdfs.py
DavAug: I was wondering the same at first. What speaks against doing that are mainly two things: 1. I couldn't really come up with a biologically motivated problem that would require pooling parameters across different ODE models, but we could maybe find such a problem; 2. The interface for pooling across models with varying parameter spaces would have to be much more complicated, because you would have to specify which parameter (i.e position) across all individuals is going to be published, and you would have to do that for every parameter that you want to pool. 

So I would suggest that we leave this for now, unless someone has a really convincing use case. What do you think?

pints/_log_pdfs.py
DavAug: I am not sure whether there is a ref. But I tried to extend the comment a little bit so it's hopefully clearer.

Essentially what is happening is that if you have a sum of likelihoods that have all independent parameters, then the partial with respect to any of the sum of likelihoods is equal to the partial of the associated individual likelihood. But if you pool a parameter across all likelihoods in the sum, a partial derivative with respect to that parameter equals to the sum of partials.

DavAug:
@chonlei 
Thanks for the review! I incorporated your change requests or commented on them

pints/_log_pdfs.py
chonlei: Yeah, probably hard to find ref. Yes, I understand what you did, but maybe explain that in the docstring (not for me, for the users)? I don't know, @MichaelClerx what do you think?

pints/_log_pdfs.py
chonlei: OK, if there's no "textbook reference", I remember you mentioned this has been used in PKPD modelling, if there's a ref or method paper that explains this thoroughly, maybe worth pointing to?

pints/_log_pdfs.py
chonlei: Hmm... I don't know, what kind of annoys me is that it looks like "this method is very general, but not that general"...

pints/_log_pdfs.py
chonlei: Hmm... I don't know, what kind of annoys me is that it looks like "this method is very general, but not that general"...

pints/_log_pdfs.py
chonlei: I see, I must have misread the docstring as "LogPDF" instead of "PooledLogPDF", sorry!

chonlei: And probably because of the notation in the example "psi and theta" tricks me into thinking they must be ordered too...!

chonlei: This re-ordering of the parameters can be tricky! Because if I had `L_1(psi_{1,1}, theta_1, psi_{1,2})` and `L_2(psi_{2,1}, theta_1, psi_{2,2})`, this would give the final parameter as `psi_{1,1}, psi_{1,2}, psi_{2,1}, psi_{2,2}, theta_1`. This needs to be really really clear! (As in it is _not_ `psi_{1,1}, psi_{2,1}, psi_{1,2}, psi_{2,2}, theta_1` or anything else!) Because by the end, elsewhere in PINTS, we would only refer to it as parameter_1, 2, 3, 4, and 5, and we would have no idea how this is re-ordered.

pints/_log_pdfs.py
MichaelClerx: I thought it was more complicated than it is until I remembered the function is scalar.

So you just get `f=fi` and `df/dpj = dfi/dpj` for an unpooled parameter `pj` from model `i`
And for pooled ones it becomes `f=sum(fi)` and `df/dpj = sum(dfi/dpj)`

pints/_log_pdfs.py
DavAug: It is used in the PKPD modelling community, but the paper I've seen so far tend to be not very technical. I.e. they use it without reasoning why they use it. But I hope to include something on this in a future paper. :D

pints/_log_pdfs.py
DavAug: I'd say it's as general as it can be for biologically relevant problems, and the docstring specifies that all likelihoods have n parameters. 

I personally don't think that the gain in generality for allowing likelihoods with varying parameter spaces outweighs the benefits of a simple api for the use case of having likelihoods with the same parameter space. 

Anyway I'm happy to discuss this in the next meeting.

pints/_log_pdfs.py
DavAug: Agreed. So you'd say that 

"The parameters in a :class:`PooledLogPDF` are ordered such that unpooled
    parameters come before pooled parameters. The order of the unpooled
    parameters is determined by the order of the individual likelihoods."

is not explicit enough?

pints/_log_pdfs.py
DavAug: I am happy to add something to the docstring though.

chonlei: Yes, be good to have this in the docstring, I think.

chonlei: Also, say or even emphasise that you've assumed the parameters are independent variables! Because the maths break if they are not, so it's important to mention that in the docstring too (if you haven't already done so).

DavAug:
@chonlei 
Hey I incorporated some more docstring refinements. Let me know what you think

pints/_log_pdfs.py
chonlei: We decided to keep this simpler, so what it is now will do!

pints/_log_pdfs.py
chonlei: If it is not used so often, then it could even be more useful to have example references here. Such that users like me who can't come up with practical examples as to when to use this method can have something to refer to.

pints/_log_pdfs.py
chonlei: I don't know to be honest, as I (the user) am no longer the one who explicitly define the order of the parameters but the class (as elsewhere I am the one defining all the order of model parameters!), and I have no way to check what these reordered parameters are afterwards in Pints (I don't think I can!). So this makes me worried. Maybe an example with a few more parameters (just like the one I showed above) would help...? @MichaelClerx what do you think?

MichaelClerx: I agree the docstring needs to be very clear on this, with a clear description and an example (or two?)

DavAug: @chonlei I tried to make it as clear as possible. Let me know what you think

chonlei: @DavAug Thank you, with the example, it is much clearer now! I have made some slight tweaking to make it a bit more explicit. Make sure you git pull before git committing!

chonlei: I am happy now. Thanks!

chonlei (approved):

chonlei:
Ha we have forgotten about the CHANGELOG.md :smile:

DavAug:
> Ha we forgot updating the CHANGELOG.md 😄

Upps, let me fix that now

Issue #45, opened 2017-10-05
Add Model wrapper that uses Gaussian processes to evaluate


MichaelClerx:
Possibly based on GPy or GPflow

martinjrobins:
I can see three options here:
1. Use GPflow or GPy (I'm leaning towards GPy, since it doesn't have tensorflow as a dependency). This is the easiest, but adds another dependency. All that probably would be required is writing an example showing how to use GPy/GPflow to emulate a pints model
2. Write code to implement a GP model in pints. Myself and @sanmitraghosh are planning to do this for a paper, but the code would be (partially) in C++, so would require adding C++ compilation functionality back into pints
3. Same as 2, but split this code into a separate repository in the pints organisation, then the main pints code can add this as a dependency. This might be a pain to develop both code-bases in parallel....

I'm leaning towards (1) for the short term, leading towards (2) in the longer term if the paper with @sanmitraghosh  shows some good result for our approach. Any other thoughts?

martinjrobins:
I've got some good results at using Aboria to solve for the GP, so I'm going to experiment with fitting a GP using this.

sanmitraghosh:
@martinjrobins  Great news !

martinjrobins:
I'll let you know how it goes. I'm going to stick to the matern 3/2 kernel for the moment, as that seems to be the easiest to solve (Gaussian I haven't managed to do efficiently for dimensions higher than 3)

sanmitraghosh:
I think you need not worry about Gaussian kernel, matern is much more useful in practical applications.

martinjrobins:
couple of implementation points, I'm planning to:
1. implement the Adam stochastic gradient descent algorithm in pints, to use for fitting the GP
2. the GP itself will be derived from  pints.log_pdf

MichaelClerx:
Hey @martinjrobins what's the status of this? Is it still within scope for PINTS?

MichaelClerx:
@martinjrobins can we close this? Seems like it would be a project using PINTS, rather than a part of PINTS? Could even be something that lives in another `pints-team` repo?

PR #1226, opened 2020-11-03
Issue 45 gp

MichaelClerx:
Investigation into #45.
Have created a draft PR just so it's easier to look at!


ben18785:
@martinjrobins I'm guessing that, given that this has C++ code in it, we wouldn't want this as a PR? Do we want to close and keep as a branch?

martinjrobins:
Yup, closing now

MichaelClerx:
Thanks both!
Should this still be here then? Or should it be e.g. in its own repo? Ideally I'd like to make sure these things are documented outside of our collective headspace :D

PR #1228, opened 2020-11-03
Experiments with standard library-based evaluators

MichaelClerx:
At some stage, it'd be ideal if we could drop the parallel evaluator for something implemented in python's standard library

This draft PR explores some options

### DIFF
--- a/pints/__init__.py
+++ b/pints/__init__.py
@@ -149,7 +149,9 @@ from ._error_measures import (
 from ._evaluation import (
     evaluate,
     Evaluator,
+    FuturesPoolEvaluator,
     ParallelEvaluator,
+    PoolEvaluator,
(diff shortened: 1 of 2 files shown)


ben18785:
Hi @MichaelClerx -- interesting! Can you explain to a very layman like me why we might want to do this?

MichaelClerx:
Hi @ben18785 !
Just the usual reasons
- less code in pints means less maintenance
- other people might be better at this type of thing / have more time
with the usual downside
- it might not do exactly what we want

In this case what we want (repeatedly executing a method in parallel) is so simple & common that I thought the standard solutions must be better for sure. Quick efforts to use them don't make it seem that way, but could just be I'm using them wrong? Happy to close this PR and leave the branch for later

Issue #732, opened 2019-02-20
Implicit parameters in HES model?


MichaelClerx:
```
    def implicit_parameters(self):
        """
        Returns the implicit parameters of this model.
        """
        return [self._p0[0], self._p0[1], self._kdeg]
```

It's not clear from the docstring what this does. (Also, unless we really need this functionality somewhere we should just leave them at some default value)

MichaelClerx:
Ping!

MichaelClerx:
@chonlei ?

PR #1229, opened 2020-11-03
Remaining toy model sensitivities

MichaelClerx:
Partially addresses #298 

- Hes1Model


ben18785:
@chonlei When you get a minute, could you have a look at this?

pints/toy/_hes1_michaelis_menten.py
```diff
@@ -50,14 +49,75 @@ class Hes1Model(pints.ForwardModel, ToyModel):
...
+        ret[1, 3] = 0
+        ret[2, 0] = 0
+        ret[2, 1] = 0
+        ret[2, 2] = p1
+        ret[2, 3] = 0
+        return ret
+
+    def initial_conditions(self):
+        """
+        Returns m(0).
```
MichaelClerx, line 88: Are these from ToyODEModel ?

pints/toy/_hes1_michaelis_menten.py
```diff
@@ -50,14 +49,75 @@ class Hes1Model(pints.ForwardModel, ToyModel):
...
+        """
+        return self._y0[0]
+
+    def implicit_parameters(self):
+        """
+        Returns the implicit parameters of this model.
```
MichaelClerx, line 94: Are these from ToyODEModel?

pints/toy/_hes1_michaelis_menten.py
```diff
@@ -50,14 +49,75 @@ class Hes1Model(pints.ForwardModel, ToyModel):
...
+        ret[1, 0] = v
+        ret[1, 1] = -k1 - k_deg
+        ret[1, 2] = 0
+        ret[2, 0] = 0
+        ret[2, 1] = k1
+        ret[2, 2] = -k_deg
+        return ret
+
+    def n_states(self):
+        """ See :meth:`pints.ForwardModel.n_states()`. """
```
MichaelClerx, line 119: ForwardModel -> ToyODEModel

pints/toy/_hes1_michaelis_menten.py
```diff
@@ -50,14 +49,75 @@ class Hes1Model(pints.ForwardModel, ToyModel):
...
+        """ See :meth:`pints.ToyModel.jacobian()`. """
+        m, p1, p2 = state
+        P0, v, k1, h = parameters
+        p2_over_p0 = p2 / P0
+        p2_over_p0_h = p2_over_p0**h
+        one_plus_p2_expression_sq = (1 + p2_over_p0_h)**2
+        ret = np.empty((self.n_states(), self.n_parameters()))
+        ret[0, 0] = h * p2 * p2_over_p0**(h - 1) / (
+            P0**2 * one_plus_p2_expression_sq)
+        ret[0, 1] = 0
```
MichaelClerx, line 71: These =0s aren't strictly necessary, but can keep them in if you think it makes it more readable

MichaelClerx: Oh wait, they are if you use np.empty, but you could use np.zeros instead ?

MichaelClerx:
@ben18785 I approve of this (once a few docstrings have been fixed). Happy to merge if you are

chonlei (approved):
Looks great, thanks @ben18785 !

pints/toy/_hes1_michaelis_menten.py
ben18785: I actually like these as it lays out the matrix which means it's easy to check vs what's expected when reading it.

pints/toy/_hes1_michaelis_menten.py
ben18785: Not sure I understand? The initial conditions are set in this class's `set_initial_conditions` method.

ben18785: There is no initial_conditions thing in the parent class... We could do this, I suppose. At the moment, it's a bit idiosyncractic because the `initial_conditions` is just a single parameter (we hardcode the other two parameters' initial conditions).

I'm actually thinking it may be better to avoid this ambiguity and set_m0 to make it explicit that we're only referring to one element of the initial conditions. We could then include in the parent class setters / getters for all initial conditions.

pints/toy/_hes1_michaelis_menten.py
MichaelClerx: Can we still use np.zeros above, just to avoid any errors if we forget to set one explicitly?

pints/toy/_hes1_michaelis_menten.py
ben18785: Yes, this is vague (and not my terminology). I may rename and will certainly re-doc!

pints/toy/_hes1_michaelis_menten.py
MichaelClerx: Just a question :D 
Yeah fair enough if they're local methods then this is fine. Docstring could be slightly longer maybe?

pints/toy/_hes1_michaelis_menten.py
MichaelClerx: Thanks!

pints/toy/_hes1_michaelis_menten.py
ben18785: Good spot! (You are eagle-eyed -- I have no idea how you saw that.)

pints/toy/_hes1_michaelis_menten.py
ben18785: Yep, done

pints/toy/_hes1_michaelis_menten.py
ben18785: I've moved these methods to the parent class since there was a lot of ambiguity over what initial_conditions should / would return. In this class, I've put a getter for m0 only since its an estimated parameter of the model.

pints/toy/_hes1_michaelis_menten.py
ben18785: Done

MichaelClerx:
I'm happy for this to be merged @ben18785 . It also 

Closes #732

PR #1230, opened 2020-11-03
Adds noisy gradient options to HMC sampler

MichaelClerx:
Don't see a ticket for this one!

Since we do have O(10) tickets open about HMC variants, I wonder if we should all add them to the base class like this, or if we need lots of subclasses?

Thoughts @ben18785 ? Is this almost ready to go in?


ben18785:
I think this concept would need more thought before including as it essentially introduces a "noisy gradients" class into PINTS, which could behave quite differently. So closing for now.

Issue #1242, opened 2020-12-09
Add Python 3.9 testing


MichaelClerx:
Released in October

PR #1243, opened 2020-12-09
Testing Python3.9 testing

MichaelClerx:
See #1242

### DIFF
--- a/.github/workflows/unit-test-python-coverage.yml
+++ b/.github/workflows/unit-test-python-coverage.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.4, 3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


fcooper8472:
This looks fine.

We should also retire testing for 3.5, which has reached end of life.  New ticket for that, or just add it to this change?

MichaelClerx:
Can we wait for it to break? I'm sure there's still voluntary and involuntary 3.5 users out there in the academic ecosystem

fcooper8472:
Yeah, that's definitely something that we can do.  And that might be appropriate, if there are still systems out there without access to 3.6.

The opportunity cost is not getting to use new features such as f-strings, which will make PINTS code nicer.  Not a huge deal, though.

Why did we drop 3.4?

MichaelClerx:
Good question!

We're still testing 2.7 so f strings are out anyway. But 3.4... I think we just couldn't be bothered to wait for more than 4 unit tests? Maybe we should add it back in now, since the github actions thing seems to be working fairly quickly?

MichaelClerx:
> Error: Version 3.4 with arch x64 not found

Maybe that was why :D

MichaelClerx:
Psssst @fcooper8472 !

fcooper8472 (approved):
Looks good to me!

MichaelClerx:
Thanks!

Issue #1237, opened 2020-11-13
Log-likelihood storage in population MCMC is wrong


MichaelClerx:
This example is too elaborate, but does show it:

If you set `cheap_log_pdfs = True` it uses the log_pdfs returned by (the relatively new function) `mcmc.log_pdfs()`, if you set it to `False` it recalculates them. With `True` many points don't fall on the line they're supposed to.

```
#!/usr/bin/env python3
import pints
import pints.toy as toy
import pints.plot
import numpy as np
import matplotlib
import matplotlib.gridspec
import matplotlib.pyplot as plt
import seaborn
import pandas



#method = pints.HaarioBardenetACMC
method = pints.PopulationMCMC
n_chains = 9
n_samples = 1000

cheap_log_pdfs = True


# Create log pdf
log_pdf = pints.toy.GaussianLogPDF(mean=[0], sigma=[1])

# Run MCMC
xs = np.random.normal(0, 3, size=(n_chains, 1))
mcmc = pints.MCMCController(
    log_pdf, n_chains, xs, method=method)
if cheap_log_pdfs:
    mcmc.set_log_pdf_storage(True)
mcmc.set_max_iterations(n_samples)
#mcmc.set_log_to_screen(False)
print('Running...')
chains = mcmc.run().reshape((n_chains, n_samples))
print('Done!')
if cheap_log_pdfs:
    scores = mcmc.log_pdfs()
else:
    print('Recalculating log pdfs...')
    scores = []
    for i, chain in enumerate(chains):
        print(f'Chain {i + 1}')
        scores.append(log_pdf(chain))

# Create figure
fig = plt.figure(figsize=(9, 10))
fig.subplots_adjust(0.12, 0.08, 0.99, 0.99)
grid = matplotlib.gridspec.GridSpec(4, 6, wspace=0.80, hspace=0.30)

# Determine range for plot
r = np.max(np.abs(chains)) * 1.1

# Evaluate logpdf for plots
x = np.linspace(-r, r, 1000)
y = log_pdf(x)

# Plot PDF
ax1 = fig.add_subplot(grid[0, :])
ax1.set_ylabel('f')
ax1.set_xlim(-r, r)
ax1.axvline(xs[0], color='#dddddd', label='Starting point')
for x0 in xs[1:]:
    ax1.axvline(x0, color='#dddddd')
ax1.plot(x, np.exp(y), color='k')
for i, (chain, score) in enumerate(zip(chains, scores)):
[...]

Issue #660, opened 2019-02-05
Add accepted() and current() to MCMCSampler or update return type of tell()


MichaelClerx:
At various places in the code we need to know more from a sampler than just the current position it is at.
I.E.

- when storing evaluations inside the controller, we need to know the evaluation for the accepted sample (which is not necessarily the last sample that was evaluated)
- inside population MCMC we need to know whether or not a tell() accepted the sample, to update our internal mirror of its state
- for some model selection methods, I've been told

My initial proposal was to add methods `accepted() -> bool` that tell you what happened on the last step, and `current()` that would give the current position + evalution possibly.

Now thinking it might be better to have this as part of tell. E.g.
```
class MCMCSampler
  def tell(fx):
    ...
    return current, current_log_pdf, accepted_yes_or_no
```
or
```
class MCMCSampler
  def tell_verbose(fx)
    ...
    return current, current_log_pdf, accepted_yes_or_no
  def tell(fx)
    return self.tell_verbose(fx)[0]
```
or
```
class MCMCSampler
  def tell(fx, verbose=False)
    if verbose:
      return current, current_log_pdf, accepted_yes_or_no
    return current
```

My preference would be for 1 or 2, I think.

Thoughts on this much appreciated @ben18785 @martinjrobins @chonlei @fcooper8472 

---------------------------------------------------------------------------------------------------------------------------------------------------------

Original message:

*Add accepted() and current() to MCMCSampler*

To get more info about what happened on the last step. This makes it easier to use MCMC samplers inside another method

MichaelClerx:
@martinjrobins ?

We need these e.g. in https://github.com/pints-team/pints/pull/664/files which currently uses hacky methods not guaranteed to work

MichaelClerx:
Might allow `current_log_pdf` to be dropped

MichaelClerx:
Closed in #1250

PR #1250, opened 2021-01-14
Likelihood storage bug for methods like PopulationMCMC

MichaelClerx:
- Add test showing issue #1237
- Implement changes to `MCMCSampler.tell()` #660 
- Fix bug


MichaelClerx:
Quick test shows that data is shared by reference even with co-routines

```
#!/usr/bin/env python3

def test():
    last = [1, 2, 3]
    while True:
        nxt = yield
        print(last, nxt)
        last = nxt

coroutine = test()
next(coroutine)

x = [1, 1, 1]
coroutine.send(x)
x[0] = 9
coroutine.send([2, 3, 4])
```
shows
```
[1, 2, 3] [1, 1, 1]
[9, 1, 1] [2, 3, 4]
```

So should use copy in nuts, I think

MichaelClerx:
@ben18785 could you have a look at the slice stepout method please?

I've update tell() to return the current point and its logpdf (which can be logged by the controller, for various purposes). Only problem here is the point seems to be updated between receiving it's f(x) and being added to the chain?

https://github.com/pints-team/pints/blob/1237-likelihood-storage/pints/_mcmc/_slice_stepout.py#L605-L635

Is this correct?

MichaelClerx:
Hi all!

I've gone ahead and implemented the changes to `tell()` I proposed earlier.
Please can you have a look and see what you think?

(There's an issue with slice stepout that causes the tests to fail, all other tests pass with this code, including new ones which fail with the current code in master)

MichaelClerx:
@chonlei I imagine this impacts the hierarchical code?

MichaelClerx:
Since the functional tests are no longer running automatically, I've just started a script to run 1000 tests on skip, so that we have some data from before this change and after (plus before/after the popmcmc fix, which should be a nice example)

chonlei:
Which hierarchical code? @MichaelClerx

MichaelClerx:
@chonlei The code for IKr fitting that runs hierarchical inference. I imagine that uses the ask-tell interface directly?

ben18785 (changes requested):
Thanks @MichaelClerx -- a lot of work here! Re: slice stepout -- that log_y thing is just a way of avoiding doing accept/reject on the linear scale (opposed to the log scale). (It's on page 712 of the referenced paper if you're interested.)

I think it looks good but (as I said in the PINTS meeting) I think it'd be good to think about how we test functionally that all works after these changes: mainly because they're so wide-sweeping. I don't know how to do this, to be honest, given that functional testing isn't up and running...

One thing also mentioned in the meeting was the notebooks: I suspect some of them use ask/tell so may need to be updated.

MichaelClerx:
> Re: slice stepout -- that log_y thing is just a way of avoiding doing accept/reject on the linear scale (opposed to the log scale). (It's on page 712 of the referenced paper if you're interested.)

Thanks Ben!
How do you propose we change the code so that we can return `x, f(x)` instead of `x, f(some-other-point)` which we have at the moment?

(Thinking ahead, I'd be fine to change the contract so that `x, None` is acceptable, in case we add methods that use x and f(x) to add some untested point x' to the chain. But guessing they might not exist :D )

ben18785:
> > Re: slice stepout -- that log_y thing is just a way of avoiding doing accept/reject on the linear scale (opposed to the log scale). (It's on page 712 of the referenced paper if you're interested.)
> 
> Thanks Ben!
> How do you propose we change the code so that we can return `x, f(x)` instead of `x, f(some-other-point)` which we have at the moment?
> 
> (Thinking ahead, I'd be fine to change the contract so that `x, None` is acceptable, in case we add methods that use x and f(x) to add some untested point x' to the chain. But guessing they might not exist :D )

Not sure I follow? We do return x, f(x) at the moment. The log_y thing is just a convenience thing to determine accept/reject.

MichaelClerx:
Hmmm!

OK I need to check the details then.

This test shows that slice_stepout returns `x, f(not x)`:

https://github.com/pints-team/pints/pull/1250/checks?check_run_id=1710841304

I haven't checked, but I think it may be due to these lines:

https://github.com/pints-team/pints/blob/ed07435cd3dd5e9e4a04416a2f490b5a00a57a13/pints/_mcmc/_slice_stepout.py#L601-L606

that change `proposed` back to `current` (at least in part?)

so that these lines are no longer appropriate:

https://github.com/pints-team/pints/blob/ed07435cd3dd5e9e4a04416a2f490b5a00a57a13/pints/_mcmc/_slice_stepout.py#L620-L625

MichaelClerx:
If I understand correctly the top bit could just say "proposed = current; fx = current_fx" to fix it? Or the method could even stop at this point ("rject") ?

MichaelClerx:
Oh dear, looks like the likelihood storage doesn't work quite as we thought it does in slice stepout @ben18785 !
I can't quite see how, but resetting the param change does not seem to equate to reverting the proposal to the current position

ben18785:
@MichaelClerx Hmm. When you say "doesn't work quite as we thought it", what do you mean? Is the fx value that's returned wrong (i.e. fx != model(x_current))?

MichaelClerx:
@ben18785 from looking at the code we thought that the method had a `current` and `current_log_pdf`, then changed 1 parameter to get a proposal with evaluation `fx`. When we reset this proposal we thought we'd be back at `current` so that `fx = current_log_pdf` would fix our problem. But I tried and it doesn't

MichaelClerx:
Hi @ben18785 ! 

Could you have another look at slice stepout? I'm worried that if we leave this too long it'll get harder to merge this PR

MichaelClerx:
> Hi @ben18785 !
> 
> Could you have another look at slice stepout? I'm worried that if we leave this too long it'll get harder to merge this PR

I could try and dig in myself some time next week, @ben18785 , but not sure if that's a good idea without understanding how the method is supposed to work 😅

ben18785:
Sure -- I'll take a look this week

MichaelClerx:
Thank you!

MichaelClerx:
@ben18785 ping!

pints/_mcmc/_hamiltonian.py
```diff
@@ -355,8 +356,13 @@ def tell(self, reply):
                 self._mcmc_acceptance = (
                     (self._mcmc_iteration * self._mcmc_acceptance + accept) /
                     (self._mcmc_iteration + 1))
-                self._current.setflags(write=False)
-                return self._current
+
+                # Return current state
+                return (
+                    self._current,
+                    (-self._current_energy, -self._current_gradient),
```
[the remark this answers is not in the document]
MichaelClerx: Updated!

MichaelClerx:
@ben18785 shall we merge this now?

(Note coverage is 100%; report above just says 13 lines that were covered in tests have now been removed from the codebase)

MichaelClerx:
@DavAug @chonlei either of you have time to review this? It's a big change, but it fixes a bug so good to get this merged in soon

MichaelClerx:
I've manually instigated a run of tests on skip, so that we have functional testing data from before merging this. Will do that again after we merge, so that we have a chance of picking up any bugs not picked up by the unit tests

pints/_mcmc/__init__.py
```diff
@@ -684,14 +684,14 @@ def run(self):
...
-                xs_iterator = iter(xs)
                 fxs_iterator = iter(fxs)
                 for i in list(active):  # new list: active may be modified
-                    x = next(xs_iterator)
-                    fx = next(fxs_iterator)
-                    y = self._samplers[i].tell(fx)
+                    reply = self._samplers[i].tell(next(fxs_iterator))
+
+                    if reply is not None:
+                        # Unpack
```
chonlei, line 692: Maybe say something like
`# Unpack (current position, evaluated value, acceptance of last evaluated sample)`

pints/_mcmc/__init__.py
```diff
@@ -746,10 +744,13 @@ def run(self):
                 # Multi-chain methods
 
                 # Get all chains samples at once
-                ys = self._samplers[0].tell(fxs)
-                intermediate_step = ys is None
+                reply = self._samplers[0].tell(fxs)
+                intermediate_step = reply is None
 
                 if not intermediate_step:
+                    # Unpack
```
chonlei, line 751: Similarly as above, say what are the unpacked values?

pints/_mcmc/_adaptive_covariance.py
```diff
@@ -115,6 +115,7 @@ def ask(self):
         if not self._running:
             self._running = True
             self._proposed = self._x0
+            self._proposed.setflags(write=False)
```
chonlei, line 118: Why do we want to set write to False here?

MichaelClerx: When we accept the point we do `self._current = self._proposed` (no need to create a new array!), and we then want to return the current point as write only

I think we might have allowed users access to `self._proposed` at some point too

CHANGELOG.md
```diff
@@ -17,12 +17,15 @@ All notable changes to this project will be documented in this file.
 - [#1112](https://github.com/pints-team/pints/pull/1112) A new `NoUTurnMCMC` sampler (NUTS) was added, along with a `DualAveragingAdaption` class to adaptively tune related Hamiltonian Monte Carlo methods.
 - [#1025](https://github.com/pints-team/pints/pull/1025) Added a stochastic logistic growth problem for use with ABC.
 ### Changed
+- [#1250](https://github.com/pints-team/pints/pull/1250) The returned values from `SingleChainMCMC.tell()` and `MultiChainMCMC.tell()` have been extended from current position `x` to `x, fx, accepted`, where `fx` is the current log likelihood and `accepted` is a bool indicating whether tell performed an acceptance step in this call.
```
chonlei, line 20: Still thinking be best to return a dict for `fx` and `accepted`... I don't know what else we'd ever want to return in the tell, but since we *encourage* others to use the ask-tell methods, we probably don't want to make changes to the API. I thought _maybe_ having it as a dict would minimise the chance of changing the API even if we want to add something more to be returned in `tell`...

Because I'm imaging I'd be doing
```
x, fx, accepted = mcmc.tell()
```
and if we add another one output to it, then the code above breaks (not enough value to unpack - as I need `x, fx, accepted, _` to unpack). But if we return a dictionary, then we'd ask others to do something like
```
next = mcmc.tell()
x = next['x']
fx = next['fx']
accepted = next['accepted']
```
and even if we add another key (I don't know what, but e.g.) `fx2` to it, the above code would still work.

Not sure, what do you think?

pints/_mcmc/_adaptive_covariance.py
chonlei: Cool - want to explain that here? In case someone else later wants to use this as a template to write another MCMC method, they'd know why we want to set this to `write=False`.

pints/_mcmc/_differential_evolution.py
```diff
@@ -308,14 +305,15 @@ def tell(self, proposed_log_pdfs):
...
         # Perform iteration
         next = np.array(self._current, copy=True)
         next_log_pdfs = np.array(self._current_log_pdfs, copy=True)
 
         # Sample uniform numbers
-        u = np.log(np.random.uniform(size=self._chains))
+        u = np.log(np.random.uniform(size=self._n_chains))
 
         # Get chains to be updated
         i = u < (proposed_log_pdfs - self._current_log_pdfs)
```
chonlei, line 319: Rename `u` and `i` to something else...?

CHANGELOG.md
MichaelClerx: I really don't want to return dicts from tell.
It would mean lots of unnecessary dict creation and dict lookups, in a method we're expecting to call 10000s of times.

And the fact that `x, fx, accepted = ...` would break is a feature. It means we have a rigid interface that people need to stick to, so that everyone knows what to expected. Otherwise we need to make the whole "contract" for using ask-tell a much softer thing that the user needs to check instead of having the programming language check it.

To me those downsides are much bigger than the possible benefit of having some future extension of `tell()` being slightly easier.

MichaelClerx: And we've found that `x`, `fx`, and `accepted` are used by the controller and in some meta-methods, so there's a clear case for them being there.
And we've found that they exist and make sense for each method so far, so a good case for having them being solid variables rather than things that might exist in a dict

pints/_mcmc/_differential_evolution.py
MichaelClerx: Separate ticket :D

pints/_mcmc/_nuts.py
```diff
@@ -742,7 +737,11 @@ def tell(self, reply):
             self._next = next(self._nuts)
 
             # Return current position as next sample in the chain
-            return self._current
+            return (
+                np.copy(self._current),
+                (current_logpdf, np.copy(current_gradient)),
+                True
```
chonlei, line 743: I forgot how NUTS works, does it always accept?

pints/_mcmc/_population.py
```diff
@@ -138,9 +134,10 @@ def _initialise(self):
         # Next chain to exchange / crossover with
         self._j = 1
 
-        # Ask all inner samplers for first point (should be x0, so ignore!)
+        # Ask all inner samplers for first point, which should be x0
         for chain in self._chains:
-            chain.ask()
+            x0 = chain.ask()
+            assert(np.all(x0 == self._x0))
```
chonlei, line 140: Why this assert statement here?

pints/_mcmc/_population.py
```diff
@@ -275,16 +272,17 @@ def tell(self, fx):
...
-        sample = self._chains[self._i].tell(fx * (1 - self._schedule[self._i]))
+        sample, sample_log_pdf, accepted = self._chains[self._i].tell(
+            fx * (1 - self._schedule[self._i]))
 
-        # Check if accepted
-        if np.any(sample != self._current[self._i]):
-            # Update current sample and untempered log pdf
+        # Update current sample and untempered log pdf
+        if accepted:
+            # Don't use sample_log_pdf here, which is tempered
```
chonlei, line 285: Was this the bug?

chonlei: Or it was https://github.com/pints-team/pints/pull/1250/files#diff-dd2521d7bad8361241509fcc5108df3add9b81e36c9d78cb281894a2f4898d7fR329-R330?

pints/_mcmc/_population.py
```diff
@@ -327,10 +325,12 @@ def tell(self, fx):
...
 
+        # Return "accepted" any time the zero-chain moves
+        changed = accepted and self._i == 0
+        changed |= self._have_exchanged and (self._i == 0 or self._j == 0)
+
         # Return new point for chain 0
-        sample = np.array(self._current[0], copy=False)
-        sample.setflags(write=False)
-        return sample
+        return np.copy(self._current[0]), self._current_log_pdfs[0], changed
```
chonlei, line 333: Why do we return a copy of `self._current[0]` instead?

pints/_mcmc/_relativistic.py
```diff
@@ -351,11 +347,16 @@ def tell(self, reply):
             # Increase iteration count
             self._mcmc_iteration += 1
 
-            # Mark current as read-only, so it can be safely returned
+            # Mark current as read-only, so it can be safely returned.
+            # Gradient won't be returned (only -gradient, so no need.
```
chonlei, line 351: Although it's just the comment, an incomplete parenthesis.

pints/_mcmc/_relativistic.py
```diff
@@ -413,7 +418,8 @@ def tell(self, reply):
                     self._current_energy = energy
                     self._current_gradient = gradient
 
-                    # Mark current as read-only, so it can be safely returned
+                    # Mark current as read-only, so it can be safely returned.
+                    # Gradient won't be returned (only -gradient, so no need.
```
chonlei, line 422: Same as above.

pints/tests/test_mcmc_monomial_gamma_hmc.py
```diff
@@ -34,22 +34,26 @@ def test_method(self):
...
-                self.assertEqual(mcmc.current_log_pdf(), fx)
+            reply = mcmc.tell((fx, gr))
+            if reply is not None:
+                y, fy, ac = reply
+                if i >= 25 * ifrog:
+                    chain.append(y)
+                if ac:
+                    self.assertTrue(np.all(x == y))
+                    self.assertEqual(fx, fy[0])
+                    self.assertTrue(np.all(gr == fy[1]))
```
chonlei, line 54: Do we want to check `ac` is a boolean?

pints/tests/test_mcmc_nuts.py
```diff
@@ -41,13 +41,14 @@ def test_method(self):
...
-            if sample is not None:
-                chain.append(sample)
-            if np.all(sample == x):
-                self.assertEqual(mcmc.current_log_pdf(), fx)
+            if reply is not None:
+                y, fy, ac = reply
+                chain.append(y)
+                recalc = log_pdf.evaluateS1(y)
+                self.assertEqual(fy[0], recalc[0])
+                self.assertTrue(np.all(fy[1] == recalc[1]))
```
chonlei, line 52: Similar check `ac` is a boolean? (And for the rest too.)

CHANGELOG.md
chonlei, line 20: Do we want to say we've changed the version to 0.4.0 here as well...?

chonlei (approved):
Thanks Michael! I have added some minor comments, feel free to address or ignore, otherwise I am happy for it to be merged.

pints/_mcmc/_nuts.py
MichaelClerx: No, but this is the bit in the code where it accepts :D

CHANGELOG.md
MichaelClerx: Maybe!
But I think we release version 0.4.0, and then people go look to see what's new, so then they don't need to see "it's 0.4.0 now" again :D

pints/_mcmc/_population.py
MichaelClerx: There were two bugs in pop mcmc itself, both fixed here: https://github.com/pints-team/pints/pull/1246/files

This ticket solves a separate but related bug: the controller couldn't store the accepted sample's log likelihood without these changes to tell

pints/_mcmc/_population.py
MichaelClerx: self._current holds samples for several (tempered) chains, but you only want the samples from chain 0

pints/_mcmc/_population.py
MichaelClerx: Removed it!

pints/tests/test_mcmc_monomial_gamma_hmc.py
MichaelClerx: ok

MichaelClerx:
Thanks Chon!

MichaelClerx:
@ben18785 can you hit approve as well so I can merge and re-run the functional tests?

ben18785:
Sure. Is it easier to merge then run these rather than run then merge?

MichaelClerx:
Yes

ben18785 (approved):

ben18785:
@MichaelClerx Done

MichaelClerx:
🥳

MichaelClerx:
Running fc now

PR #1257, opened 2021-01-25
Corrects instantiation bug for GaussianLogPrior

ben18785:


pints/tests/test_log_priors.py
```diff
@@ -441,6 +441,10 @@ def test_gaussian_prior(self):
         p = pints.NormalLogPrior(mean, std)
         self.assertIsInstance(p, pints.GaussianLogPrior)
 
+        # Test assertRaises with negative sd
```
fcooper8472, line 444: think this should technically say non-positive

fcooper8472 (approved):
Looks good to me

PR #1261, opened 2021-01-26
Ensure all github actions have nice job names

MichaelClerx:
Hi @ben18785 ! This should make it easier for us to see which tests are which


ben18785 (approved):
Looks good! Thanks @MichaelClerx

Issue #1198, opened 2020-08-25
check NUTS behaviour if a NaN gets returned


martinjrobins:
there has been some problems with the Nuts sampler for models returning nans, see #1112. There are some checks for this that have solved the immediate problems, but need to check this more thoroughly.

MichaelClerx:
See #794 and #251

ben18785:
Victor (whom Martin and I are supervising) has resolved this and is going to create a PR.

ben18785:
Closed by #1264

PR #1264, opened 2021-01-26
Added divergence detection for NUTS in the build_tree function

phumtutum:
See #1198


ben18785 (changes requested):
Thanks @phumtutum Can we add a test for this using the Pints logistic model with strange parameter values? Basically, it'd be good to check that the sampler now executes...

ben18785 (approved):
Thanks @phumtutum -- looks good to me.

PR #1278, opened 2021-02-08
Fix small typo in transformation docs

### LINKED ISSUE #1277: Typo in RectangularBoundariesTransformation docs
https://github.com/pints-team/pints/blob/a6a1a39c322c8419ff929b82f73ed3f7aa8b72ef/pints/_transformation.py#L708-L723

chonlei:
Fix #1277

### DIFF
--- a/pints/_transformation.py
+++ b/pints/_transformation.py
@@ -718,9 +718,9 @@ class RectangularBoundariesTransformation(Transformation):
                  = \log(p - a) - \log(b - p),
 
     where :math:`p` is the model parameter vector and :math:`q` is the
-    search space vector. The range includes the lower (:math:`a`), but not the
-    upper (:math:`b`) boundaries. Note that :class:`LogitTransformation` is a
-    special case where :math:`a = 0` and :math:`b = 1`.
+    search space vector. The range does not include the boundaries (:math:`a`
+    and :math:`b`). Note that :class:`LogitTransformation` is a special case
+    where :math:`a = 0` and :math:`b = 1`.
 
     The Jacobian adjustment of the transformation is given by
 
--- a/pints/_transformation.py
+++ b/pints/_transformation.py
@@ -708,13 +708,13 @@ class LogTransformation(Transformation):
 class RectangularBoundariesTransformation(Transformation):
     r"""
     A generalised version of the logit transformation for the model parameters,
-    which transform an interval or rectangular boundaries :math:`[a, b)` to
+    which transform an interval or rectangular boundaries :math:`(a, b)` to
     all real number.
 
     The transformation is given by
 
     .. math::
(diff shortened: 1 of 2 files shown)


pints/_transformation.py
```diff
@@ -708,13 +708,13 @@ def to_search(self, p):
 class RectangularBoundariesTransformation(Transformation):
     r"""
     A generalised version of the logit transformation for the model parameters,
-    which transform an interval or rectangular boundaries :math:`[a, b)` to
+    which transform an interval or rectangular boundaries :math:`(a, b)` to
```
MichaelClerx, line 711: transform -> transforms.

Is it a () interval then, not a half-open one?

MichaelClerx:
Tests are failing. I see Ben has a fix here: https://github.com/pints-team/pints/pull/1058/files

Maybe stick this fix in this PR too? Makes more sense here :D

pints/_transformation.py
chonlei: () is an interval, yes.

chonlei:
Michael did you mean stick this fix (https://github.com/pints-team/pints/pull/1058/files#diff-43f514bb09c1eb58423de94253a01bf25b3232f4324e4a9d8606ee6db00039b5R160) to here?

pints/_transformation.py
MichaelClerx: But you're saying the interval doesn't include `a` then? So it's an open rather than a half-open interval?

MichaelClerx:
> Michael did you mean stick this fix (https://github.com/pints-team/pints/pull/1058/files#diff-43f514bb09c1eb58423de94253a01bf25b3232f4324e4a9d8606ee6db00039b5R160) to here?

Yes please!

pints/_transformation.py
chonlei: Ha, need correcting that line too! :smile:

chonlei: It's an open () interval. I think now the docs is consistent...?

MichaelClerx: :D Are you sure? The rectangular boundaries docs say they define a half-open interval https://pints.readthedocs.io/en/stable/boundaries.html#pints.RectangularBoundaries

chonlei: Oh... didn't realise that!

chonlei: I can't remember the reason - why do we want to include the lower bound?

MichaelClerx: Not sure, but maybe a separate ticket then? This can just fix the missing "s" and test :D :D :D

chonlei: See #1279.

MichaelClerx (approved):

Issue #1101, opened 2020-04-06
Better docs and explanation for diagnostics


chonlei:
Maybe we could have slightly better docs for this page:
https://pints.readthedocs.io/en/latest/diagnostics.html#mcmc-diagnostics?

For example what are the expected inputs and outputs. It took me a few minutes to figure it out, I suspect it'd be harder for other users...

MichaelClerx:
Is "diagnosting" really a verb?

PR #1282, opened 2021-02-09
Documentation (and naming?) changes in pints.diagnostics: redux

ben18785:
See #1101 

- [x] Update docstring for autocorrelate_negative: I don't understand what it does?
- [x] Deprecate ess_single_param in favour of effective_sample_size_single_parameter, as we don't use "param" anywhere else in PINTS names, and the next method in the same file is called effective sample size ?
- [x] Haven't made effective_sample_size_single_parameter private as I can imagine wanting to use it

### DIFF
--- a/examples/sampling/effective-sample-size.ipynb
+++ b/examples/sampling/effective-sample-size.ipynb
@@ -355,7 +355,7 @@
    "source": [
     "import pints._diagnostics as diagnostics\n",
     "\n",
-    "print('est. dependent (rho = 0.8) ESS = ' + str(diagnostics.ess_single_param(lXLowerRho)))"
+    "print('est. dependent (rho = 0.8) ESS = ' + str(diagnostics.effective_sample_size_single_parameter(lXLowerRho)))"
    ]
   }
  ],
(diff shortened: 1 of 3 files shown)


MichaelClerx:
Thanks @ben18785 !

```
>>> def _autocorrelate_negative(autocorrelation):
...     T = 1
...     for a in autocorrelation:
...         if a < 0:
...             return T - 1
...         T += 1
...     return T
... 
>>> _autocorrelate_negative([0, 1, 2, 3, 4, 5, 6])
8
>>> _autocorrelate_negative([0, 1, 2, 3, 4, -5, -6])
5
```

I'm thinking the 8 should be a 7? Have updated code & docstring accordingly

MichaelClerx:
```
def _autocorrelate_negative(autocorrelation):
    """
    Returns the index of the first negative entry in ``autocorrelation``, or
    ``len(autocorrelation)`` if no negative entry is found.
    """
    try:
        return np.where(np.asarray(autocorrelation) < 0)[0][0]
    except IndexError:
        return len(autocorrelation)

>>> _autocorrelate_negative([0, 1, 2, 3, 4, 5, 6])
7
>>> _autocorrelate_negative([0, 1, 2, 3, 4, -5, -6])
5
```

- less readable
+ faster

ben18785:
Great: no need for it to be readable so long as it does what it needs to do.

MichaelClerx (approved):

MichaelClerx:
THanks @ben18785 !

PR #1287, opened 2021-02-10
Correct wrong calls to MultimodalGaussianLogPDF in notebook

ben18785:


ben18785:
Closing since #1288 now covers this

Issue #1286, opened 2021-02-10
Check for non-symmetric matrices used in MultimodalNormalLogPDF across PINTS


ben18785:
related to #498 (we're not going to test for symmetric cov matrices).

populationMCMC notebook has a bug I think, but need to check for others

ben18785:
Think we may now add these tests but still need to fix notebooks / other places where this is wrongly called

ben18785:
Closed by #1288

Issue #498, opened 2018-09-03
Make sure MultimodalNormalLogPDF throws an error when supplying non-posdefinite cov matrix


ben18785:
The populationMCMC notebook has this, I think.

MichaelClerx:
If it does that'll be from the underlying scipy / numpy code then (so not something we need to test)

ben18785:
I don't understand why the below returns a non-zero number:

```
dist = scipy.stats.multivariate_normal(mean=[16, 12], cov=[[0.8, 0.2], [0.1, 1.4]])
dist.pdf([3, 4])
```

here `cov` is not symmetric, so does not satisfy, "The covariance matrix cov must be a (symmetric) positive semi-definite matrix" in the scipy docs...

This looks like a bug in scipy to me! So, we need to test.

ben18785:
Wrote a bug issue on Scipy: https://github.com/scipy/scipy/issues/13541

MichaelClerx:
I'm happy for us to stick in a test that the matrix must be symmetric, although we'll run into the precision issues mentioned by the scipy devs.

ben18785:
It's fine for us not to test it in that case (I do think the scipy
reasoning is a bit weak though: if you're afraid of numerical round off,
fine, but most other libraries manage to find a sensible middle ground).
Although we need to make sure that we use the method properly, so need to
fix the populationMC notebook (and more generally check
wherever MultiModalLogPDF is called). I'll create a separate ticket for
this though and close this one.

ben18785:
Actually, I'm now less sure: we _do_ have such a test in the GaussianLogPDF, so I'm just going to grab this from there.

PR #1288, opened 2021-02-10
Covariance matrix tests for multimodalGaussianLogPDF

ben18785:
- added tests
- corrected tests where we were using incorrect cov matrices
- corrected incorrect call in notebook


ben18785:
addresses #498 and #1286

MichaelClerx (approved):

PR #1295, opened 2021-02-16
#1294 functional test HaarioBardenet

### LINKED ISSUE #1294: Create a functional testing module with a module for the HarrioBardenerACMC tests
The module will contain the functional tests in pints, e.g. one file for each sampler and each optimiser. We start with the HarrioBardenerACMC.

DavAug:
Closes #1294 .

### DIFF
--- a/pints/functionaltests/functest_haario_bardenet_acmc.py
+++ b/pints/functionaltests/functest_haario_bardenet_acmc.py
@@ -1,0 +1,75 @@
+import numpy as np
+
+import pints
+import pints.toy
+
+
+class TestHaarioBardenetACMCOn2dimGaussianDistribution(object):
+    """
+    Tests the pints.HaarioBardenetACMC on a standard 2 dimensional Gaussian
+    distribution.
+    """
+    def __init__(self):
+        # Define calibrated test parameters
+        n_chains = 3
+        n_iterations = 4000
+        warmup = 1000
+        method = pints.HaarioBardenetACMC
+
+        # Define pdf
+        self.pdf = pints.toy.GaussianLogPDF(mean=[0, 0], sigma=[1, 1])
+
+        # Get initial parameters
+        log_prior = pints.ComposedLogPrior(
+            pints.GaussianLogPrior(mean=0, sd=100),
+            pints.GaussianLogPrior(mean=0, sd=100))
+        initial_parameters = log_prior.sample(n=n_chains)
+
+        # Set up sampler
+        sampler = pints.MCMCController(
+            self.pdf, n_chains, initial_parameters, method=method)
+        sampler.set_max_iterations(n_iterations)
+        sampler.set_log_to_screen(False)
+
+        # Infer posterior and throw away warm-up
+        chains = sampler.run()
+        self.chains = chains[:, warmup:]
+
+    def estimate_kld(self):
+        """
+        Estimates and returns the Kullback-Leibler divergence between the
+        approximate posterior and the true posterior assuming that the
+        approximated posterior has Gaussian shape.
+        """
+        # Pool samples from chains
+        chains_x = self.chains[:, :, 0].flatten()
+        chains_y = self.chains[:, :, 1].flatten()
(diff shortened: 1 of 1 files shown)


DavAug:
Something like this @fcooper8472 ?

pints/functionaltests/functest_haario_bardenet_acmc.py
```diff
@@ -0,0 +1,91 @@
...
+import pints
+import pints.toy
+
+
+class TestHaarioBardenetACMCOn2dimGaussianDistribution(object):
+    """
+    Tests the pints.HaarioBardenetACMC on a standard 2 dimensional Gaussian
+    distribution.
+    """
+    def __init__(self):
```
MichaelClerx, line 21: Not sure why this is 4 methods?

Would it make sense to have a single method that returns `name, results` ? Then we could have some class for the TestHaarioBardenetACMC where each method is a different test?

MichaelClerx:
Thanks both! Looks like it would work to me, which is the most important thing :D

pints/functionaltests/_problems.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        # Estimate mean ESS for each chain
+        n_chains, _, n_parameters = self.chains.shape
+        ess = np.empty(shape=(n_chains, n_parameters))
+        for chain_id, chain in enumerate(self.chains):
+            ess[chain_id] = pints.effective_sample_size(chain)
+
+        return np.mean(ess)
+
+    def estimate_distance(self):
```
MichaelClerx, line 48: Why doesn't this method live in `RunMcmcMethodOnAnnulus` ?

pints/functionaltests/_problems.py
```diff
@@ -0,0 +1,203 @@
...
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+
+import numpy as np
+import pints
+import pints.toy
+
+
+class RunMcmcMethodOnProblem(object):
```
MichaelClerx, line 12: I suppose that if these methods are going to live in the main PINTS repo they will need to have full docstrings and be listed in the docs etc?

MichaelClerx: Placing them inside the `pints` module also means that they'll get distributed as part of the package. Is that something we want?

pints/functionaltests/differential_evolution.py
```diff
@@ -0,0 +1,79 @@
...
+
+import pints
+
+from ._problems import (RunMcmcMethodOnTwoDimGaussian,
+                        RunMcmcMethodOnBanana,
+                        RunMcmcMethodOnCorrelatedGaussian,
+                        RunMcmcMethodOnAnnulus)
+
+
+def test_differential_evolution_on_two_dim_gaussian(n_iterations=None):
```
MichaelClerx, line 18: Would it be a good idea to name these methods as `two_dim_gaussian` etc. so that we can write `pints.functionaltests.differential_evolution.two_dim_gaussian()` ?

pints/functionaltests/_problems.py
MichaelClerx: No just needs a better name

pints/cptests/differential_evolution_mcmc.py
```diff
@@ -0,0 +1,88 @@
...
+    :class:`pints.cptests.RunMcmcMethodOnAnnulus`.
+    """
+    problem = cpt.RunMcmcMethodOnAnnulus(
+        _method, 10, n_iterations, n_warmup)
+    return {
+        'distance': problem.estimate_distance(),
+        'mean-ess': problem.estimate_mean_ess()
+    }
+
+
```
YongchaoHuang, line 81: # TODO: add more test cases, e.g. Annulus

MichaelClerx:
@DavAug @fcooper8472 Any idea what's making the changepoint testing not work?

PR #1301, opened 2021-02-23
Skip draft PRs, but run when a draft is marked as ready

MichaelClerx:
Hi both!

I think I've worked out the full syntax to do this?

- The first line is new, and updates the list of "events" on a PR that trigger checks
  https://docs.github.com/en/enterprise-server@3.0/actions/reference/events-that-trigger-workflows
  It adds "ready_to_review" to the 3 default events, which means that going from draft to normal PR now triggers checks

- The second line is the one Ben added, & which Fergus correctly moved to the job. (I don't understand why, it looks more natural in the above bit to me, but that doesn't work!)

### DIFF
--- a/.github/workflows/copyright-test.yml
+++ b/.github/workflows/copyright-test.yml
@@ -5,7 +5,8 @@ on:
     branches:
       - master
   pull_request:
-    if: github.event.pull_request.draft == false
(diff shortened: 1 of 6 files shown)


MichaelClerx:
Closing this, so I can demo :D

PR #1303, opened 2021-02-23
Skip draft PRs, but run when a draft is marked as ready for review (part 3)

MichaelClerx:
Hi both!

I think I've worked out the full syntax to do this?

- The first line is new, and updates the list of "events" on a PR that trigger checks
  https://docs.github.com/en/enterprise-server@3.0/actions/reference/events-that-trigger-workflows
  It adds "ready_to_review" to the 3 default events, which means that going from draft to normal PR now triggers checks

- The second line is the one Ben added, & which Fergus correctly moved to the job. (I don't understand why, it looks more natural in the above bit to me, but that doesn't work!)

**Update** forgot to create the first PR as draft to show that this works

### DIFF
--- a/.github/workflows/copyright-test.yml
+++ b/.github/workflows/copyright-test.yml
@@ -5,7 +5,8 @@ on:
     branches:
       - master
   pull_request:
-    if: github.event.pull_request.draft == false
+    # Run workflow when PR is changed, or when changed from draft to ready
+    types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - '**'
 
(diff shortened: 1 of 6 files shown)


MichaelClerx:
3d time lucky... that "draft" button is annoyingly hard to hit!

MichaelClerx:
@ben18785 @fcooper8472 once you've both seen this, can one of you hit the "Ready for review" button to trigger the checks?

MichaelClerx:
@ben18785 @fcooper8472 Sorry for the flurry earlier 😄 
But if you could have a look at this one that'd be nice :-)

ben18785:
Thanks @MichaelClerx -- hitting "Ready to review" seemed to trigger it. I'm happy to merge if it seems to work for you. It's a bit hard to do trial and error on these without seeing it in action.

ben18785 (approved):

Issue #1305, opened 2021-03-02
PyStan interface test raises 9000 lines of warnings


MichaelClerx:
This makes the unit test output rather difficult to read

MichaelClerx:
I can try and suppress these?

ben18785:
Happy for you to suppress these: they're just compiler warnings I think.

MichaelClerx:
OK, the easy method (capturing python's stdout stream) doesn't work. So back to you!

MichaelClerx:
OK I'm trying a dirtier version of capturing now, seems to work

MichaelClerx:
Closed in #1306

PR #1306, opened 2021-03-02
Capturing pystan test output

MichaelClerx:
See #1305 

Makes the unit test output readable again.

~Although possibly harder to debug now if things go wrong.~
Have now made it so that it prints all if an error occurs

### DIFF
--- a/pints/tests/shared.py
+++ b/pints/tests/shared.py
@@ -26,6 +26,8 @@ class StreamCapture(object):
     """
     A context manager that redirects and captures the output stdout, stderr,
     or both.
+    
+    Warning: This class is not thread-safe (or multiprocessing-safe).
     """
     def __init__(self, stdout=True, stderr=False):
         super(StreamCapture, self).__init__()
@@ -146,6 +148,8 @@ class SubCapture(object):
 
     The argument ``dump_on_error`` can be set to ``True`` to print all output
     if an error occurs while the context manager is active.
+    
+    Warning: This class is not thread-safe (or multiprocessing-safe).
     """
     def __init__(self, dump_on_error=False):
         super(SubCapture, self).__init__()
--- a/pints/tests/shared.py
+++ b/pints/tests/shared.py
@@ -26,7 +26,7 @@ class StreamCapture(object):
     """
     A context manager that redirects and captures the output stdout, stderr,
     or both.
-    
+
     Warning: This class is not thread-safe (or multiprocessing-safe).
     """
     def __init__(self, stdout=True, stderr=False):
@@ -148,7 +148,7 @@ class SubCapture(object):
 
     The argument ``dump_on_error`` can be set to ``True`` to print all output
     if an error occurs while the context manager is active.
-    
+
     Warning: This class is not thread-safe (or multiprocessing-safe).
     """
     def __init__(self, dump_on_error=False):
--- a/pints/tests/shared.py
+++ b/pints/tests/shared.py
@@ -136,6 +136,159 @@ class StreamCapture(object):
         return self._stderr_captured    # Could be None
 
 
+class SubCapture(object):
+    """
+    A context manager that redirects and captures the standard and error output
+    of the current process, using low-level file descriptor duplication.
+
+    This can be useful to capture output that comes from C extensions, e.g. in
+    the interface classes.
+
+    The argument ``dump_on_error`` can be set to ``True`` to print all output
+    if an error occurs while the context manager is active.
+    """
+    def __init__(self, dump_on_error=False):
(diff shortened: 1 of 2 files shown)


MichaelClerx:
@ben18785 mind merging this before I make further changes?

ben18785:
I've had a look and don't really understand what this all does: how about
you make one PR that fixes the output of the Stan file and tries to fix the
dictionary thing? Then perhaps on Tuesday you can explain it to me? Or we
get another person to review it as well as me further down the line?

ben18785:
Or can chat through it now if you like? Whatever's easiest for you.

MichaelClerx:
Usually we can catch printed output by redirecting python's `stdout` and `stderr` "streams" to some temporary string buffer. This is what the StreamCapture we already have in testing does.
But pystan compiles it's own C++ code, and when that prints it doesn't go through python's streams, so we need a different method to catch it. This is what the SubCapture class does, using some magic with file descriptors and stuff that I partially understood when I wrote it, and partially copied from a variety of tutorials.

So what happens in the pystan interface tests now is that a SubCapture is created before the operations that print to screen, and this catches all output. If there's an error I print that output to screen again, in case it helps in debugging. That's about it :D

MichaelClerx:
The functional bits of the SubCapture code were copied from Myokit, btw, so it has been tested for a few years.

I've not added in any tests for it in pints as it's inside the `tests` directory, which is not tested. (We don't test the tests)

MichaelClerx:
The only things I'm not too sure about are 

- whether the error handling code in the pystan test works - I should probably try this out?
- whether we should capture output in the test and leave it enabled by default, all 1000s of line, or whether we should move the SubCapture (and StreamCapture?) classes into pints main, add tests, and capture the pystan printed output all the time. This would stop it interfering with out controller log output, but could hide stuff from the user that stan devs didn't hide - maybe for a reason?

MichaelClerx:
@fcooper8472 and @martinjrobins any thoughts on this? Code look ok? And should we suppress extension-module output only in the test, or do it in the main method?

ben18785:
Thanks @MichaelClerx for the explanation. It sounds good to me! But, probs best to get feedback from a C++sy person rather than me!

MichaelClerx:
It's this one @fcooper8472

fcooper8472:
@MichaelClerx the code looks fine to me.  What I'm not clear on is why it would be in pints/tests?  It sure doesn't look like it's a test to me!

Is this functionality that you would expect someone to be able to use from PINTS?

MichaelClerx:
Thanks @fcooper8472 , and good question too!

I can think of a couple of different ways to answer it

1. The `SubCapture` class is used only by tests, so goes with the other ThingsThatAreUsedByTestsButAreNotThemselvesTests in `tests/shared.py`.
2. There shouldn't be any ThingsThatAreUsedByTestsButAreNotThemselvesTests, instead these should be part of the (public or private) PINTS API, and have tests of their own
3. There might be some ThingsThatAreUsedByTestsButAreNotThemselvesTests, but this particular class should be part of the (public or private) PINTS API, so that it can be used inside the stan interface, and not by the test

I can be swayed either way in under a minute.

@ben18785 do you have a preference?

MichaelClerx:
Discussed in meeting, easiest to leave it here for now and move later if needed

fcooper8472:
Also discussed that @martinjrobins will take a quick look.  But I'm happy with it.

MichaelClerx:
Thanks Fergus!

MichaelClerx:
One thing I've taken from @martinjrobins experience is that we should slap a label on the class saying it's not thread safe.
Unless we can make `_start_capturing` atomic somehow? But again, easiest route is to slap a warning on and not worry 😅

martinjrobins (changes requested):
this looks in general great @MichaelClerx. As I mentioned in the meeting, my main worry is that there might be different instances of `SubCapture`, maybe on different threads, both calling `_start_capturing` and `_stop_capturing`. You are currently assuming that `_stop_capturing` is always called after `_start_capturing`. Can you use a `Lock` (https://docs.python.org/3/library/threading.html#threading.Lock.acquire) to make sure that only one `SubCapture` is between `_start_capturing` and `_stop_capturing` at a time? And can you do the same in myokit :)

martinjrobins:
> One thing I've taken from @martinjrobins experience is that we should slap a label on the class saying it's not thread safe.
> Unless we can make `_start_capturing` atomic somehow? But again, easiest route is to slap a warning on and not worry

I'd be happy with a not thread safe warning as well :)

martinjrobins (approved):
great, happy with that

PR #1309, opened 2021-03-03
Stan simplified (i.e. sans dictionary stuff)

ben18785:
- removed dictionaries from Stan interface as they're no longer needed (since I made a change a long while ago that removed the transformation stuff from within the Stan interface itself)
- added a note in the notebook about the warnings and also showed how to print the names in a Stan model


MichaelClerx (approved):
Cool!
I've added a few further tweaks to the code and docstrings. Looks good to me

Issue #1312, opened 2021-03-04
Create a .sampler() method for nested controller


ben18785:
At the moment, setting hyperparameters through the controller uses the internal ._sampler method, which is not ideal.

PR #1313, opened 2021-03-04
Created sampler() method within NestedController

ben18785:
- created new getter
- used getter in notebooks and test files

Addresses #1312


pints/_nested/__init__.py
```diff
@@ -667,6 +667,10 @@ def run(self):
 
         return self._m_posterior_samples
 
+    def sampler(self):
+        """ Returns sampler used by the controller. """
```
rcw5890, line 671: What do you think of including this specific comment, which is similar to the sampler() docstring on MCMC controller?
```suggestion
        """Returns the underlying :class:`NestedSampler` object."""
```

ben18785: Good point. I'll incorporate this into my new PR about multinest.

ben18785:
Closing as out of date due to #1034

PR #1317, opened 2021-03-09
Fixed Neal's Funnel Derivative bug

phumtutum:
The derivative of Neal's Funnel is now returned(in the evaluateS1 function) as a numpy array which makes it usable with NUTS.


ben18785 (approved):
Looks good -- nice find @phumtutum

PR #1319, opened 2021-03-13
MultiNest

ben18785:
Created MultiNest sampler and simplified ellipsoidal nested sampling by creating an Ellipsoid class shared by both samplers. Things done.


pints/_nested/__init__.py
```diff
@@ -780,13 +795,191 @@ def _update_logger(self):
...
 
                 # Choose next logging point
                 if self._i_message > self._message_warm_up:
                     self._next_message = self._message_interval * (
                         1 + self._i_message // self._message_interval)
+
+
+class Ellipsoid():
+    """
+    A class to represent N dimensional ellipsoids, which are needed by both
```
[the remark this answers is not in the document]
MichaelClerx: Or should it be a maths N, and maths again below?

pints/_nested/__init__.py
```diff
@@ -780,13 +795,191 @@ def _update_logger(self):
...
+        # calculate useful quantities
+        self._A_inv = np.linalg.inv(A)
+        # don't calculate volume unless needed
+        self._volume = None
+
+        # don't cache points unless constructed using minimum_volume_ellipsoid
+        self._points = None
+        self._n_points = 0
+
+    def __add__(self, other):
```
MichaelClerx, line 857: You're defining a very unusual algebra here, where x + y is an object outside of the space x and y live in...

MichaelClerx: Also... is `[a, b]` too much effort to write versus `a + b` ?

pints/_nested/__init__.py
```diff
@@ -780,13 +795,191 @@ def _update_logger(self):
...
+        return [self, other]
+
+    def add_points(self, points):
```
MichaelClerx, line 860: Looks to be setting rather than adding?

pints/_nested/__init__.py
```diff
@@ -780,13 +795,191 @@ def _update_logger(self):
...
+        """ Returns centroid of ellispoid. """
+        return self._c
+
+    def enlarge(self, enlargement_factor):
+        """ Enlarges ellipsoid by a factor. """
+        self._A *= (1 / enlargement_factor)
+        self._A_inv *= enlargement_factor
+        self._volume = None
+
+    @staticmethod
```
MichaelClerx, line 875: Why is this not just a method mahalanobis_distance(self, point, c) ?
Or even mahalanobis_distance(self, point), with c replaced by self._c in the equation?

pints/_nested/__init__.py
```diff
@@ -780,13 +795,191 @@ def _update_logger(self):
...
+        for i in range(len(points)):
+            dist[i] = Ellipsoid.mahalanobis_distance(points[i], cov_inv, c)
+        enlargement_factor = np.max(dist)
+        A = (1.0 / enlargement_factor) * cov_inv
+        obj = cls(A, c)
+        obj.add_points(points)
+        return obj
+
+    def n_points(self):
+        """ Returns number of points within bounding ellipsoid. """
```
MichaelClerx, line 902: the

(few more of these)

pints/_nested/__init__.py
```diff
@@ -414,6 +419,8 @@ def _initialise_logger(self):
             self._logger.add_time('Time m:s')
             self._logger.add_float('Delta_log(z)')
             self._logger.add_float('Acceptance rate')
+            if self._sampler._multiple_ellipsoids:
```
MichaelClerx, line 422: In the other controllers we handle this in a more general way, by letting each sampler/optimiser define a method _log_init (that does nothing by default) which can update the logging

https://github.com/pints-team/pints/blob/master/pints/_mcmc/__init__.py#L623-L624

and `_log_write`:

https://github.com/pints-team/pints/blob/master/pints/_mcmc/__init__.py#L820-L821

MichaelClerx (changes requested):
Hi Ben!
Some minor comments on the first few files. Will look at the rest at some later point in time :-)

examples/sampling/nested-multinest.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "n = 400\n",
+    "log_pdf = pints.toy.AnnulusLogPDF()\n",
```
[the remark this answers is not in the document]
ben18785: Thanks. Yep, I think I just need to explain this better to be honest. The Gaussian distribution comes in as a prior but I hadn't explained this before.

examples/sampling/nested-multinest.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+    "Now we'll fit an ellipsoid tree to the data, which provides an ellipsoidal decomposition of the domain covered by the sample draws."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pints._nested.__init__ import Ellipsoid\n",
```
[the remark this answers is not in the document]
ben18785: Yeah, I wasn't sure this. The notebook using them for pedagogical purposes (although as your comment above suggests, these bits need to be improved), but, for running MultiNest, they're not necessary. That's why I kept them non-public... I'm happy for them to be public but I just wasn't sure about their utility for a user?

pints/_nested/__init__.py
ben18785: Yep, but I wasn't sure how to handle combining lots of ellipsoids. Specifically, I wasn't sure how to make `leaf_ellipsoids` work using lists. I also thought it has a nice elegance which has a nice analogy to `n_leaf_ellipsoids`?

I take the point about the addition resulting in a different object though I'm not sure people would necessarily confuse it with a well-defined algebra...

pints/_nested/__init__.py
ben18785: Not sure I get ya?

pints/_nested/__init__.py
ben18785: I could have a separate method, that's fine. Just thought since it'd only be used within ellipsoid stuff, I would put it here.

I don't want to have the later `mahalanobis_distance(self, point)` because I want to have the freedom to use the method without actually having to construct ellipsoid objects to avoid the overhead (as is done in some of the functions that use it).

docs/source/nested_samplers/nested_multinest_sampler.rst
```diff
@@ -0,0 +1,7 @@
+*****************
+Multinest sampler
```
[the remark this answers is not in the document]
ben18785: Yep, I'll change this. Good spot.

docs/source/nested_samplers/index.rst
```diff
@@ -7,4 +7,4 @@ Nested samplers
     nested_sampler
     nested_ellipsoid_sampler
     nested_rejection_sampler
-
+    nested_multinest_sampler
```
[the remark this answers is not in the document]
ben18785: MultiNest a type of nested sampler so this was more a case of organising the library opposed to using their real names.

examples/sampling/nested-multinest.ipynb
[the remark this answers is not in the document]
MichaelClerx: Wait, the multinest class doesn't use them?

pints/_nested/__init__.py
MichaelClerx: What about a + b + c though, that'd become either [[a, b], c] or [a, [b, c]] deciding on which order python makes the calls in?

pints/_nested/__init__.py
MichaelClerx: the number of points within a(?) bounding ellipsoid

examples/sampling/nested-multinest.ipynb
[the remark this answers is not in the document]
ben18785: MultiNest uses them but a user wouldn't need ever to call EllipsoidTree or Ellipsoid (or any of their methods) themselves.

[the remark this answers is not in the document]
MichaelClerx: I dunno. If I was using multinest I think Gary would make me visualise the ellipses to check what it was doing :D

pints/_nested/__init__.py
ben18785: So, hilariously, despite much energy and thought on my side, I wasn't actually using this add function. I thought I was but actually I was just adding lists! (So, I've deleted it.)

pints/_nested/__init__.py
ben18785: Yep, done.

pints/_nested/__init__.py
ben18785: I think this is ok, to be honest: it's more a personal style issue. (There are also examples like this abound throughout PINTS.)

examples/sampling/nested-multinest.ipynb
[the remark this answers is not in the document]
ben18785: Hmm, re: location, I'm not sure either. Ellipsoid is required by ellipsoidal nested sampling and multinest; ellipsoidtree is only required by multinest...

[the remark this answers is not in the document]
MichaelClerx: We could change from pints._nested to pints.nested, but only import the samplers?
So then you'd do
```
import pints
pints.MultiNest
```
but
```
import pints.nested
e = pints.nested.Ellipsoid
```
?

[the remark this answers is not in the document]
MichaelClerx: You'd still be able to do ``pints.nested.MultiNest``, so that would be _slightly_ messy

PR #1321, opened 2021-03-16
Subcapture warning pt 2

MichaelClerx:
Hi @martinjrobins !

I did a bit of testing and it looks like these capture methods should be OK when used in multiprocessing: https://github.com/MichaelClerx/myokit/issues/727
I've checked with `spawn` and `fork`, so should handle the windows & linux defaults

So I've downgradedd the warning to "not thread safe" (which I think is the issue you were having)

This is good news w.r.t. myokit as well, as it's much easier to make it thread-safe than it would be to fix similar issues for multiprocessing!

### DIFF
--- a/pints/_evaluation.py
+++ b/pints/_evaluation.py
@@ -452,7 +452,7 @@ class _Worker(multiprocessing.Process):
 
     def run(self):
         # Worker processes should never write to stdout or stderr.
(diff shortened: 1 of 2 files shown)


martinjrobins (approved):
great to hear :) I guess multiprocessing uses separate processes, which maybe all have their own stdout/err....? Glad it works ok though

Issue #1325, opened 2021-03-23
Mistake in acceptance ratio in Emcee hammer


ben18785:
During functional testing, I realised there's a mistake in Emcee hammer that means it doesn't converge to the correct target distribution. It should be $q = z^(N-1) p(Y)/p(X)$ where $N$ is the dimensionality of the space _not_ the number of chains.

PR #1326, opened 2021-03-23
Bug fix in emcee hammer

ben18785:
- Separated sample_z out into its own function as this makes more explicit what's happening
- Cleaned up choosing a member from the complementary set
- Corrected the setter for scale (and corrected its test)

Addresses #1325

### DIFF
--- a/pints/_mcmc/_emcee_hammer.py
+++ b/pints/_mcmc/_emcee_hammer.py
@@ -102,8 +102,7 @@ class EmceeHammerMCMC(pints.MultiChainMCMC):
         Selects an index uniformly at random from all chains excluding the
         index of the current chain.
         """
-        all_chains = list(range(self._n_chains))
-        free_chains = all_chains.copy()
+        free_chains = list(range(self._n_chains))
         free_chains.remove(current_index)
         other_index = np.random.choice(free_chains)
         return other_index
--- a/pints/_mcmc/_emcee_hammer.py
+++ b/pints/_mcmc/_emcee_hammer.py
@@ -25,15 +25,16 @@ class EmceeHammerMCMC(pints.MultiChainMCMC):
 
     - Set ``Y = X_j(t) + z[X_k(t) - X_j(t)]``.
 
-    - Set ``q = z^{N - 1} p(Y) / p(X_k(t))``.
+    - Set ``q = z^{d - 1} p(Y) / p(X_k(t))``.
 
     - Sample ``r ~ U(0, 1)``.
 
     - If ``r <= q``, set ``X_k(t + 1)`` equal to ``Y``, if not use ``X_k(t)``.
 
-    Here, ``N`` is the number of chains (or walkers), and ``g(z)`` is
-    proportional to ``1 / sqrt(z)`` if ``z`` is in  ``[1 / a, a]`` or to 0,
-    otherwise (where ``a`` is a parameter with default value ``2``).
+    Here, ``N`` is the number of chains (or walkers), ``d`` is the
+    dimensionality of the space, and ``g(z)`` is proportional to
+    ``1 / sqrt(z)`` if ``z`` is in  ``[1 / a, a]`` or to 0, otherwise (where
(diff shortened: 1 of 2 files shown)


pints/_mcmc/_emcee_hammer.py
```diff
@@ -82,15 +82,13 @@ def ask(self):
             self._remaining = np.delete(self._remaining, 0)
 
             # Pick j from the complementary ensemble
-            j = np.random.randint(self._n_chains - 1)
-            if j >= self._k:
-                j += 1
+            compl_indices = list(set(range(self._n_chains)) -
+                                 set([self._k]))
+            j = np.random.choice(compl_indices)
```
DavAug, line 87: Sorry I don't know much about this sampler. 
- k is the index of the current runner/chain?
- j is the new runner that you want to pick?

Would it potentially be easier (and at least for me easier to understand) if you did something like:
```
# Create list of runners (Could also be done in the init)
self._runners = list(range(self._n_chains)) 

# Pick free runner
free_runners = self._runners.copy()
free_runners.remove(self._k)
j = np.random.choice(free_runners)
```

pints/_mcmc/_emcee_hammer.py
```diff
@@ -101,6 +99,14 @@ def ask(self):
         # Return proposed points
         return self._proposed
 
+    def _sample_z(self, a):
+        """
+        Samples ``z~g(z)`` where ``g(z)`` is proportional to ``1 / sqrt(z)``
+        if ``z`` is in ``[1 / a, a]``; otherwise 0.
+        """
+        r = np.random.rand()
+        return ((1 + r * (a - 1))**2) / a
```
DavAug, line 108: Could you explain how this formula relates to g(z) propto 1/sqrt(z) and the interval condition on z?

pints/_mcmc/_emcee_hammer.py
```diff
@@ -160,7 +166,7 @@ def tell(self, fx):
         accepted = np.array([False] * self._n_chains)
         r_log = np.log(np.random.rand())
         q = (
-            (self._n_chains - 1) * np.log(self._z)
+            (self._n_parameters - 1) * np.log(self._z)
             + proposed_log_pdf[0] - self._current_log_pdfs[self._k])
```
DavAug, line 170: Could you explain this q to me? Is this the log q of the q mentioned in the docstring
```
q = z^{N - 1} p(Y) / p(X_k(t))?
```
If so, maybe `log_q` would be more clear? What do you think?

In the docstring it also says: ``N`` is the number of chains (or walkers). So do you really want to use the number of parameters, or is the docstring wrong?

DavAug (changes requested):
Thanks @ben18785 !

I don't know much about the sampler, so I apologise for potentially uninformed comments! It looks good though, there are just some bits that are not immediately clear to me.

pints/_mcmc/_emcee_hammer.py
ben18785: Good point: I've created another method that handles this and uses the code you've posted. Re: k and j, yes, that's right: I know j and k aren't informative but they match the notation in the paper, so I'd like to keep them like that.

pints/_mcmc/_emcee_hammer.py
ben18785: I've added a note to indicate that the method uses inverse transform sampling. I'm not sure I want to explain the formula itself though...

pints/_mcmc/_emcee_hammer.py
ben18785: Renamed to `log_q`: good idea! And, yes, good spot in the docstrings: they were wrong and have now been corrected.

ben18785:
Thanks for the review @DavAug -- I think I've changed everything as you wanted. Let me know if you're now happy for this to go in.

DavAug (approved):
Thanks @ben18785 . It looks really good to me!

ben18785:
Thanks @DavAug !

Issue #1328, opened 2021-03-24
Add evaluateS1 method for HighDimensionalGaussianLogPDF


ben18785:
Why? Because it'll be very useful for functionally testing gradient-based methods. I think this can just be copied straight over from that for `GaussianLogPDF`.

Whilst we're at it, it might be good to update the docstrings as they're not very informative.

ben18785:
Also, the `sigma_inv` calculation in the method in `GaussianLogPDF` can be cached, which can be addressed in same PR, I reckon.

ben18785:
Also, add note about result being from chapter 2 of matrix cookbook (plus maybe https://stats.stackexchange.com/questions/27436/how-to-take-derivative-of-multivariate-normal-density).

MichaelClerx:
Fixed in #1329

PR #1329, opened 2021-03-24
Added evaluateS1 for high dim Gaussian

ben18785:
- Made docstrings for high dim Gaussian more informative
- Cached covariance matrix for Gaussian case
- Added reference for derivative of log density for Gaussian

Addresses #1328

### DIFF
--- a/pints/toy/_high_dimensional_gaussian.py
+++ b/pints/toy/_high_dimensional_gaussian.py
@@ -15,8 +15,10 @@ from . import ToyLogPDF
 
 class HighDimensionalGaussianLogPDF(ToyLogPDF):
     """
-    High-dimensional multivariate Gaussian log pdf, with off-diagonal
-    correlations.
+    High-dimensional zero-mean multivariate Gaussian log pdf, with off-diagonal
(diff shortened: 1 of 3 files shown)


pints/toy/_high_dimensional_gaussian.py
```diff
@@ -15,8 +15,10 @@
 
 class HighDimensionalGaussianLogPDF(ToyLogPDF):
     """
-    High-dimensional multivariate Gaussian log pdf, with off-diagonal
-    correlations.
+    High-dimensional zero-mean multivariate Gaussian log pdf, with off-diagonal
+    correlations. Specifically, the covariance matrix Sigma is constructed so
```
rcw5890, line 19: Maybe we should start a new paragraph at "Specifically ... " as it is quite detailed.

rcw5890 (approved):
This change seems to be working well for me.
I made one optional suggestion on the docstring.

pints/toy/_high_dimensional_gaussian.py
ben18785: Good idea: done.

Issue #1330, opened 2021-03-26
Matplotlib error on testing


ben18785:
Matplotlib looks like it has now added error checking for the value of `alpha` (which controls the transparency of plots) as I now get the following types of error when running the unit tests on Github.

```
ERROR: test_series (test_plot.TestPlot)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pints/pints/pints/tests/test_plot.py", line 416, in test_series
    pints.plot.series(few_samples, self.problem, thinning=1)
  File "/home/runner/work/pints/pints/pints/plot/_series.py", line 110, in series
    plt.plot(times, v, color='#1f77b4', alpha=alpha)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/pyplot.py", line 2988, in plot
    return gca().plot(
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/axes/_axes.py", line 1605, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 315, in __call__
    yield from self._plot_args(this, kwargs)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 539, in _plot_args
    return [l[0] for l in result]
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 539, in <listcomp>
    return [l[0] for l in result]
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 532, in <genexpr>
    result = (make_artist(x[:, j % ncx], y[:, j % ncy], kw,
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 354, in _makeline
    seg = mlines.Line2D(x, y, **kw)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/matplotlib/lines.py", line 397, in __init__
    self.update(kwargs)
[...]

PR #1331, opened 2021-03-26
Max alpha value of 1 for plotting

ben18785:
Looks like Matplotlib introduced error checking for alpha to ensure its max value is 1. Before, we were allowing alpha > 1.

Addresses #1330

### DIFF
--- a/pints/plot/_series.py
+++ b/pints/plot/_series.py
@@ -92,7 +92,7 @@ def series(samples, problem, ref_parameters=None, thinning=None):
 
     # Guess appropriate alpha (0.05 worked for 1000 plots)
     alpha = max(0.05 * (1000 / (n_sample / thinning)), 0.5)
-    if alpha > 1:
+    if alpha > 1:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


pints/plot/_series.py
```diff
@@ -92,6 +92,8 @@ def series(samples, problem, ref_parameters=None, thinning=None):
 
     # Guess appropriate alpha (0.05 worked for 1000 plots)
     alpha = max(0.05 * (1000 / (n_sample / thinning)), 0.5)
+    if alpha > 1:  # pragma: no cover
```
MichaelClerx, line 95: You can avoid the pragma by doing e.g. `alpha = min(1, max(0.05...))`

MichaelClerx (approved):
Minor comment. Otherwise fine

pints/plot/_series.py
ben18785: Thanks -- I've addressed this in another PR since I needed it to pass the tests.

Issue #1300, opened 2021-02-21
Compute pointwise log-likelihood for each observation


DavAug:
ArviZ provides a simple API to compute the LOO or WAIC for performance assessment of models, see https://arviz-devs.github.io/arviz/api/generated/arviz.waic.html. 

What this would require however is the pointwise log-likelihood scores of the parameters in a chain for each observation. So for N obervations and M iterations and K chains, we would need to store N*M*K log-pdf values.

The computationally most efficient way to generate the pointwise log-likelihoods would potentially be to store the while running the chain before summing them up across observations. That would require some changes in our pints.LogPDF, pints.LogPosterior and the pints.MCMCSampler / pints.MCMCController though. 

Alternatively, we could consider to implement a routine that takes the LogPDF of a problem, and the chains and then computes the log-pdfs for the observations again. This would still require us to implement an additional method for the LogPDFs which returns the pointwise log-pdfs.

MichaelClerx:
Discussed in meeting today:

- Pointwise log-likelihood = log likelihood of every point in a [ProblemLogLikelihood](https://pints.readthedocs.io/en/stable/log_pdfs.html#pints.ProblemLogLikelihood), before summing
- Logical entry to add this in would be somewhere in ProblemLogLikelihood ?

ben18785:
I've actually realised that Stan doesn't save a point-wise log-likelihood as it runs. Instead, it computes it afterwards using each posterior sample. I think, however, that we should probably try to improve on this since our models are generally more expensive to run.

Rebecca-Rumney:
I've been looking into how to do this and this is my idea: 
It requires changing the `__call__` function of each ProblemLogLikelihood and adding 2 new functions so that:
     
    def __call__(self, x):
        pointwise = self.create_pointwise_loglikelihoods(x)
        self._last_pointwise_loglikelihoods = pointwise
        return np.sum(pointwise)
    
    def create_pointwise_loglikelihoods(self, parameters):
        """
        Returns a matrix of size nt x no containing the log likelihood of each observation and at each time point 
        with the given parameters
        """
    
    def get_last_pointwise_loglikelihoods(self):
        return self._last_pointwise_loglikelihoods

[...]

PR #1334, opened 2021-03-31
I1300 pointwise log likelihoods

Rebecca-Rumney:
Addresses #1300 .
Simplest solution to the problem. Just adds a function evaluate_pointwise_loglikelihoods which can be called in the ask-tell loop or calculated separately from the inference.


ben18785:
@Rebecca-Rumney this looks really good! It looks like it just needs a bit of tweaking with the tests and ready to go to review, right?

Issue #1337, opened 2021-04-20
Bug in DramACMC


ben18785:
Tried running DramACMC on `pints.GaussianLogPDF` and it looks like the distribution converges to a stationary distribution that is substantially wider than the target.

PR #1339, opened 2021-04-20
Dram simplification (and addressing bug)

ben18785:
- Simplifies DRAM so it only takes 2 proposal kernels (I think it was overkill to allow more before)
- In doing so, sorts bug highlighted in #1337
- Notebook updated to reflect new functioning

To do:
- [ ] Run functional tests and illustrate convergence in each

For the review, it'd be good if the persons charged with this task could look at the set_hyperparameters function. At the moment, I'm not sure whether this is optimal since one of the hyperparameters is a list.


pints/_mcmc/_dram_ac.py
```diff
@@ -84,131 +62,91 @@ class DramACMC(pints.AdaptiveCovarianceMC):
...
-        self._upper_scale = 1000
-        self._Y = [None] * self._n_kernels
         self._Y_log_pdf = np.zeros(self._n_kernels)
+        self.set_sigma_scale([1, 0.01])  # scale used in [1]_ for experiments
 
     def _adapt_sigma(self):
         """
-        Updates the covariance matrices of the various kernels being used
-        according to adaptive Metropolis routine.
+        Updates the covariance matrices of the 2 kernels being used according
```
MichaelClerx, line 74:
```suggestion
        Updates the covariance matrices of the two kernels being used according
```

pints/_mcmc/_dram_ac.py
```diff
@@ -69,8 +47,8 @@ class DramACMC(pints.AdaptiveCovarianceMC):
 
     where ``gamma = adaptations^-eta``, ``theta`` is the current state of
     the Markov chain and ``accepted`` is a binary indicator for whether any of
-    the series of proposals were accepted. The kernels for the all proposals
-    are then adapted as ``[scale_1, scale_2, ..., scale_k] * sigma``, where the
+    the series of proposals were accepted. The kernels for the 2 proposals
```
MichaelClerx, line 50:
```suggestion
    the series of proposals were accepted. The kernels for the two proposals
```

pints/_mcmc/_dram_ac.py
```diff
@@ -84,131 +62,91 @@ class DramACMC(pints.AdaptiveCovarianceMC):
...
-    def set_n_kernels(self, n_kernels):
-        """ Sets number of proposal kernels. """
-        if n_kernels < 1:
-            raise ValueError('Number of proposal kernels must be equal to ' +
-                             'or greater than 1.')
-        self._n_kernels = int(n_kernels)
+        self.set_sigma_scale(x[1])
 
-    def set_upper_scale(self, upper_scale):
+    def set_sigma_scale(self, scales):
```
MichaelClerx, line 139: Will users want to put in a list here, or should we just make the arguments be scale1 and scale2?
(Question, not a suggestion! I have no idea)

MichaelClerx (approved):
Looks good! Two minor tweaks and a question, but once those are handled I'm happy for this to go in.

pints/_mcmc/_dram_ac.py
MichaelClerx: Actually, change this to scale1 and scale2 please, so that you resolve the issue with hyperparameters too

MichaelClerx (changes requested):
No reason for scales to be a list now that there's always exactly two.

MichaelClerx (approved):

PR #1343, opened 2021-05-13
Docs typos

### LINKED ISSUE #1342: Fix typos in toy and prior docstrings
While working on the pints paper I found some minor typos in the docstrings

- [x] toy models
- [x] likelihoods
- [x] priors

rcw5890:
Closes #1342


pints/_log_priors.py
```diff
@@ -499,7 +499,7 @@ class HalfCauchyLogPrior(pints.LogPrior):
     r"""
     Defines a 1-d half-Cauchy (log) prior with a given ``location`` and
     ``scale``. This is a Cauchy distribution that has been truncated to lie in
-    between :math:`[0,\infty]`, with pdf
+    between :math:`[0,\infty)`, with pdf
```
MichaelClerx, line 502: Sure about this one? I don't quite see where the truncation happens in the code below?

MichaelClerx (approved):

pints/_log_priors.py
rcw5890: Hmm, this originally looked wrong to me just on the basis of notation, as in my experience, the square brackets at infinity ("[-∞,..." and "...,∞]") are only used for intervals on the extended reals, whereas the support of a cauchy distribution is always written (-∞, ∞).
I'd be happy to hear from @ben18785  who seems to have written the code here

rcw5890: Looking at the code, (0, ∞) is the support of this distribution

Issue #1160, opened 2020-06-02
1-D line search alogrithms


alisterde:
As we discovered when looking at the L-BFGS optimiser (#1083 ) line search algorithms are not that straightforward and are an integral part of Quasi-Newton and conjugate gradient methods. Furthermore, a lot of algorithms for such methods are reliant on legacy code implemented in FORTRAN. It would be extremely beneficial to have a working ask/tell line search in pints as it would pave the way to include numerous Quasi-Newton and conjugate gradient methods, while maintaining pints's lightweight low dependency ethos.

Good initial literature on the area is listed below:
https://www.math.lsu.edu/~hozhang/papers/cg_descent.pdf
http://www.caam.rice.edu/~yzhang/caam554/pdf/cgsurvey.pdf
https://dl.acm.org/doi/pdf/10.1145/1132973.1132979

This shows the 'Hager-Zhang' line search would be of particular interest to implement.

This is part of #684.

alisterde:
A starting points for this could be just getting a working line search algorithm (maybe even start just as a for loop) in a Jupyter notebook for a simply parabola with parameters of the same order of magnitude, increase the complexity of the model, then implement an ask/tell line search in the pints framework.

PR #1344, opened 2021-05-24
Hager_Zhang first commit

alisterde:
@MichaelClerx @ben18785
I've got an initial draft of the Hager-Zhang line search algorithm #1160  it is passing all of the tests I've written and appears to be functional.
However, I want to write more tests to check specific conditions in the logic before I'll be confident in the algorithm, I also think I need to add a few more convergence checks but I'll address that tomorrow.

I would appreciate any initial thoughts you have about it.

### DIFF
--- a/pints/_optimisers/_hager_zhang_line_search.py
+++ b/pints/_optimisers/_hager_zhang_line_search.py
@@ -1,0 +1,899 @@
+#
+# Fixed learning-rate gradient descent.
+#
+# This file is part of PINTS (https://github.com/pints-team/pints/) which is
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+
+import pints
+import numpy as np
+from numpy.linalg import norm
+
+
+class HagerZhang(pints.Optimiser):
+    """
+    Gradient-descent method with a fixed learning rate.
+    """
+
+    def __init__(self, x0, sigma0=None, boundaries=None):
+        super(HagerZhang, self).__init__(x0, sigma0, boundaries)
+
+        # Set optimiser state
+        self._running = False
+        self._ready_for_tell = False
+
+        # Best solution found
+        self._xbest = self._x0
+        self._fbest = float('inf')
+        self._dfdx_best = [float('inf'), float('inf')]
+
+        # Number of iterations run
+        self._iterations = 0
+
+        # Parameters for wolfe conditions on line search
+
+        # As c1 approaches 0 and c2 approaches 1, the line search
+        # terminates more quickly.
+        self._c1 = 1E-4  # Parameter for Armijo condition rule, 0 < c1 < 0.5
+        self._c2 = 0.9  # Parameter for curvature condition rule, c1 < c2 < 1.0
+
+        # boundary values of alpha
+        self._minimum_alpha = 0.0
+        self._maximum_alpha = float("inf")
+        self._proposed_alpha = 0.001  # same default value as used in stan
+
+        # Increase allowed between accepted positions when using approximate
+        # wolfe conditions, this takes into acccount machine error and
+        # insures decreasing.
+        self.epsilon = 1E-6
+
+        # range (0, 1), used in the ``self.__update()`` and
+        # ``self.__initial_bracket()`` when the potential intervals violate
+        # the opposite slope condition (see function definition)
+        self.theta = 0.5
+
+        self.__gamma = 0.66
+
+        # range (0, 1) small factor used in initial guess of step size
+        self.__ps_0 = 0.01
+        # range (0, 1) small factor used in subsequent guesses of step size
+        self.__ps_1 = 0.1
+        # range (1, inf) factor used in subsequent guesses of step size
+        self.__ps_2 = 2.0
+
+        self.__current_alpha = None
+
+        # approximate inverse hessian
+        # initial the identity is used
+        self._B = np.identity(self._n_parameters)
+
+        # newton direction
+        self.__px = None
+
+        # number of accepted steps/ newton direction updates
+        self._k = 0
+
+        # number of steps in the current line search iteration
+        self.__j = 0
+        self.__logic_steps_left = 0
+
+        # maximum number of line search steps before a successful point
+
+        # packed reply of objective function and gradient evaluations
+        # needed by generator implementation
+        # this approach is taken rather than the send() keyword to
+        # back date past Python 2.5
+        self.__reply_f_and_dfdx = None
+
+        # Current point, score, and gradient
+        self._current = self._x0
+        self._current_f = None
+        self._current_dfdx = None
+
+        # Proposed next point (read-only, so can be passed to user)
+        self._proposed = self._x0
+        self._proposed.setflags(write=False)
+
+        self.__convergence = False
+
+        # ask generator
+        self.__ask_generator = self.__ask_generator()
+
+        # need to check wolfe conditions
+        self.wolfe_check = False
+        # continue main body of line search loop
+        self.__line_search_looping = None
+        # wheather the wolfe conditions have been meet
+        self.__wolfe_cond_meet = False
+
+    def name(self):
+        """ See :meth:`Optimiser.name()`. """
+        return 'Hager-Zhang Line Search'
+
+    def fbest(self):
+        """ See :meth:`Optimiser.fbest()`. """
+        return self._fbest
+
+    def wolfe_line_search_parameters(self):
+        """
+        Returns the  wolfe line search parameters this optimiser is using
+        as a vector ``[c1, c2]``.
+        As c1 approaches 0 and c2 approaches 1, the line search terminates
+        more quickly. ``c1`` is the parameter for the Armijo condition
+        rule, ``0 < c1 < 0.5``. ``c2 `` is the parameter for the
+        curvature condition rule, ``c1 < c2 < 1.0``.
+        """
+        return (self._c1, self._c2)
+
+    def needs_sensitivities(self):
+        """ See :meth:`Optimiser.needs_sensitivities()`. """
+        return True
+
+    def n_hyper_parameters(self):
+        """ See :meth:`pints.TunableMethod.n_hyper_parameters()`. """
+        return 2
+
+    def running(self):
+        """ See :meth:`Optimiser.running()`. """
+        return self._running
+
+    def set_hyper_parameters(self, x):
+        """
+        See :meth:`pints.TunableMethod.set_hyper_parameters()`.
+
+        The hyper-parameter vector is ``[c1, c2, m]``.
+        ``c1`` is the parameter for the Armijo condition rule,
+        ``0 < c1 < 0.5``.
+        ``c2`` is the parameter for the curvature condition rule,
+        ``c1 < c2 < 1.0``.
+        ``m`` is the number of maximum number of correction matrices
+        that can be stored for the LM-BFGS update.
+        """
+
+        self.__set_wolfe_line_search_parameters(x[0], x[1])
+
+    def __set_wolfe_line_search_parameters(self, c1, c2):
+        """
+        Sets the parameters for the wolfe conditions.
+
+        Parameters
+        ----------
+        c1: float
+        Parameter for the Armijo condition rule, ``0 < c1 < 0.5``.
+
+        c2: float
+        Parameter for the curvature condition rule, ``c1 < c2 < 1.0``.
+        """
+        if(0 < c1 < 0.5 and c1 < c2 < 1.0):
+            self._c1 = c1
+            self._c2 = c2
+        else:
+            cs = self.wolfe_line_search_parameters()
+            raise ValueError('''Invalid wolfe line search parameters!
+                                Parameters must meet the following conditions:
+                                0 < c1 < 0.5 and c1 < c2 < 1.0
+                                currently c1 = ''', cs[0], 'and  c2 = ', cs[1])
+
+    def set_search_direction(self, px):
+        """
+        Sets the search direction vector px.
+        If this is not a descent direction the line searches will fail
+        i.e return nonsense or actually fail
+
+        Parameters
+        ----------
+        px: vector
+        search direction vector this must be of the same length as x
+        """
+        # TODO: maybe add a check that px contains numbers?
+        if(len(px) == self._n_parameters):
+            self.__px = np.asarray(px)
+        else:
+            raise ValueError('Invalid search direction, px must be of length '
+                             + str(self._n_parameters))
+
+    def xbest(self):
+        """ See :meth:`Optimiser.xbest()`. """
+        return self._xbest
+
+    def stop(self):
+        """ See :meth:`Optimiser.stop()`. """
+
+        if self.__wolfe_cond_meet:
+            # wolfe conditions have been meet
+            # or convergence has occurred
+
+            # resetting so optimiser can be recalled
+            self.__wolfe_cond_meet = False
+            self._running = False
+            return 'Wolfe conditions meet'
+        else:
+
+            return False
+
+    def ask(self):
+        """ See :meth:`Optimiser.ask()`. """
+        if np.any(self.__px) is None:
+            raise ValueError(''''Warning search direction has not been set
+                                please set search direction using:
+                                set_search_direction(px = '')
+                                before optimising.
+                                ''')
+
+        self.wolfe_check, self._proposed = next(self.__ask_generator)
+
+        # Running, and ready for tell now
+        self._ready_for_tell = True
+        self._running = True
+
+        return [self._proposed]
+
+    def __ask_generator(self):
+        """ See :meth:`Optimiser.ask()`. """
+        # check wolfe conditions
+        # yes_wolfe_check = True
+        # don't check wolfe conditions
+        no_wolfe_check = False
+        # if not self._running:
+        self._proposed = np.asarray(self._x0)
+        # need to evaluate initial position
+        yield no_wolfe_check, self._proposed
+
+        # working out an initial stepsize value alpha
+        alpha_0 = self.__initialising(k=self._k,
+                                      alpha_k0=self.__current_alpha)
+
+        # Creating an initial bracketing interval of alpha values
+        # satisfying the opposite slope condition (see function
+        # docstring) beginning with initial guess [0, alpha_initial].
+        initial_bracket_gen = self.__initial_bracket(c=alpha_0)
+        continue_bracketing, postion = next(initial_bracket_gen)
+        while continue_bracketing is True:
+            # print(postion)
+            yield no_wolfe_check, postion
+            continue_bracketing, postion = next(initial_bracket_gen)
+        # saving bracket information
+        bracket = postion
+        self._minimum_alpha = bracket[0]
+        self._maximum_alpha = bracket[1]
+
+        # loop of steps L0-L3 from [1]
+        self.__line_search_looping = True
+        while self.__line_search_looping:
+
+            a = self._minimum_alpha
+            b = self._maximum_alpha
+
+            converged = self.__very_close(a, b)
+            if converged:
+                pass
+            # step L1 of line search from [1]
+            # calling secant2 as a generator
+            secant2_gen = self.__secant2(a, b)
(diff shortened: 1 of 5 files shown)


pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
+#
+# Fixed learning-rate gradient descent.
```
MichaelClerx, line 2:
```suggestion
# Hager-Zhang line search.
```

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+
+import pints
+import numpy as np
+from numpy.linalg import norm
+
+
+class HagerZhang(pints.Optimiser):
```
MichaelClerx, line 16: Best to have this consistent with the file name, so either HagerZhangLineSearch or _hager_zhang.py

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+    """
+    Gradient-descent method with a fixed learning rate.
```
MichaelClerx, line 18: Needs a docstring with references etc.

pints/_optimisers/_hager_zhang_line_search.py
MichaelClerx, line 16: Do we need an interface for line search methods? E.g. `class HagerZhang(pints.LineSearch)` with `class LineSearch(pints.Optimiser)` ?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+    """
+
+    def __init__(self, x0, sigma0=None, boundaries=None):
+        super(HagerZhang, self).__init__(x0, sigma0, boundaries)
+
+        # Set optimiser state
```
MichaelClerx, line 24: I take it this method only works for 1d problems. Should there be a check in the constructor for this? Or in a LineSearch class that this one extends?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+        # boundary values of alpha
+        self._minimum_alpha = 0.0
+        self._maximum_alpha = float("inf")
+        self._proposed_alpha = 0.001  # same default value as used in stan
+
+        # Increase allowed between accepted positions when using approximate
+        # wolfe conditions, this takes into acccount machine error and
+        # insures decreasing.
+        self.epsilon = 1E-6
```
MichaelClerx, line 51: We've shied away from public properties in PINTS so far, so this should probably be _epsilon with methods to get and set it

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+        # range (0, 1), used in the ``self.__update()`` and
+        # ``self.__initial_bracket()`` when the potential intervals violate
+        # the opposite slope condition (see function definition)
+        self.theta = 0.5
```
MichaelClerx, line 56: Same

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+        self.__gamma = 0.66
```
MichaelClerx, line 58: We're not using __ anywhere in PINTS, just _ to mark this as private

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        if self._fbest > proposed_f:
+            self._fbest = proposed_f
+            self._dfdx_best = proposed_dfdx
+            self._xbest = self._current
+
+        # print('finished tell')
+
+    def __initialising(self, k, alpha_k0):
+        '''
+        This function is part of the Hager-Zhang line search method [1].
```
MichaelClerx, line 385: This line's not really needed!

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        # We need the evaluation of the gradients before we can start the BFGS,
+        # the first tell gets these.
+        if self._current_f is None:
+
+            # Move to proposed point
+            self._current = self._proposed
+            self._current_f = np.asarray(proposed_f)
+            self._current_dfdx = np.asarray(proposed_dfdx)
+
+        elif self.wolfe_check is True:
```
MichaelClerx, line 324:
```suggestion
        elif self.wolfe_check:
```

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+import pints
+import numpy as np
+from numpy.linalg import norm
+
+
+class HagerZhang(pints.Optimiser):
+    """
+    Gradient-descent method with a fixed learning rate.
+    """
+
```
MichaelClerx, line 20: I think we should add a big long `# comment` here explaining the structure of this class and how the `yield` stuff is supposed to work

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        self.epsilon = 1E-6
+
+        # range (0, 1), used in the ``self.__update()`` and
+        # ``self.__initial_bracket()`` when the potential intervals violate
+        # the opposite slope condition (see function definition)
+        self.theta = 0.5
+
+        self.__gamma = 0.66
+
+        # range (0, 1) small factor used in initial guess of step size
```
MichaelClerx, line 60: There's quite a lot of properties declared down here. Seems this method has a lot of "state"!

- Are they all strictly necessary?
- And if so, would there be any sense in bundling them per subtask, e.g. some object for wolfe-related properties, something like that?

MichaelClerx (commented):
Thanks @alisterde 

This looks BIG! I've added some initial comments on coding and style, but I'll need some more explanation before I can review what the code actually does

pints/tests/test_opt_hager_zhang.py
```diff
@@ -0,0 +1,476 @@
...
+        b = pints.RectangularBoundaries([-1, -1], [1, 1])
+        px = [0.0001, 0.0001]
+        return r, x, s, b, px
+
+    def test_unbounded(self):
+        # Runs an optimisation without boundaries.
+        r, x, s, b, px = self.problem()
+        opt = pints.OptimisationController(r, x, method=method)
+        m = opt.optimiser()
+        _B = np.identity(m._n_parameters)
```
MichaelClerx, line 53: Why the underscore?

pints/tests/test_opt_hager_zhang.py
```diff
@@ -0,0 +1,476 @@
...
+        reply = r.evaluateS1(proposed)
+        m.tell([reply])
+        # grad = r.evaluateS1(x)[1]
+        # grad =np.matmul(np.transpose(grad), np.asarray([0.0001, 0.0001]))
+        # print('grad ',grad)
+        a = 0.5  # d = 0.75
+        b = 1.0
+
+        def temp(a=a, b=b, m=m):
+            generator = m._HagerZhang__bisect_or_secant(a=a, b=b)
```
MichaelClerx, line 271: Not sure what the best way to test these is. _As a rule of thumb_ you don't test the private API. So we need to figure out if this is an exception or if there's something we're not seeing that would break this down into smaller testable blocks with public methods

MichaelClerx:
Maybe we can link to this somewhere in the file. I can never find it :D
https://docs.python.org/3.8/reference/expressions.html#yield-expressions

MichaelClerx:
And this one, for the discussion https://docs.python.org/3.8/glossary.html#term-asynchronous-generator

MichaelClerx:
https://docs.python.org/3.8/glossary.html#term-asynchronous-generator
https://docs.python.org/3/reference/compound_stmts.html#coroutines

alisterde:
This is the paper referenced:
https://epubs.siam.org/doi/abs/10.1137/030601880?casa_token=GhCspIQb5DUAAAAA:aVcINgBT6bK9q154vVI9UDxbnPuwNEUCc-Ph_tTVKR6hyfEjkWsk6xNZD6KQMG0ZGH7Xa8XnVCplew

pints/_optimisers/_hager_zhang_line_search.py
ben18785: Pseudocode too would I think be really useful. For pseudocode in the other methods, we tend not to incorporate ask/tell but rather lay out the algorithm as given in the paper (but made tidy and understandable for us).

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        self._dfdx_best = [float('inf'), float('inf')]
+
+        # Number of iterations run
+        self._iterations = 0
+
+        # Parameters for wolfe conditions on line search
+
+        # As c1 approaches 0 and c2 approaches 1, the line search
+        # terminates more quickly.
+        self._c1 = 1E-4  # Parameter for Armijo condition rule, 0 < c1 < 0.5
```
ben18785, line 40: I couldn't find the Armijo rule in the paper? Could you point to where this is from? I'm guessing this is "delta" on page 184 of Hager and Zhang?

ben18785: Also, where the default value was taken from?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        self._c2 = 0.9  # Parameter for curvature condition rule, c1 < c2 < 1.0
```
ben18785, line 41: I'm guessing this is sigma on page 184 of the paper? Of course, we don't need to use the same parameter names as them but it might help in places to cross-check.

ben18785: Similarly as above for the default value.

pints/_optimisers/_hager_zhang_line_search.py
ben18785: Again, would be good to know where the default value comes from here.

pints/_optimisers/_hager_zhang_line_search.py
ben18785: I agree re: bundling. Could we perhaps bundle according to first and second Wolfe and approximate? Or would that not work?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        self.__convergence = False
+
+        # ask generator
+        self.__ask_generator = self.__ask_generator()
+
+        # need to check wolfe conditions
+        self.wolfe_check = False
+        # continue main body of line search loop
+        self.__line_search_looping = None
+        # wheather the wolfe conditions have been meet
```
ben18785, line 109: whether*

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+    def name(self):
+        """ See :meth:`Optimiser.name()`. """
+        return 'Hager-Zhang Line Search'
+
+    def fbest(self):
+        """ See :meth:`Optimiser.fbest()`. """
+        return self._fbest
+
+    def wolfe_line_search_parameters(self):
```
ben18785, line 120: Ah ha -- you've already bundled them here. How about calling them something like: `wolfe_line_param_1` and `wolfe_line_param_2`?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        The hyper-parameter vector is ``[c1, c2, m]``.
+        ``c1`` is the parameter for the Armijo condition rule,
+        ``0 < c1 < 0.5``.
+        ``c2`` is the parameter for the curvature condition rule,
+        ``c1 < c2 < 1.0``.
+        ``m`` is the number of maximum number of correction matrices
+        that can be stored for the LM-BFGS update.
+        """
+
+        self.__set_wolfe_line_search_parameters(x[0], x[1])
```
ben18785, line 156: There seem to be three parameters mentioned in the docstrings but only two set here? Or is one of them a vector?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        """ See :meth:`Optimiser.stop()`. """
+
+        if self.__wolfe_cond_meet:
+            # wolfe conditions have been meet
+            # or convergence has occurred
+
+            # resetting so optimiser can be recalled
+            self.__wolfe_cond_meet = False
+            self._running = False
+            return 'Wolfe conditions meet'
```
ben18785, line 212: Seems a bit strange to return a string if true and a false if not?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+        # Update xbest and fbest
+        if self._fbest > proposed_f:
+            self._fbest = proposed_f
+            self._dfdx_best = proposed_dfdx
+            self._xbest = self._current
+
+        # print('finished tell')
+
+    def __initialising(self, k, alpha_k0):
```
ben18785, line 383: Could this method have a more informative name like `initialise_alpha`?

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+            self._running = False
+            return 'Wolfe conditions meet'
+        else:
+
+            return False
+
+    def ask(self):
+        """ See :meth:`Optimiser.ask()`. """
+        if np.any(self.__px) is None:
+            raise ValueError(''''Warning search direction has not been set
```
ben18785, line 220: I would remove "warning" since it makes it appear as if it's a warning not an error...

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+            # Checking this is unnecessary as it is the opposite to the above
+            # conditions. We are therefore in the same situation as when we
+            # started the loop so we update the upper bracket and continue.
+            # (steps U3c from [1])
+            else:
+                b = d
+
+    def __initial_bracket(self, c, rho=5):
+        '''
+        This function is part of the Hager-Zhang line search method [1].
```
ben18785, line 571: This line isn't needed.

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+        This function is used to generate an initial interval [a, b] for alpha
+        satisfying the opposite slope conditions and therefore bracketing
+        the minimum, beginning with the initial guess [0, c].
+        The opposite slope conditions:
+                                φ(a) ≤ φ(0) + epsilon,
+                                φ'(a) < 0 (w.r.t every parameter),
```
ben18785, line 578: Isn't this a 1d search? (As in, why the "every parameter"?)

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+                                φ'(b) ≥ 0 (w.r.t every parameter).
+        where φ(α) = f(x+α*px), f() is the function binging minimised, x is
```
ben18785, line 580: being*

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+        satisfy condition because by assumption, we are in a direction where
+        the function value is decreasing. The second and third conditions
+        together demand that there is at least one zero of the derivative in
+        between a and b. In addition to the interval, the update algorithm
+        requires a third point to be supplied. Usually, this point would lie
+        within the interval [a, b]. If the point is outside this interval,
+        the current interval is returned. If the point lies within the
+        interval, the behaviour of the function and derivative value at this
+        point is used to squeeze the original interval in a manner that
+        preserves the opposite slope conditions.
```
ben18785, line 707: Not a suggested change but I really like these explanation paragraphs! Helped to to understand it.

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+                        bisect_or_secant = False
+                        # returning new brackets
+                        # and ending __initial_bracket()
+                        yield False, output
+
+    def __secant_for_alpha(self, a, b, dfda, dfdb):
+        '''
+        This function is part of the Hager-Zhang line search method [1].
+
+        Preforms a secant step to propose a value of alpha. This is the same as
```
ben18785, line 768: Performs*

ben18785: Or is it preform? (You use the same below?).

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+            proposed_grad = np.matmul(np.transpose(proposed_dfdx), self.__px)
+
+            wolfe_curvature = (proposed_grad >= self._c2 *
+                               np.matmul(np.transpose(self._current_dfdx),
+                                         self.__px))
+
+            exact_wolfe_suff_dec = (self._c1 *
+                                    np.matmul(np.transpose(self._current_dfdx),
+                                              self.__px)
+                                    >= proposed_f - self._current_f)
```
ben18785, line 335: `np.matmul(np.transpose(self._current_dfdx), self.__px)` seems to be being repeated three times here? I'd precalculate it and then reuse.

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+            exact_wolfe = (exact_wolfe_suff_dec and wolfe_curvature)
+
+            # Checking if approximate wolfe conditions are meet.
+            apprx_wolfe_suff_dec = ((2.0 * self._c1 - 1.0) *
+                                    np.matmul(np.transpose(self._current_dfdx),
```
ben18785, line 341: Another time.

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+                                              self.__px) >= proposed_grad)
+
+            apprx_wolfe_applies = proposed_f <= self._current_f + self.epsilon
+
+            approximate_wolfe = (apprx_wolfe_suff_dec and wolfe_curvature
+                                 and apprx_wolfe_applies)
+
+            # If wolfe conditions are meet the line search is stopped.
```
ben18785, line 349: met

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+            pass
+
+        # For the subsequent line search do the following
+        # (step I2 in [1])
+        else:
+            # Increases the step size of the previous accepted step as it
+            # is only decreased in subsequent boundary manipulation.
+            return ps_2 * alpha_k0
+
+    def __very_close(self, x, y):
```
ben18785, line 471: `is_very_close`? That makes it clear that it returns a Boolean.

pints/_optimisers/_hager_zhang_line_search.py
```diff
@@ -0,0 +1,899 @@
...
+
+        This is the same as the bracket routine described in [1] as steps B0-3
+
+        :param c: initial guess for maximum value of alpha
+        :param row: range (1, ∞), expansion factor used in the bracket rule to
+                    increase the upper limit c (c_j+1 = row*c_j) until a
+                    suitable interval is found that contains the minimum.
+        '''
+
+        # (steps B0 from [1])
```
ben18785, line 592: Perhaps I'm missing it but I can't see "B0" in the paper?

ben18785 (changes requested):
Looks really good @alisterde -- it's a beast of a method! My points are mainly surface level things as I'm finding following the logic tricky. As Michael says, having some description in the docstrings that describes the ask/telling would, I think, really help.

Issue #1267, opened 2021-01-27
Eight schools notebook failing in daily test


MichaelClerx:
```
Running examples/toy/distribution-eight-schools.ipynb ... ERROR

{'J': 8, 'y': [28, 8, -3, 7, -1, 1, 18, 12], 'sigma': [15, 10, 16, 11, 9, 11, 10, 18]}
Running...
Using Hamiltonian Monte Carlo
Generating 4 chains.
Running in sequential mode.
Iter. Eval. Accept.   Accept.   Accept.   Accept.   Time m:s
0     4      0         0         0         0          0:00.0
1     84     0.333     0.333     0.333     0.333      0:00.0
2     164    0.5       0.5       0.5       0.5        0:00.0
3     244    0.6       0.6       0.6       0.6        0:00.0
100   8004   0.98      0.941     0.98      0.98       0:01.4
200   16004  0.990099  0.926     0.990099  0.990099   0:02.7

Traceback (most recent call last):
  File "<string>", line 112, in <module>
  File "/home/runner/work/pints/pints/pints/_mcmc/__init__.py", line 668, in run
    fxs = evaluator.evaluate(xs)
  File "/home/runner/work/pints/pints/pints/_evaluation.py", line 106, in evaluate
    return self._evaluate(positions)
  File "/home/runner/work/pints/pints/pints/_evaluation.py", line 398, in _evaluate
    scores[k] = self._function(x, *self._args)
  File "/home/runner/work/pints/pints/pints/toy/_eight_schools.py", line 121, in evaluateS1
    log_prior = pints.GaussianLogPrior(mu, tau)
  File "/home/runner/work/pints/pints/pints/_log_priors.py", line 462, in __init__
    raise ValueError(\'sd parameter must be positive\')
ValueError: sd parameter must be positive
```

ben18785:
Ha, I'm glad we recently changed pints.GaussianLogPrior(mu, tau) in #1256 -- before this was failing silently! I will take a look. Thanks.

ben18785:
This is actually an interesting one: more interesting than it might first appear! It also impinges on how we'd want hierarchical models to work.

In a non-hierarchical model, we'd probably want to raise an error if someone tried to create a Gaussian prior with a negative SD. I.e.:

`pints.GaussianLogPrior(0, -1)` 

should raise an error (as it now does -- it didn't before update #1256).

[...]

PR #1347, opened 2021-06-08
Eight schools negative standard deviation fix

ben18785:
- Returns negative infinity if sd proposed is negative
- Avoids us having to change the way that all log priors work

addresses #1267


MichaelClerx (approved):
Looks sensible to me!

PR #1349, opened 2021-06-24
Updated sphinx conf.py

MichaelClerx:
- Removed use of [`autodoc_default_flags`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_flags), because:
  - It was deprecated in 1.8, sphinx is now onto 4.x
  - The check to see which version to use assumed we were still on 1.x
- Changed copyright year from 2019 to 2021

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -40,17 +40,10 @@ extensions = [
 ]
 
 # Autodoc defaults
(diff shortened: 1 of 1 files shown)


MichaelClerx:
@rcw5890 quick look?

rcw5890 (approved):
This looks great to me

PR #1353, opened 2021-07-09
Fix alpha equation for MALAMCMC

steven-lang:
The equation for `alpha` seems to have a typo in which `\theta^*` was present in the nominator and denominator. The correct version should be with `\theta_t` in the denominator, see https://en.wikipedia.org/wiki/Metropolis-adjusted_Langevin_algorithm.


MichaelClerx (approved):
Thanks!

Issue #1356, opened 2021-07-14
Stop numpy (and others) from using threads inside multiprocessing worker processes


MichaelClerx:
Ran into a weird issue yesterday, and took me absolutely forever too find it, but here it is:

Numpy sometimes uses multi-threading (in a C library, so not limited by the GIL), causing it to address several CPU cores at once. Usually we'd say this is a good (great?) thing, but if you're already trying to implement multiprocessing it can cause a serious slowdown.

The code below demonstrates the issue:
```
#!/usr/bin/env python3
# Numpy parallelisation
np_par = False
# Pints parallelisation
mc_par = False

if not np_par:
    # Disable numpy parallelisation (on by default) by setting some environment variables within this process
    import os
    os.environ["OMP_NUM_THREADS"] = '1'
    os.environ["OPENBLAS_NUM_THREADS"] = '1'
    os.environ["MKL_NUM_THREADS"] = '1'
    os.environ["VECLIB_MAXIMUM_THREADS"] = '1'
    os.environ["NUMEXPR_NUM_THREADS"] = '1'

import numpy as np
import pints

def essp(x):
    x = (x - np.mean(x)) / (np.std(x) * np.sqrt(len(x)))
    rho = np.correlate(x, x, mode='full')
    rho = rho[int(rho.size / 2):]
    return len(x) / (1 + 2 * np.sum(rho[0]))

def test(n_iterations=None):
    chains = np.random.random((1, 60000, 2))
    n_chains, n_samples, n_parameters = chains.shape
    ess = np.empty(shape=(n_chains, n_parameters))
    for chain_id, chain in enumerate(chains):
        ess[chain_id] = [essp(chain[:, i]) for i in range(0, n_parameters)]

def run():
    iterations = [30000]*10
    t = pints.Timer()
    pints.evaluate(test, iterations, mc_par)
    print(f'Completed in {t.format()}')

if __name__ == '__main__':
    run()
```

On my system, running this code fully sequentially takes 17.8s

Enabling only numpy parallelisation, this drops down to 8.5s (a speed-up of 2.09x)

Enabling only pints parallelisation, this drops down to 4.2s (a speed-up of 4.24x)

Enabling both types of parallelisation, the fans switch on, the lights go down, and after five minutes I get bored and hit Ctrl-C. This means it's _at least_ 17x slower than fully sequential, and possibly even deadlocked.

MichaelClerx:
[...]

PR #1357, opened 2021-07-14
Restrict numpy multithreading when multiprocessing

MichaelClerx:
See #1356 

Current solution allows customisation of `ParallelEvaluator` class, but doesn't provide options for the `pints.evaluate` method or for controllers (which just use the new default)


MichaelClerx:
Seems `threadpoolctl` has no 2.7 support.

Also now noticing that numpy seeds are inherited by processes. This has never been a problem before as each worker gets a different task, but giving them all the same task now in functional testing, they are all returning the same result :D

MichaelClerx:
Unit tests for this block indefinitely on ubuntu 16 (which is EOL since april, and will be removed from github actions in sept). Not sure why the tests block now...

MichaelClerx:
Updated for Python 3 only now, and ready for review!

MichaelClerx:
@tlestang I hear you had the same problem? Did you also solve it with threadpoolctl ?

DavAug (approved):
Thanks @MichaelClerx ! Looks convincing to me, but I am also not super familiar with the possible pitfalls of threading.

MichaelClerx:
Thanks @DavAug !

Issue #1358, opened 2021-07-15
Multiprocessing: Seed worker threads


MichaelClerx:
Few points about multiprocessing:

- Currently, each subprocess is created by copying a lot of info from the main process, including the "random state". This means that each subprocess will generate the same sequence of random numbers
- For our current use, this doesn't matter too much: we evaluate a pdf or error in the subprocesses, and we don't cater to stochastic ones particularly
- For other uses, e.g. running repeated experiments ("replicates"), we _do_ want each process to have a different seed
- Ideally, we'd have some mechanism to ensure reproducibility when using a fixed seed in the main process. However, the allocation of tasks to workers is not deterministic, ~so this can't be done!~

~Proposal:~
- ~Call `np.random.seed()` at the start of each subprocess~
- ~Add a note to the "parallel" switch of `pints.evaluate` and to `pints.ParallelEvaluator`, saying that (1) each process will be seeded, and (2) because of that + random allocation of tasks, the results from using np.random inside "tasks" will not be reproducible~

See below!

MichaelClerx:
Just fyi, a worker process looks roughly like this:

```
def start():
    # do seeding here!
    for i in range(max_tasks):
        get_task_from_queue
        perform_task
        set_result
```

so letting the user do the seeding would mean 1 seeding per task, rather than 1 per worker

MichaelClerx:
Come to think of it, we could generate `n_tasks` random seeds at the start of the evaluation and pass them in to the workers along with the tasks. That would make the results deterministic (regardless of the order of execution/allocation)

MichaelClerx:
I wouldn't mind making this seeding-per-task-for-reproducibility the default behaviour then, doesn't add much overhead:

```
#!/usr/bin/env python3
import pints
import numpy as np
import sys

n = 100000
timer = pints.Timer()
# Max seed is 2**32, but randint uses a half-open interval (upper not included)
seeds = np.random.randint(0, 2**32, n)
for seed in seeds:
    np.random.seed(seed)
t = timer.time()
print(f'Total time: {timer.format(t)}')
print(f'Per task: {timer.format(t / n)}')
```

Total time: 0.2 seconds
Per task: 1.9918578599936155e-06 seconds

PR #1360, opened 2021-07-15
Seed worker threads for different, but reproducible, results

MichaelClerx:
See #1358

Tasks are evaluated by "workers". Each worker starts with a copy of the main process's random generator. So unless we do some seeding they will all use the same random numbers!

Works like this:

```
main process wants to evaluate f(x) on 100 points x
generates a list of 100 tuples (index, seed, x)

each worker waits for a task, and when it gets one it:
- seeds its private random generator using "seed"
- evaluates f(x)
- passes back a tuple (index, f(x))

main process receives all the (index, f(x)) pairs, 
stores them at results[index] = f(x)
```

By seeding in this way it doesn't matter which worker evaluates f(x), or in which order the tasks are evaluated - the results are always reproducible

(And of course if we don't seed the main process, then they are reproducible but pseudo-random)


chonlei (approved):
Thanks Michael! I think it looks great! My only comment would be do we want to tell the users (in the docs or something) that running it with or without parallel would get different results even when it's seeded?

MichaelClerx:
Thanks @chonlei ! I've added a note to the evaluate() method and the Evaluator interface

Issue #1362, opened 2021-07-15
Drop Python 2.7 support


MichaelClerx:
Provided we all agree with it in the next PINTS meeting, we can drop Python 2.7 support by:

- [x] Updating the github workflows and removing 2.7
- [x] Removing the branch checks for 2.7 on github
- [x] Updating setup.py to require 3.5 or higher
- [x] Removing `assertRaisesRegexp` in tests
- [x] Removing references to `basestring`
- [x]  Removing references to `cStringIO`
- [x] Removing the Python3 pragmas from coverage, and removing code that is now no longer covered.
- [x] Updating the CHANGELOG
- [x] Removing all `from __future__ imports`
- [x] Doing some searching for "2.7" and "Python 3" and related terms, removing any remaining obsolete code
- [x] Grep for `ImportError`

PR #1365, opened 2021-07-20
Restrict PINTS to Python 3.5 and higher.

MichaelClerx:
See #1362

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,17 +154,6 @@ Guidelines for writing unit tests:
 3. There are hundreds of unit tests, and good developers run all of them several times a day. Therefore, unit tests should be _fast_.
 4. If you're testing something stochastic, seed the number generator as part of the test, i.e. with `np.random.seed(1)`.
 
-### Running more tests
-
-If you want to check your tests on Python 2 and 3, use
-
-```
-$ python2 run-tests.py --unit
-$ python3 run-tests.py --unit
-```
-
-When you commit anything to PINTS, these checks will also be run automatically (see [infrastructure](#infrastructure)).
-
 ### Testing notebooks
 
 To test all example notebooks, use
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,21 +90,9 @@ Class names are CamelCase, and start with an upper case letter, for example `Sup
 
 To be consistent with the work so far, all PINTS material in the repository (code, comments, docstrings, documentation, etc.) should be written in UK english, the only exception being when quoting other sources, e.g. titles of scientific articles in references.
 
-### Python 2 and 3
+### Supported Python versions.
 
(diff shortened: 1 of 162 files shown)


CONTRIBUTING.md
```diff
@@ -281,7 +258,7 @@ Using [Sphinx](http://www.sphinx-doc.org/en/stable/) the documentation in `docs`
             # Differrent weights:
             weights = [
                 1.0,
-                2.7,
+                2.6,
```
fcooper8472, line 261: I don't understand this change

MichaelClerx: That was just because I was grepping "2.7" :D
Happy to undo it although it won't affect anything

pints/__init__.py
```diff
@@ -227,11 +227,12 @@ def version(formatted=False):
...
-    from ._mcmc._nuts import NoUTurnMCMC
-else:   # pragma: no python 3 cover
-    import warnings
-    warnings.warn('No-U-Turn sampler unsupported for Python version < 3.3')
+# We may want to move to the newer asyncio and exclude older Pythons for NUTS?
+#if sys.hexversion >= 0x03030000:
+#    from ._mcmc._nuts import NoUTurnMCMC
+#else:   # pragma: no cover
+#    import warnings
+#    warnings.warn('No-U-Turn sampler unsupported for Python version < 3.3')
```
fcooper8472, line 235: Can these lines be rationalised?

MichaelClerx: They're no longer needed, but we will probably need to update this soon to use the newer syntax for asyncio and coroutines etc., so then we'll need to stick something like this back in again, I think.

Happy to remove

pints/tests/test_mcmc_nuts.py
```diff
@@ -6,15 +6,14 @@
...
-import sys
 
 import pints
 import pints.toy
 
 from shared import StreamCapture
 
 
-@unittest.skipIf(sys.hexversion < 0x03030000, 'No NUTS on Python < 3.3')
+#@unittest.skipIf(sys.hexversion < 0x03030000, 'No NUTS on Python < 3.3')
```
fcooper8472, line 16: Ditto: remove comment entirely?

fcooper8472 (commented):
Looks great! See the few v minor comments I added.

fcooper8472:
Also, out of curiosity, if someone does `pip install pints` on Python 2.7 after this change, will they still get the previous version of PINTS that DID support Python 2.7?

CONTRIBUTING.md
fcooper8472: Ha, I wondered if it was this ;)

Might be useful to have an unofficial rule about using constants 3.x to make it easier when we get rid of old versions.

fcooper8472 (approved):

MichaelClerx:
> Also, out of curiosity, if someone does pip install pints on Python 2.7 after this change, will they still get the previous version of PINTS that DID support Python 2.7?

Good question! If they're really smart it might pick that up from the `setup.py` info? I have no idea 😅

PR #1375, opened 2021-08-05
i1373-transformATION

### LINKED ISSUE #1373: Transform/Transformation
Before we release a new version of PINTS, we need to sort out the naming of transformations, which are sometimes called "transforms".

E.g. MCMCController and OptimisationController have an argument `transform`. This is confusing as the object you pass in is a `Transformation`.

The method `optimise()` similarly takes a `transform` where it should be a `transformation`.

The method `mcmc_sample` doesn't have a transformation argument at all

The nested sampler controller doesn't take a transformation argument

chonlei:
Change all `transform` to `transformation`. Closes #1373.


MichaelClerx (approved):
Thanks @chonlei !

Issue #1376, opened 2021-08-05
Add more useful log likelihoods


ben18785:
There are a number of log-likelihoods which would be useful (this came up in discussion for the [PKPD app](https://github.com/pkpdapp-team/pkpdapp).):

- log-normal: used a lot in PKPD modelling and elsewhere for handling multiplicative noise
- a version of the current `MultiplicativeGaussianLogLikelihood` where $\eta=1$: it seems like this 
- Poisson: useful for handling counts (useful in SIR-type modelling)

DavAug:
Yes, I have thought about noise a little that we'd like to have a constant standard deviation on the log-scale. Such a noise is the log-normal noise model, which in most software packages is stated as this
```
log X = log y + sigma * eta,
```
where `X` is measureable, `y` is the ODE output and `eta` is a standard normal random variable. Then `log X` has the expectation value `log y` and standard deviation `sigma`. Note, however, that the expectation of `X` is *not* `y`, which is usually the standard assumption for error models around ODE solutions. In other words, `y` is not the underlying "truth" when we assuming measurement noise is corrupting the measurements without bias. This has consequences for the inference problem and will change the inferred parameters for the model `y`.

Looking a little bit more into this you can find that `y` is the median of `X` and that the expectation of `X` is given by 
```
E[X] = y * exp(sigma**2 / 2).
```
So, in order to impose the common assumption that `E[X] = y`, we need to shift the mean of the Gaussian:
```
log X = log y + mu + sigma * eta,
```
with `mu = - sigma**2 / 2`. This will give us a distribution that is has mean `E[X] = y` and a constant standard deviation `sigma` on the log scale.

Note, the current `MulitplicativeGaussianLogLikelihood` is an approximation of the log-normal distribution for `sigma <<1`. Here, the standard deviation of the on the log scale is in general not constant for varying `y`.

ben18785:
Interesting. I suppose we could include the shift of the mean as an optional argument to the log-likelihood then? I would probably still have it as a default so that the expectation is the median since this is standard (and I can see arguments for modelling the median opposed to the mean).

Re: the multiplicative model, yep, I see the approximation. I don't think it matters too much if the standard deviation is non-constant on the log scale: or rather, it may do, if that isn't an appropriate assumption to make. That's why it'd be interesting to write a short paper on this.

DavAug:
Sounds good! Do you have an example for when the median might be more appropriate to model?

ben18785:
[...]

PR #1378, opened 2021-08-06
Log-normal log-likelihood

ben18785:
Adds a log-normal distribution both without (default) and with mean-adjustment as described in #1376


pints/_log_likelihoods.py
```diff
@@ -777,6 +777,122 @@ def __init__(self, problem, sigma):
...
+
+        self._mean_adjust = mean_adjust
+
+    def __call__(self, x):
+        sigma = np.asarray(x[-self._no:])
+        if any(sigma < 0):
+            return -np.inf
+
+        soln = self._problem.evaluate(x[:-self._no])
+        if np.any(soln < 0):
```
DavAug, line 849: The pythonic way to do this and a to have a try except statement around the error computation below. Because soln <= 0 will lead to an exception in np.log(soln), you can catch it there. 
This also saves a little bit of unnecessary masking.

pints/_log_likelihoods.py
```diff
@@ -777,6 +777,122 @@ def __init__(self, problem, sigma):
...
+
+        sigma = np.asarray(x[-self._no:])
+
+        # Evaluate, and get residuals
+        y, dy = self._problem.evaluateS1(x[:-self._no])
+
+        # Reshape dy, in case we're working with a single-output problem
+        dy = dy.reshape(self._nt, self._no, self._n_parameters - self._no)
+
+        # Note: Must be (np.log(data) - simulation), sign now matters!
```
DavAug, line 875: I saw this comment also in other log-likelihoods and I got confused because the sign only matters be because we assume this sign for the partial derivative below 😂 
In general the sign doesn't matter. Could you specify that?

pints/tests/test_log_likelihoods.py
```diff
@@ -1405,6 +1405,150 @@ def simulate(self, x, times):
...
+                          problem)
+
+    def test_call(self):
+        # test calls of log-likelihood
+
+        # single output problem
+        sigma = 1
+        mu = 3.7
+        log_like = self.log_likelihood([mu, sigma])
+        self.assertAlmostEqual(log_like, -10.164703123713256)
```
DavAug, line 1448: I personally don't think that those tests are particularly meaningful, because those are just numbers I can't really interpret. Does it potentially make more sense to test known edge cases to make sure the likelihood's and sensitivities are correct? 
For example the expressions reduce when soln=1, or sigma=1. Then you could compare manually computer values from the parameters, which are easier to understand, and compare them to the methods.

DavAug (changes requested):
Looks really good @ben18785 ! I like the Boolean for switching between the means. I just have a couple minor comments and a general comment on testing.

DavAug (approved):

MichaelClerx:
@ben18785 @DavAug this is approved. Want to merge it in now?

pints/_log_likelihoods.py
```diff
@@ -777,6 +777,123 @@ def __init__(self, problem, sigma):
...
+    .. math::
+        \log y(t) \sim \text{Normal}(\log f(\theta, t), \sigma)
+
+    Important to note is that:
+
+    .. math::
+        \mathbb{E}(y(t)) = f(\theta, t) \exp(\sigma^2 / 2)
+
+    As such, the optional parameter `mean_adjust` (if true) adjusts the
+    mean of the distribution so that its expectation is ``f(\theta, t)``. This
```
MichaelClerx, line 802:
```suggestion
    mean of the distribution so that its expectation is :math:`f(\theta, t)`. This
```

Issue #1377, opened 2021-08-06
Return negative infinity for negative standard deviation for various likelihoods


ben18785:
At the moment,

```
log_likelihood = pints.GaussianLogLikelihood(problem)
sigma = -2
log_likelihood([other_params, sigma])
```

returns `nan`. It should return -infinity. This needs to be applied to:

- `AR1LogLikelihood`
- `ARMA11LogLikelihood`
- `CauchyLogLikelihood`
- `ConstantAndMultiplicativeGaussianLogLikelihood`
- `GaussianLogLikelihood`
- `MultiplicativeGaussianLogLikelihood`
- `StudentTLogLikelihood`: for both `nu` and `sigma`

Also, I think that when log-prob is -inf, then the gradients returned _should_ be nans.

ben18785:
Closed by #1379

PR #1379, opened 2021-08-06
Added -inf returns for non-positive scale parameters for log-likelihoods

ben18785:
Added:

- `-np.inf` returns for `pints.LogLikelihood`s if any scale parameters are non-positive
- in the above event, if an `evaluateS1` method exists, the gradient returned is an np array of `np.nan`s

This addresses #1377


pints/_log_likelihoods.py
```diff
@@ -741,6 +752,11 @@ def evaluateS1(self, x):
         """ See :meth:`LogPDF.evaluateS1()`. """
         sigma = np.asarray(x[-self._no:])
 
+        # Calculate log-likelihood
+        L = self.__call__(x)
+        if L == -np.inf:
```
MichaelClerx, line 757: I'm not sure what you want here, but there is a method np.isfinite(L) that is True on a slightly wider variety of conditions (i.e. -inf, +inf, NaN)

ben18785: Good shout: I've changed it to `np.isneginf`

MichaelClerx (approved):

PR #1384, opened 2021-08-11
Added method to 'untwist' samples from the twisted gaussian distribution.

MichaelClerx:
This was already used to calculate a KLD, but is now available to users as well.

Useful e.g. to visualise performance, or check normality of obtained samples


MichaelClerx:
It'd be really great if someone could check the maths here, e.g. by writing a test independently of the code I wrote :D

pints/toy/_twisted_gaussian_banana.py
```diff
@@ -68,6 +68,16 @@ def __call__(self, x):
         y[1] += self._b * ((x[0] ** 2) - self._V)
         return self._phi.logpdf(y)
 
+    def detransform(self, samples):
```
ben18785, line 71: Why not call it "untwist" since we don't mention transforming anywhere (and untwisting seems more evocative) :).

ben18785 (approved):
Looks great. I checked that the samples when untwisting things were normal with the correct mean and std dev.

I've put a comment about naming the function: up to you whether you take it or leave it!

PR #1387, opened 2021-08-11
Added residuals_diagnostics to doc coverage tests

MichaelClerx:
- Added `residuals_diagnostics` to doc coverage tests
- This showed that 3 external modules were exposed as part of `pints.residuals_diagnostics`, so changed the place where they were imported.


rcw5890 (approved):
Thanks for catching this, it looks good to me

MichaelClerx:
Thanks

Issue #1389, opened 2021-08-14
PINTS histogram plot (from trace) looks strange due to various bin sizes


ben18785:
I quite often get histograms of samples that look odd. An example is this one (particularly the middle histogram plot):

![image](https://github.com/pints-team/comp-bio-paper/blob/master/figures/logistic-sampling.pdf)

which is obtained from running the logistic model (as shown [here](https://github.com/pints-team/comp-bio-paper/blob/master/src/logistic-fitting.ipynb)).

These plots make it look like the various chains have sampled way more differently than they actually have. I think this is because the histograms for each chain have different bins. I propose to generate decent bins using information from all the chains and use these for the individual histograms.

To find decent bins, I'd just find the min and max values of each dimension (or percentiles as is also allowed in the [plotting function](https://github.com/pints-team/pints/blob/a308c8ea810c31b97757b9cfdb8b94b57b5c84dd/pints/plot/_trace.py#L11-L128)), then create a linspace with these.

ben18785:
I've labelled this as priority since we need it for the paper.

ben18785:
Closed by #1390

PR #1390, opened 2021-08-16
Standardise histogram bins across chains

ben18785:
When chains run across different areas, since they use separate bin sizes across the chains, the resultant histograms can be incomparable. This PR picks a consistent bin size across them.

addresses #1389

### DIFF
--- a/pints/plot/_histogram.py
+++ b/pints/plot/_histogram.py
@@ -88,33 +88,38 @@ def histogram(
         squeeze=False,    # Tell matlab to always return a 2d axes object
     )
 
+    # Find ranges across all samples
+    stacked_chains = np.vstack(samples)
+    if n_percentiles is None:
+        xmin = np.min(stacked_chains, axis=0)
+        xmax = np.max(stacked_chains, axis=0)
+    else:
+        xmin = np.percentile(stacked_chains,
+                             50 - n_percentiles / 2.,
+                             axis=0)
+        xmax = np.percentile(stacked_chains,
+                             50 + n_percentiles / 2.,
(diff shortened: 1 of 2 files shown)


pints/plot/_histogram.py
```diff
@@ -88,33 +88,38 @@ def histogram(
         squeeze=False,    # Tell matlab to always return a 2d axes object
     )
 
+    # Find ranges across all samples
```
MichaelClerx, line 91: Should this bit go in a private(?) method so that both methods can share it?

MichaelClerx (approved):

pints/plot/_trace.py
```diff
@@ -102,8 +107,8 @@ def trace(
             axes[i, 1].plot(samples_j[:, i], alpha=alpha)
 
             # Set ylim
-            ymin_all = ymin_all if ymin_all < xmin else xmin
-            ymax_all = ymax_all if ymax_all > xmax else xmax
+            ymin_all = ymin_all if ymin_all < xmin[i] else xmin[i]
```
MichaelClerx, line 110: Could this by
```ymin_all = min(ymin_all, xmin[i])``` ?

pints/plot/_histogram.py
ben18785: I mean, it _could_ but I'm not sure it'll be used anywhere else, so think can just leave it as is.

pints/plot/_trace.py
ben18785: Done -- thanks

PR #1405, opened 2021-10-20
Wragged output methods

ben18785:
Implements `SubProblem` and `ProblemCollection` methods which collectively allow:

- problems to be constructed where different outputs are measured at different times
- separate subproblems to be created for different subsets of outputs (and, accordingly, for inference, given different score functions or log-likelihoods)
- caching of simulation outputs between evaluations for the same parameter set


docs/source/core_classes_and_methods.rst
```diff
@@ -4,8 +4,9 @@ Core classes and methods
 
 .. currentmodule:: pints
 
-Pints provides the :class:`SingleOutputProblem` and
-:class:`MultiOutputProblem` classes to formulate
+Pints provides the :class:`SingleOutputProblem`,
+:class:`MultiOutputProblem`, :class:`ProblemCollection`
```
MichaelClerx, line 8: Not a blocker: But can we think of some names that make it clear that:

1. A ProblemCollection can't just contain any old problem. We can't add singleoutput problems, for example
2. The ProblemCollection and SubProblem go together

I'm not sure what these names would be :D 
MultipartProblem and ProblemPart ?

examples/README.md
```diff
@@ -99,6 +99,7 @@ relevant code.
 - [Multiplicative Gaussian noise](./stats/multiplicative-gaussian-errors.ipynb)
 - [Pooling parameters](./stats/pooling.ipynb)
 - [Student-t noise model](./stats/student-t-sampling-error.ipynb)
+- [Wragged output model](./stats/multiple_wragged_outputs.ipynb)
```
MichaelClerx, line 102: I've never come across the word "wragged" before. Is it a technical term?

examples/stats/multiple_wragged_outputs.ipynb
```diff
@@ -0,0 +1,424 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Multi-output model with wragged time measurements\n",
+    "\n",
+    "This notebook describes how to fit a model in PINTS which has multiple outputs with different measurement times for each. Additionally, we illustrate how this same framework allows different measurement models to be used for different outputs within the same forward model when doing inference.\n",
```
MichaelClerx, line 9: This is too quick for me.
Can we change it to say "This notebook describes... for each. This situation arises when..." 1-3 lines on why this would occur, e.g. when you measure different system outputs individually.
Then leave the "additionally" for later, just add in a header and give a 2nd intro?

examples/stats/multiple_wragged_outputs.ipynb
```diff
@@ -0,0 +1,424 @@
...
+    "noisy_values[:, 0] += np.random.normal(0, noise1, len(times))\n",
+    "noisy_values[:, 1] += np.random.normal(0, noise2, len(times))\n",
+    "noisy_values[:, 2] += np.random.normal(0, noise3, len(times))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we assume that only the first half of time measurements are taken for the first two outputs; and only the second half for the third."
```
MichaelClerx, line 62: Can we make this a bit less artificial? I.e. start with wordy version, saying e.g. we measure X every 6 hours, but Y only once a day. Then show how that leads to code like given below?

docs/source/core_classes_and_methods.rst
MichaelClerx: Maybe get the word "output" into the name for ProblemCollection, to show that all 3 problem types are actually classified by their output. So SingleOutput, MultiOutput, and... SplitOutputProblem? ComposedOutputProblem?

MichaelClerx: Since the user never creates a SubProblem, (and since it is a very different thing from the other 3 classes called XProblem), is this a rare case for having a class inside a class?
E.g.
```
class ProblemCollection(object):
    ...
    
    class SubProblem(object):
        ...
        
    def subproblem(i):
        return ProblemCollection.SubProblem(...)
```

examples/stats/multiple_wragged_outputs.ipynb
```diff
@@ -0,0 +1,424 @@
...
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "# solve system\n",
+    "sol_12 = problem_0.evaluate(real_parameters)\n",
+    "\n",
+    "## this uses cached result from first solve under the hood\n",
```
MichaelClerx, line 149:
```suggestion
    "# this uses cached result from first solve under the hood\n",
```

examples/stats/multiple_wragged_outputs.ipynb
```diff
@@ -0,0 +1,424 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 35,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "class CombinedLogLikelihood(pints.LogPDF):\n",
```
MichaelClerx, line 246: This is exactly the kind of thing I'd want PINTS to handle for me, after I've gone to the trouble of setting up the special problem class!

MichaelClerx (changes requested):
Hi Ben,

It looks quite good, but I'm a bit confused that the culmination of all of this is a custom-made likelihood class. I've thought about it a bit, and can't see any easy way out other than adding a Composed/Combined/Composite Likelihood, Prior, and Error in the same PR?

We've got a ComposedLogPrior already: https://github.com/pints-team/pints/blob/master/pints/_log_priors.py#L170

Issue #1408, opened 2021-10-31
OptimisationController: Callback method per iteration


MichaelClerx:
Sometimes, instead of taking over entirely, a user might want to run some simple code at each iteration, e.g. to visualise progress without reimplementing stopping criteria etc.

This could be handled with a callback passed to the controller, e.g.

```
def cb(opt):
    plot(opt.xbest())

opt = pints.OptimisationController(x0, boundaries=b, method=pints.CMAES, callback=cb)
opt.run()
```
or
```
opt = pints.OptimisationController(x0, boundaries=b, method=pints.CMAES)
opt.run(callback=cb)
```
or more in keeping with the rest of the customisation options
```
opt = pints.OptimisationController(x0, boundaries=b, method=pints.CMAES)
opt.set_callback(cb)
opt.run()
```

Thoughts @ben18785 @martinjrobins ?

MichaelClerx:
Closed in #1409

PR #1409, opened 2021-10-31
Callback function in optimisation controller.

MichaelClerx:
See #1408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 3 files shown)


DavAug (approved):
I like it! Thanks @MichaelClerx .  Just checking: the callback is always executed for each proposal in a sequential order right? So the callback can automatically infer which iteration the optimiser is at. Otherwise it might be useful to pass the iteration also as input to the callback.

DavAug:
> > I like it! Thanks @MichaelClerx . Just checking: the callback is always executed for each proposal in a sequential order right? So the callback can automatically infer which iteration the optimiser is at. Otherwise it might be useful to pass the iteration also as input to the callback.
> 
> It should be called in order, yeah, but maybe it makes sense to add iteration anyway, to save the user some work?

Yep sounds like a good idea! 🥳

DavAug:
> @DavAug done!

Great thanks!

PR #1413, opened 2021-11-08
ABC Routine

phumtutum:
- Added an ABC base class + controller
- Added an implementation of the ABC rejection algorithm
- Added tests for those classes
- Added a jupyter notebook to highlight the rejection algorithm, here is the resulting plot:
output

PR #1417, opened 2021-11-20
Generic stochastic model

phumtutum:
In this pr:
- A generic stochastic model is added (MarkovJumpModel) which uses Gillespie's algorithm
- An example of a model that illustrates how simply other stochastic models can be added from now on (Michaelis Menten model)
- Added a degradation model for testing as well
- Removed old degradation model


phumtutum:
Now, the old degradation model is removed and the pr is ready for review

docs/source/toy/stochastic/index.rst
```diff
@@ -0,0 +1,15 @@
+***********************
+Stochastic Toy Problems
+***********************
+
+The `stochastic` module provides toy :class:`models<pints.toy.stochastic.MarkovJumpModel>`,
```
MichaelClerx, line 5: I guess initially it's just models?

examples/README.md
```diff
@@ -117,6 +117,7 @@ relevant code.
 - [SIR Epidemiology model](./toy/model-sir.ipynb)
 - [Stochastic Degradation model](./toy/model-stochastic-degradation.ipynb)
 - [Stochastic Logistic model](./toy/model-stochastic-logistic-growth.ipynb)
+- [Michaelis Menten model](./toy/model-stochastic-michaelis-menten.ipynb)
```
MichaelClerx, line 120: Separate category here too?

phumtutum:
This pr updates the stochastic degradation model. This model is used in a ipynb in https://github.com/pints-team/pints/pull/1413 and therefore changes need to be made to that ipynb in this pr once that pr is merged into master. I will make the requested changes at once with those changes.

pints/toy/stochastic/_markov_jump_model.py
```diff
@@ -0,0 +1,149 @@
+#
+# Stochastic degradation toy model.
```
MichaelClerx, line 2:
```suggestion
# Markov jump model.
```

pints/toy/stochastic/_michaelis_menten_model.py
```diff
@@ -0,0 +1,44 @@
+#
+# Stochastic degradation toy model.
```
MichaelClerx, line 2:
```suggestion
# Stochastic michaelis-menten toy model.
```

MichaelClerx (approved):
Fine with me, if @ben18785 likes it!

examples/README.md
ben18785: Yes, I think separate out the stochastic ones into their own subsection of "toy models".

pints/toy/stochastic/_degradation_model.py
```diff
@@ -0,0 +1,87 @@
...
+        k = parameters[0]
+
+        if k <= 0:
+            raise ValueError('Rate constant must be positive.')
+
+        times = np.asarray(times)
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+
+        variance = np.exp(-2 * k * times) * (-1 + np.exp(k * times)) * self._x0
```
ben18785, line 78: I checked -- this is correct

pints/toy/stochastic/_degradation_model.py
```diff
@@ -0,0 +1,87 @@
...
+        k = parameters[0]
+
+        if k <= 0:
+            raise ValueError('Rate constant must be positive.')
+
+        times = np.asarray(times)
+        if np.any(times < 0):
+            raise ValueError('Negative times are not allowed.')
+
+        mean = self._x0 * np.exp(-k * times)
```
ben18785, line 58: Correct

pints/toy/stochastic/_markov_jump_model.py
```diff
@@ -0,0 +1,149 @@
...
+import random
+
+from .. import ToyModel
+
+
+class MarkovJumpModel(pints.ForwardModel, ToyModel):
+    r"""
+    A general purpose Markov Jump model used for any systems of reactions
+    that proceed through jumps. We simulate a population of N different species
+    reacting through M different mechanisms
```
ben18785, line 20: "reacting through M different reaction equations"

pints/toy/stochastic/_markov_jump_model.py
```diff
@@ -0,0 +1,149 @@
...
+
+    1. Sample values :math:`r_0`, :math:`r_1`, from a uniform distribution
+
+    .. math::
+        r_0, r_1 \sim U(0,1)
+
+    2. Calculate the time :math:`\tau` until the next single reaction as
+
+    .. math::
+        \tau = \frac{-\ln(r)}{a_0}
```
ben18785, line 32: We haven't said what a_0 is.

pints/toy/stochastic/_markov_jump_model.py
```diff
@@ -0,0 +1,149 @@
...
+
+    3. Decide which reaction, i, takes place using r_1 * a_0 and iterating
+
+    through propensities
```
ben18785, line 36: This needs quite a bit more explanation, especially since this is the key point of the algorithm

ben18785 (changes requested):
Thanks @phumtutum -- looks really good. There are a few things that I think would be good to change, then I'm happy to merge.

In the degradation notebook:

- I think this line needs rewriting: "averaging the concentration values at each time step, produces a reproducible result" -- I don't think that "reproducible result" is the correct terminology here. I guess you just want to say that the mean of the distribution is that exponential thingy?

- "The deterministic mean (from above) is plotted with the deterministic standard deviation" -- I think this is a bit ambiguous. I think you want to say, "We now plot the analytic mean and standard deviation of this process."

In the michaelis-menten notebook:

- "The Michaelis Menten model is a stochastic model, more details can be found here.": the MM model actually isn't necessarily a stochastic model (we have a deterministic version of a similar but not same version of the model in the package), and the stuff you link to is about the deterministic model. So, can we say, "This is a stochastic version of the Michaelis-Menten model, and more details about the deterministic analogue can be found here."
- Can you also provide a little bit of a description about what this model represents in the first part of this notebook?
- "This function, given a set of parameters and times, computes the appropriate times and values using Gillispie's algorithm and then uses interpolation to find the values at the exact times requested" -- it's not clear what the values are (i.e. they are counts of molecules) and "Gillespie" has a typo. I think you also need to explain what the interpolation does and why it is needed. I've noticed "Gillespie" is misspelled throughout the notebook actually.
- "Given the stochastic nature of this model we can use multiple simulations to make sure that the runs are covering the same model with the same parameters. Our simulations are close, suggesting we are obtaining accurate simulations." -- this isn't correct, really. All you are showing is that there is not a large amount of stochasticity in these models for these initial values. Indeed, can we show that when we use much smaller molecule counts as initial conditions that we get much more stochasticity?
- It would be good to show that the solutions with high molecular counts correspond with the solutions of the corresponding ODE: you could code this up using scipy's `odeint` within the notebook

phumtutum:
@ben18785 I have addressed your comments. If there is anything else I should change please let me know.

ben18785 (approved):

Issue #1411, opened 2021-11-02
xbest and fbest returned by ES methods


MichaelClerx:
CMAES, xNES, SNES, and our BareCMAES all maintain a "best point seen" as well as a "distribution mean".
The idea behind the methods is that the distribution mean ends up being the best guess for the optimum, even if the error at the "best point seen" was lower.
In problems I saw so far, the distinction was minimal and the two points pretty much overlapped.

In our interface, we have methods `xbest` and `fbest` to return the "best" points found.
Having the mean be approximately equal to the best seen was convenient, as we could return `(xbest, fbest)` instead of `(mu, f(mu))`, which requires an extra evaluation of `f`.

Now I'm seeing some problems where the two are quite different though.

It leads to slightly confusing situations, e.g. the last 6 points evaluated were:
```
  -251010.2,  528860.6
  -251010.2,  528860.6
  -251010.2,  528860.6
  -251010.2,  528860.6
  -251010.2,  528860.6
  -251010.2,  528860.6
```
(with some rounding!)
while the best point returned was:
```
-243985.5,  542142.3
```
A bit of checking showed that the best point wasn't even a local optimum, just a point that happened to occur in one of the searches and had the best score yet.

So what do we do?

I think we should

1. Update all these methods to return the distribution mean as `xbest`
2. Update the interface for these methods to allow `fbest` to be `None`
3. Possibly update the interface with an extra method `xbest_seen` or something similar?

Both results are useful: the mean is meant to be a noise-filtered optimum, but depending on your data set you might not really care for noise filtering?

MichaelClerx:
Thoughts about this @ben18785 @chonlei @martinjrobins @alisterde @mirams ?

MichaelClerx:
the pycma code internally refers to these two as the "best" (lowest error) and "favorite" (current mean)

MichaelClerx:
An easy "fix" could be to add a method `xfavourite()` that returns a favourite, if the method has one, and the update the docs for `xbest` to say this is the best point seen, and may differ from the current best estimate of where the optimum is?

mirams:
You definitely want the option to get hold of the best point, but I probably would expect these to return their defaults/favourites for the algorithms if I was using it out of the box

mirams:
(I actually spent a while trying to figure out how to get pints to give me the best point because I assumed that it was giving me the CMA-ES favourite yesterday!)

MichaelClerx:
[...]

Issue #25, opened 2017-07-24
Investigate different methods of implementing boundaries


MichaelClerx:
Two options:

1. Implement a periodic transform on the parameter space (strategy used by CMA-ES, for example)
2. Implement a transform on the score function

MichaelClerx:
Wrote a short script to explain triangle-transform (for my pints presentation today):

    #!/usr/bin/env python
    import numpy as np
    import matplotlib.pyplot as pl
    from matplotlib.patches import Rectangle

    lower = -4
    upper = 6
    r = upper - lower

    def transform(x):
        y = np.remainder(x - lower, 2 * r)
        z = np.remainder(y, r)
        return ((lower + z) * (y < r) + (upper - z) * (y >= r))

    x = np.linspace(-16, 18, 1000)
    y = transform(x)

    x2 = np.linspace(lower, upper)
    y2 = transform(x2)

    pl.figure()
    pl.xlabel('Parameters the optimiser sees')
    pl.ylabel('Parameters the model sees')
    pl.gca().add_patch(Rectangle((lower, upper), r, -r, alpha=0.1))
    pl.plot(x, y)
    pl.plot(x2, y2, lw=3)

    def f(x):
        return 0.02 * (x*2)**2 + 1.3 * np.sin((x - 0.5) * 1.4)


    pl.figure()
    pl.xlabel('Parameters the optimiser sees')
    pl.ylabel('Score the optimiser sees')
    pl.ylim(-2,  4.5)

    pl.gca().add_patch(Rectangle((lower, upper + 3), r, -14, alpha=0.1))

    x3 = np.linspace(lower - 5, upper + 5, 1000)
    pl.plot(x3, f(x3), 'g--')

    pl.plot(x, f(transform(x)), lw=3)

    pl.show()

![triangle-1 svg](https://user-images.githubusercontent.com/517644/31717142-b037a168-b402-11e7-8002-151c7b6bae54.png)

![triangle-2 svg](https://user-images.githubusercontent.com/517644/31717143-b04edb8a-b402-11e7-8712-d7c9d15698c3.png)

MichaelClerx:
Some details:
- the transform creates infinite local minima, but as they all map to the same real parameter values this isn't a problem!
- the resulting landscape is continuous, but not smooth
- at the borders, you may get fake minima (example on the left) and maxima (example on the right), but this shouldn't be a huge problem for the optimisers. Especially since the best value usually won't be on the borders, but somewhere in the centre

[...]

PR #1420, opened 2021-11-30
Optimisers: best vs guessed position

MichaelClerx:
See #1411

Changes:

- The pints.Optimiser interface now has methods x_best, f_best, x_guessed, and f_guessed (replacing the old methods xbest and fbest)
- By default the `_guessed` methods just call `_best`
- But the ES methods return the mean of their proposal distribution
- The controller still uses `_best` by default, but this can be adjusted with `set_f_guessed_tracking`
- In a future PR, the controller should start using `_guessed` by default


MichaelClerx:
Uncovered an issue in xNES, and probably SNES (and PSO?) to do with the triangle transform used to check rectangular boundaries: see #25

As a result, have kept the strategy on f_best for now. Will need to address this in future PRs.

MichaelClerx:
The notebooks are not substantially changed except for `ask-and-tell.ipynb`, which now includes some visualisations of the difference between the best and guessed value.

DavAug (commented):
Thanks @MichaelClerx ! Very useful to be able to distinguish between the best seen and best guess location.

Just checking: If CMAES previously returned its best guess as 'fbest' and we now change to returning 'f_best' is this a desired change in the algorithm (even if it's only temporary and planned to be changed in a future PR)?

MichaelClerx:
> Just checking: If CMAES previously returned its best guess as 'fbest' and we now change to returning 'f_best' is this a desired change in the algorithm (even if it's only temporary and planned to be changed in a future PR)?

The way we had implemented it, the `fbest` returned was the best seen, not the best guessed. This is probably not desirable, as it can lead to situations like this one:

local-map-zoom-xbest

Here the orange circle is the returned x_best (best seen), but the blue crosses show that the method has converged to a different point (x_guessed)

MichaelClerx:
So I really want to set it to returning the best guessed point instead. But at the moment that just exposes a bug in xnes and makes lots of tests fail, so I want to fix that first 😓😓😓

DavAug:
> > Just checking: If CMAES previously returned its best guess as 'fbest' and we now change to returning 'f_best' is this a desired change in the algorithm (even if it's only temporary and planned to be changed in a future PR)?
> 
> The way we had implemented it, the `fbest` returned was the best seen, not the best guessed. This is probably not desirable, as it can lead to situations like this one:
> 
> local-map-zoom-xbest
> 
> Here the orange circle is the returned x_best (best seen), but the blue crosses show the requested points, from which you can see that the method has converged to an x_guessed != x_best_seen

That is very interesting! So just to clarify: the black points are the actual steps taken by CMAES and the blue crosses are the samples points to estimate the gradient? Or are the blue crosses really just guesses and if so, why are there more guesses than steps?

MichaelClerx:
> That is very interesting! So just to clarify: the black points are the actual steps taken by CMAES and the blue crosses are the sampled points which are used to estimate the gradient? Or are the blue crosses really just guesses and if so, why are there more guesses than steps?

Sorry I didn't explain that well did I :D 
The black dots are the successive "best seen" points, so the ones you obtain if you call x_best() at each iteration. The blue crosses meanwhile are the evaluated points. The mean of the proposal distribution isn't plotted here, but you can see that it doesn't overlap with the black dots. And at the end of the optimisation the proposal distribution is centered in the area with very dense blue crosses. So that's where CMAES thinks the optimum is, and where its putting all its effort. But this is not the point we are returning at the moment.

There are more examples of this in the updated ask-and-tell.ipynb notebook. Only in these problems (and I suspect in most problems we've looked at) the proposal distribution does end up converging to the best-seen score.

DavAug:
> > That is very interesting! So just to clarify: the black points are the actual steps taken by CMAES and the blue crosses are the sampled points which are used to estimate the gradient? Or are the blue crosses really just guesses and if so, why are there more guesses than steps?
> 
> Sorry I didn't explain that well did I :D
> 
> The black dots are the successive "best seen" points, so the ones you obtain if you call x_best() at each iteration (although this point stays the same for several iterations at a time, unlike the mean of the proposal distribution, which bounces around a lot).
> 
> The blue crosses meanwhile are the evaluated points.
> 
> The mean of the proposal distribution isn't plotted here, but you can see where it ends up because all the blue crosses converge to a single area. So that (x_guessed) is where CMAES thinks the optimum is, and where its putting all its effort. But this is not the point we are returning at the moment (x_best).
> 
> There are more examples of this in the updated ask-and-tell.ipynb notebook. Only in these problems (and I suspect in most problems we've looked at) the proposal distribution does end up converging to the best-seen score.

Is it possible at all that the in this example the CMAES algorithm decides that it has converged based on the evaluated points, but doesn't updated updated the final 'xbest', and we return the penultimate 'xbest'? I just find it suspicious that the steps never get close to the evaluated points.

DavAug (approved):

MichaelClerx:
Good question, I'll try and do a plot of both points

PR #1424, opened 2021-12-07
Stop using triangle wave transform in optimisations

MichaelClerx:
See #25 (and #1420)

It leads to errors with xnes, where it just keeps increasing sigma because it keeps finding points in the next "period" of the search space. The `TriangleWaveTransform` is used by PSO, xNES, SNES, and BareCMAES. All of these methods can easily deal with "inf" for any evaluation, and this was already the way we handled arbitrary shaped boundaries.

In addition, found and fixed a bug in PSO where we (I) initialised with too many particles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,19 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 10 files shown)


mirams (approved):
I have never liked the triangle transform for this reason!

PR #1430, opened 2022-01-19
Issue 1427 pickling no u turn mcmc

### LINKED ISSUE #1427: Refactor NoUTurnMCMC for pickling
High dimensional or hierarchical problems can take quite long to run O(hours to days) and the NUTS sampler can currently not be pickled because it relies on async generators. That means that we can't run, pickle and resume the inference at a later time.

The core of the ``NoUTurnMCMC`` is the ``nuts_sampler`` generator which maintains the state of the sampler in form of instances of the ``DualAveragingAdaption`` and the ``NutsState``. The ``nuts_sampler`` is a property of the ``NoUTurnMCMC``
which will have to be removed for pickling. However, then the state of the sampler (``DualAveragingAdaption``, ``NutsState``) is lost. 

I suggest to refactor the ``NoUTurnMCMC`` such that it maintains its state (``DualAveragingAdaption``, ``NutsState``) itself. If we then remove the ``nuts_sampler`` for pickling, it is possible to reconstruct the same ``nuts_sampler`` using the state.

DavAug:
Closes #1427 .

### DIFF
--- a/pints/_mcmc/_dual_averaging.py
+++ b/pints/_mcmc/_dual_averaging.py
@@ -89,6 +89,7 @@ class DualAveragingAdaption(object):
         self._epsilon = init_epsilon    # The adapted step size
         self._mass_matrix = None        # The adapted mass matrix (set below)
         self._inv_mass_matrix = None    # The inverse adapted mass matrix
+        self._use_dense_mass_matrix = None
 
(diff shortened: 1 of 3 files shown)


martinjrobins (approved):
This looks great, thanks @DavAug, happy to merge

PR #1432, opened 2022-02-03
Added additional stochastic models[finished]

phumtutum:
https://github.com/pints-team/pints/pull/1417 added MarkovJumpModels which can be used to add stochastic models more easily. This PR:
- removes the old stochastic logistic model and adds implements it in the new way
- will add the stochastic simulation of production and degradation model
- Schlogl's model
- remove MarkovJumpModel's return of a zero array if the starting population is empty


phumtutum:
This PR is ready for a review. As in #1413, some unit tests are failing, however, I think this is not because of this PR.

MichaelClerx (approved):

PR #1433, opened 2022-02-04
Drop support for Python 3.5

MichaelClerx:
3.5 is no longer testable on github actions.

~I've also removed the explicit python version set in many github actions, which makes it default to a sensible latest/stable one (3.10)~. Couldn't add testing on 3.10, as `pystan` fails to build, which causes all our tests to error: https://github.com/pints-team/pints/runs/5066912781?check_suite_focus=true Have reported it here: https://github.com/stan-dev/pystan/issues/339 and made a separate issue for 3.10 support here: #1434

### DIFF
--- a/.github/workflows/copyright-test.yml
+++ b/.github/workflows/copyright-test.yml
@@ -22,10 +22,9 @@ jobs:
     steps:
       - uses: actions/checkout@v1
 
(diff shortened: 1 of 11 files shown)


fcooper8472 (approved):
Looks good to me! Always exciting to see old versions dropped

Issue #1333, opened 2021-03-29
Update stan interface to use pystan 3


ben18785:
Think this is because Pystan has [recently been updated](https://pystan.readthedocs.io/en/latest/upgrading.html) and now is imported with `import stan`.

I think it [may](https://pystan.readthedocs.io/en/latest/installation.html) also be installed via `python3 -m pip install pystan`.

ben18785:
Also looks like this was a breaking release in terms of [how models are compiled](https://pystan.readthedocs.io/en/latest/upgrading.html#upgrading).

ben18785:
It also looks like it only works on Python>=3.7... @MichaelClerx @martinjrobins what do you think is best here? Keep this in PINTS or move to a separate repo?

MichaelClerx:
I'm happy to keep it in and only test on newer pythons (see nuts for an example)

The minimum Python 3 we can test on at the moment is Python 3.5. It probably won't be too long before that becomes 3.6...

MichaelClerx:
No windows support either I see

MichaelClerx:
@ben18785 We urgently need to update to pystan 3. The 2.x version is no longer supported (https://github.com/stan-dev/pystan/issues/339) meaning that we can't test PINTS on Python 3.10 at the moment. #1434

MichaelClerx:
Full to-do:

- [ ] Change the testing so that 3.6 and older no longer run any pystan tests (because pystan 3.x is not available on python 3.6)
- [ ] Update the stan interface to use pystan 3.x

ahartikainen:
Hey, just wanted to help the transition.

What interfaces do you need? 

Only sampling or also manual logp calculations?

ben18785:
Thanks @ahartikainen -- that's a very kind offer.

In PINTS, we have used Pystan to allow models to be built using Stan code but sampled from using PINTS' sampling algorithms. To do so, we use the `log_prob` and `grad_log_prob` functions (in both cases with `adjust_transform=True`) from Pystan 2.

The code that implements this is [here](https://github.com/pints-team/pints/blob/master/pints/interfaces/stan/_stan.py). As you'll see, at the moment, we do the following:

- at instantiation of our `StanLogPDF` object, we pass the stan code and use this to compile a Pystan `StanModel` object
- we then can switch between datasets using the `update_data` method, which runs sampling for one iteration and returns a Pystan object
- this object is then used to calculate the log probability and the gradient in this through Pystan's convenience functions

[...]

PR #1435, opened 2022-02-06
Update stan interface to pystan 3

MichaelClerx:
See #1333 

- [x] Update setup.py to install new version
- [x] Update tests to skip pystan on Python 3.6
- [x] Update PINTS code to pystan 3+
- [x] Update example notebook


MichaelClerx:
@ben18785 I've made this PR that has the infrastructure bits sorted out for #1333:

- The tests on Python 3.6 do not install pystan. This is detected by the unit tests, so that the stan-related tests are skipped and the unit tests pass on Python 3.6 (notice skipped=3 at the bottom)
- On Python 3.7 and higher, pystan is installed, so that the unit tests are run. Because these expect pystan 2 but we've installed pystan 3 the tests fail

So all ready to adapt to the new interface!

MichaelClerx:
Hey @ben18785 . I think this might be it? No idea what I'm doing but the tests pass.

Things that are gone in this implementation:
```
            refresh=0,
            control={'adapt_engaged': False})
```
but this is because the entire "fit" operation is skipped now

ben18785 (approved):
Thanks amillion @MichaelClerx -- it all looks good to me. The one thing I have changed is the example notebook as this was using the old interfacing. I wasn't able to check that this ran though as I am having some trouble getting pystan to install on my machine, so would you be able to check it runs for you? After that happy to merge.

ben18785:
Ah ha. Is it because ``get_param_names`` returns a list of length 3 (where one of them is a nested list)? Do we need to flatten this list?

MichaelClerx:
That sounds correct!
Do you want to write & commit a test for this issue first?

(Solution is probaly this:
```
from itertools import chain
list(chain.from_iterable([[180.0], [173.8], [164.2], [156.5], [147.2], [138.2]]))
```)

**Comment by MichaelClerx** [comment]
Hey @ben18785 do you know what's happening here?

With the 8 schools example you gave the current code thinks there's 18 parameters (2*8+1)

**Comment by ben18785** [comment]
The issue is the "transformed parameters" block in this example. The PyStan interface doesn't differentiate between parameters defined in the "parameters" block vs those defined in the "transformed parameters" block. Whereas, PINTS only needs parameters defined in the "parameters" block...

It may be possible that something like this could work based on regex: https://github.com/arviz-devs/arviz/blob/5af74712950c99d486cc0f62db66537670e3bfcc/arviz/data/base.py#L444 but this seems a bit messy!

**Comment by MichaelClerx** [comment]
@ben18785 any way to filter out the transformed parameters?
```
Posterior
================================================================================
constrain_pars
constrained_param_names
data
dims
fixed_param
grad_log_prob
hmc_nuts_diag_e_adapt
log_prob
model_name
param_names
program_code
random_seed
sample
unconstrain_pars

Module
================================================================================
constrained_param_names
fixed_param_wrapper
get_dims
get_param_names
hmc_nuts_diag_e_adapt_wrapper
log_prob
log_prob_grad
model_name
transform_inits
write_array
```

MichaelClerx:
None of those methods seem to help in this case. Hmmm!
I would prefer putting in a request with pystan to add some method that gives us the info we need, instead of trying to parse the stan model code. I think we could hack something together in 15 minutes that would acceptably parse it for our purposes, but then we'd forever be finding small bugs and new cases we hadn't considered, and basically would end up building a stan parser, when a perfectly good stan parser is already available

ben18785:
Thanks @MichaelClerx -- yes, I think that you're right. I suspect the only thing we can do is to allow the user to set the number of parameters. The one thing, as you say, is the order: need to check if this matches the declaration order (or say alphabetical order).

Alternatively, we could just say that we don't handle Stan models which have a transformed parameters block...everything else works fine I think.

MichaelClerx:
Looking in the code I saw quite a few "in the order they were declared". Will do a quick test and then it's fine, I guess :D

MichaelClerx:
Yeah declared order

ben18785:
Cool -- thanks. So I guess, we just pass this onto the user: they have to
ensure the parameters are passed in the order they are declared in the
"parameters" block. (And they have to say how many parameters there are.)

MichaelClerx:
Do we make the user say there are "1+1+8=10" parameters? Or do we let them say "3"?

ben18785:
10 since PINTS doesn't care whether a parameter is a scalar or an element
of a vector...

MichaelClerx:
Alright @ben18785 , what do you think?

ben18785:
Ha, you went and got the parameter counts! Nice one :) A quick question: how does

`len(code.split(';')) - 1`

deal with vector parameters? Doesn't this just count the total number of things (some of which may be vectors) which are declared?

MichaelClerx:
Yeah exactly. I'm assuming each declaration is a scalar or vector variable. The total "flat" variable count depends on the data, so can't be inferred form the code.

...which, come to think of it, means we need to decide what happens if the data changes. Do we

1. Remove `update_data()`. This is the easiest solution, and not necessarily unfriendly to the user, as models are cached by httpstan
2. Check if n_parameters has changed after a call to `update_data()`, and raise an error if so
3. Check if it has changed, and start reporting a different `n_parameters()`. I'm expecting LogPDFs to have a constant param count in their lifetime, but not sure we have made this official?

ben18785:
@MichaelClerx thanks. Good points. I honestly don't see the point of `update_data` now, since, as you say, models are cached. It was useful previously as it would save us having to re-compile needlessly, but now, I vote for removing it.

PR #1439, opened 2022-02-23
Test on Python 3.10

MichaelClerx:
- [x] Now testing on Python 3.10
- [x] Updated rest of 3.9 tests to just use whatever the latest supported version is

@fcooper8472 @martinjrobins what do you think of the 2nd change? Is this a good idea in your opinion (the whole `with` block can be removed in that case, x64 is default too), or are we better off explicitly specifying the newest version we want to test on?

### DIFF
--- a/.github/workflows/copyright-test.yml
+++ b/.github/workflows/copyright-test.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v1
 
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
 
       - name: install pints
         run: |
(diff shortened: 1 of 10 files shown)


MichaelClerx:
Only Python 3.11 to go now, and then we're onto Python 95!

MichaelClerx:
Python for workgroups

fcooper8472:
@MichaelClerx omitting the specific version appears to [default to](https://github.com/actions/setup-python/blob/main/action.yml) `3.x` which means [>=1.0.0 <2.0.0](https://devhints.io/semver) so I think we would lose certainty about which version specifically is being tested on.  That doesn't necessarily matter...

The other thing is we should change all `actions/setup-python@v1` to `actions/setup-python@v2` as there have been numerous enhancements to that functionality since we originally wrote our GitHub actions scripts.

fcooper8472 (approved):

PR #1440, opened 2022-03-17
Added URL meta data to setup.py

MichaelClerx:
These will show up on pypi.

As it's not a change to PINTS itself I propose to leave this out of the changelog.


phumtutum (approved):
Looks good

MichaelClerx:
Thanks @phumtutum !

PR #1442, opened 2022-03-24
ABC SMC/PMC

phumtutum:
Added the ABC SMC (https://doi.org/10.1098/rsif.2008.0172) / ABC PMC (https://arxiv.org/abs/0805.2256).

NOTE: This replaces #1055 .

This algorithm was tested on the degradation model and on the production and degradation model by sampling, computing the KDE and then computing the KL-divergence from the true posterior.

For the degradation model with 1000 samples and intermediate distributions of size 200 we get a KL-divergence of 0.706, with intermediate distributions of size 1000 we get 0.671, while using the error threshold schedule [0.9, 0.8, 0.7] for both. For comparison, using the rejection ABC with the error threshold 0.7 and 1000 samples, we get 0.81 (implying the ABC SMC performs better).

For the production and degradation model with 1000 samples and intermediate distributions of size 200 we get a KL-divergence of 0.33, using intermediate distributions of size 1000, we get 0.32, while using the error threshold schedule [9, 6, 4.4]. For comparison, using the rejection ABC with error threshold 4.4 and 1000 samples we get 0.27.

Those experiments were performed on the all-abc branch, in the (messy) jupyter notebooks ./degradation.ipynb and ./production_degradation.ipynb . Additional experiments which show the convergence for error thresholds were done in ./degradation-errors.ipynb ./production_degradation-errors.ipynb. If we consider smaller and smaller thresholds we get a smaller and smaller KL-divergence. Here are the charts (x-axis = error threshold; for the PMC/SMC the error threshold of [1.5 * error_threshold, 1.3 * error_threshold, error_threshold] were used; y-axis = KL-divergence), where modified PMC = using 200 as the intermediate distribution sizes and original PMC = using 1000:

[degradation_errors_all.pdf](https://github.com/pints-team/pints/files/8346236/degradation_errors_all.pdf)

[production_degradation_errors_all.pdf](https://github.com/pints-team/pints/files/8346240/production_degradation_errors_all.pdf)


MichaelClerx:
Does this replace the https://github.com/pints-team/pints/pull/1055 branch? Looks like a slightly different method (or at least a difference reference)?

MichaelClerx:
Does that mean we need a more specific name than ABC SMC ?

phumtutum:
This does replace the #1055 . ABC SMC is enough of a specific name I think.

phumtutum:
Now the tests and docs are fixed so this branch can be reviewed.

docs/source/index.rst
```diff
@@ -83,6 +83,8 @@ Sampling
 
    - :class:`RejectionABC`, requires a :class:`LogPrior` that can be sampled
      from and an error measure.
+   - :class:`ABCSMC`, requires a :class:`LogPrior` that can be sampled from
+     from and an error measure.
```
MichaelClerx, line 87: Missed this the last time, but for both ABCs this should be:
- an :class`ErrorMeasure`
- an :class`ErrorMeasure` to ...

docs/source/abc_samplers/index.rst
```diff
@@ -13,4 +13,5 @@ given a :class:`LogPrior` and a :class:`ErrorMeasure`.
 .. toctree::
 
     base_classes
-    rejection_abc
\ No newline at end of file
+    rejection_abc
```
MichaelClerx, line 16: I understand rejection is simpler, but these should probably be alphabetical (consistent with rest and sustainable if we have 20 abc methods)

examples/README.md
```diff
@@ -79,6 +79,7 @@ relevant code.
 
 ### ABC
 - [Rejection ABC sampling](./sampling/rejection-abc.ipynb)
+- [ABC-SMC sampling](./sampling/abc-smc.ipynb)
```
MichaelClerx, line 82: Alphabetical (see nested and mcmc methods above)

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
+#
+# ABC SMC method
+#
+# This file is part of PINTS (https://github.com/pints-team/pints/) which is
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+from __future__ import absolute_import, division
```
MichaelClerx, line 8: We've stopped supporting Python 2 so these can go

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+    ABC-SMC Algorithm  See, for example, [1]_. In each iteration of the
+    algorithm, the following steps occur::
+
+        theta* ~ p_(t-1)(theta), i.e. sample parameters from previous
+            intermediate distribution
+        theta** ~ K(theta|theta*), i.e. perturb theta* to obtain to new point
+        x ~ p(x|theta**), i.e. sample data from sampling distribution
+        if s(x) < threshold_(t), theta* added to list of samples[t]
+
+    After we have obtained nr_samples samples, t is advanced, and weights
```
MichaelClerx, line 25: nr_samples has not been introduced. Will it be used again? If it's code it should be ``nr_samples``

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+import pints
+import numpy as np
+
+
+class ABCSMC(pints.ABCSampler):
+    """
+    ABC-SMC Algorithm  See, for example, [1]_. In each iteration of the
+    algorithm, the following steps occur::
+
+        theta* ~ p_(t-1)(theta), i.e. sample parameters from previous
```
MichaelClerx, line 19: Should this be a maths block instead?

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+import pints
+import numpy as np
+
+
+class ABCSMC(pints.ABCSampler):
+    """
+    ABC-SMC Algorithm  See, for example, [1]_. In each iteration of the
```
MichaelClerx, line 16: The first paragraph of every docstring should be a single-line, stand-alone explanation. So can't refer to things, can't be 2 sentences etc.
The reason for this is that IDEs (and some python help functions) will show the first line of docstrings to help the user remember what everything is.

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+    After we have obtained nr_samples samples, t is advanced, and weights
+    are calculated for samples[t-1]. At the last value for threshold,
+    samples are returned whenever they are accepted. This algorithm is
+    also referred to as ABC Population Monte Carlo (ABC PMC) [2].
+
+    References
+    ----------
+    .. [1] "Toni, Tina, et al. Approximate Bayesian computation scheme
+            for parameter inference and model selection in dynamical systems.
+            Journal of the Royal Society Interface, 6.31: 187-202, 2009.
```
MichaelClerx, line 34: Give https://doi.org/doidoidoi for both

MichaelClerx: (Or arxiv ones, if that's what you used)

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+
+    .. [2] "Beaumont, Mark A., et al. Adaptive approximate Bayesian
+            computation. Biometrika, 96.4: 983-990, 2009."
+    """
+
+    def __init__(self, log_prior, perturbation_kernel=None):
+
+        self._log_prior = log_prior
```
MichaelClerx, line 42: Add some comments so you'll be able to fix this in 10 years time

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+            intermediate distribution
+        theta** ~ K(theta|theta*), i.e. perturb theta* to obtain to new point
+        x ~ p(x|theta**), i.e. sample data from sampling distribution
+        if s(x) < threshold_(t), theta* added to list of samples[t]
+
+    After we have obtained nr_samples samples, t is advanced, and weights
+    are calculated for samples[t-1]. At the last value for threshold,
+    samples are returned whenever they are accepted. This algorithm is
+    also referred to as ABC Population Monte Carlo (ABC PMC) [2].
+
```
MichaelClerx, line 29: Needs argument list!

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+        dim = log_prior.n_parameters()
+
+        if perturbation_kernel is None:
+            self._perturbation_kernel = pints.MultivariateGaussianLogPrior(
+                np.zeros(dim),
+                0.001 * np.identity(dim))
+        elif isinstance(perturbation_kernel, pints.LogPrior):
+            self._perturbation_kernel = perturbation_kernel
+        else:
+            raise ValueError("Provided perturbation kernel must be an instance\
```
MichaelClerx, line 62: Not sure how this renders. WIll it have lots of spaces? Usually, you'd break a string like this:
```
        raise ValueError(
            'Provided perturbation kernel must be an instance of'
            ' pints.LogPrior')
```
or
```
        raise ValueError('Provided perturbation kernel must be an instance of'
                         ' pints.LogPrior')
```

MichaelClerx: Note also that (except for docstrings) we use ' instead of " wherever possible (saves us all a bit of shift-key rsi)

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,176 @@
...
+
+            w = prior_prob / sum(mw)
+            new_weights.append(w)
+        return new_weights
+
+    def set_threshold_schedule(self, schedule):
+        """
+        Sets a schedule for the threshold error distance that determines if a
+        sample is accepted (if error < threshold).
+        Schedule should be a list of epsilon values
```
MichaelClerx, line 162: 1. Needs to be a new paragraph (see comment above).
2. What's an "epsilon value" ?

pints/tests/test_abc_smc.py
```diff
@@ -0,0 +1,159 @@
...
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+import pints
+import pints.toy as toy
+import unittest
+import numpy as np
+import pints.toy.stochastic
+
+# Consistent unit testing in Python 2 and 3
```
MichaelClerx, line 15: We've dropped python 2 support

pints/tests/test_abc_smc.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python
+#
+# Tests the basic methods of the adaptive covariance MCMC routine.
+#
+# This file is part of PINTS (https://github.com/pints-team/pints/) which is
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+import pints
```
MichaelClerx, line 9: Please order these alphabetically.

(Ideally, you'd also go 1. standard library, empty line, 2. used libraries, empty line, 3. local imports (pints))

pints/tests/test_abc_smc.py
```diff
@@ -0,0 +1,159 @@
...
+    unittest.TestCase.assertRaisesRegex
+except AttributeError:
+    unittest.TestCase.assertRaisesRegex = unittest.TestCase.assertRaisesRegexp
+
+
+class TestABCSMC(unittest.TestCase):
+    """
+    Tests the basic methods of the ABC SMC routine.
+    """
+# Set up toy model, parameter values, problem, error measure
```
MichaelClerx, line 26: Why isn't this the docstring for setUpClass ?

MichaelClerx (changes requested):
Great start! Some early feedback on code style and docs & such

pints/tests/test_abc_smc.py
phumtutum: I have opted for going with standard library, empty line, used library, empty line and local imports

ben18785 (approved):
I just had a look through and (uncharacteristically for me) I have nothing to add. Think it looks great. Thanks @phumtutum

phumtutum:
Thanks Ben! @MichaelClerx let me know if there is anything else you would like me to change on this PR.

pints/tests/test_abc_smc.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python
+#
+# Tests the basic methods of the adaptive covariance MCMC routine.
```
MichaelClerx, line 3:
```suggestion
# Unit tests for the SMC ABC class.
```

MichaelClerx:
What does the "trying t=1" stuff at the start mean? Why aren't we seeing a message like that for t=0 3 ?
Also should it not be something like `i = 200, threshold=1.5` ?

```
Running...
Using ABC-SMC
Running in sequential mode.
Trying t=1, threshold=1.5
Trying t=2, threshold=1.0
Iter. Eval. Acceptance rate Time m:s
1     2025   0.00049382716    0:01.8
2     2054   0.000973709834   0:01.9
3     2066   0.00145208132    0:01.9
```

I'm also curious about this bit:
```
Iter. Eval. Acceptance rate Time m:s
1     2025   0.00049382716    0:01.8
2     2054   0.000973709834   0:01.9
3     2066   0.00145208132    0:01.9
20    2646   0.00755857899    0:02.2
40    3300   0.0121212121     0:02.6
60    4348   0.013799448      0:03.1
80    5075   0.0157635468     0:03.5
100   5687   0.0175839634     0:03.9
120   6377   0.0188176258     0:04.3
140   6955   0.0201294033     0:04.6
160   7592   0.0210748156     0:04.9
180   8053   0.0223519185     0:05.2
200   8735   0.0228963938     0:05.6
220   9380   0.0234541578     0:05.9
240   10165  0.0236104279     0:06.4
260   10881  0.0238948626     0:06.8
280   11510  0.0243266725     0:07.2
300   12359  0.0242738086     0:07.6
320   13093  0.0244405407     0:08.1
340   13951  0.0243710128     0:08.5
360   14770  0.0243737305     0:09.0
380   15481  0.0245462179     0:09.4
400   16148  0.0247708695     0:09.8
420   16822  0.0249673047     0:10.1
440   17485  0.0251644267     0:10.5
460   18470  0.0249052518     0:11.0
480   19275  0.0249027237     0:11.4
500   20037  0.0249538354     0:11.8
520   20750  0.025060241      0:12.2
540   21577  0.0250266487     0:12.7
```
why don't we see a big change in acceptance rate at the 200th and 400th iterations?

MichaelClerx:
I think we need a 3d section in the notebook where you don't use the controller but ask-tell directly, and plot histograms for the three different "waves" (0-200, 200-400, 400-1000)

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+# released under the BSD 3-clause license. See accompanying LICENSE.md for
+# copyright notice and full license details.
+#
+import pints
+import numpy as np
+
+
+class ABCSMC(pints.ABCSampler):
+    r"""
+    Implements the ABC-SMC algorithm as describd in [1].
```
MichaelClerx, line 14:
```suggestion
    Implements the ABC-SMC algorithm as described in [1].
```

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+    .. [1] "Toni, Tina, et al. Approximate Bayesian computation scheme
+            for parameter inference and model selection in dynamical systems.
+            Journal of the Royal Society Interface, 6.31: 187-202, 2009.
+            https://doi.org/10.1098/rsif.2008.0172
+
+    .. [2] "Beaumont, Mark A., et al. Adaptive approximate Bayesian
+            computation. Biometrika, 96.4: 983-990, 2009."
+            https://doi.org/10.48550/arXiv.0805.2256
+
+    Parameters
```
MichaelClerx, line 46: A ``Parameters`` block should come before the references, and should list the arguments taken by a method, or in this case by the class constructor. So `log_prior` and `perturbation_kernel`.

Number of samples and  error schedule should be explained in the text

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+        self._accepted_count = 0
+        self._weights = []
+        self._xs = None
+        self._ready_for_tell = False
+        self._t = 0
+        dim = log_prior.n_parameters()
+
+        # Setting the perturbation kernel
+        if perturbation_kernel is None:
+            self._perturbation_kernel = pints.MultivariateGaussianLogPrior(
```
MichaelClerx, line 78:
```suggestion

        # Setting the perturbation kernel
        if perturbation_kernel is None:
            dim = log_prior.n_parameters()
            self._perturbation_kernel = pints.MultivariateGaussianLogPrior(
```

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+
+            mw = [old_weights[j] * np.exp(self._perturbation_kernel(
+                np.subtract(new_samples[i], old_samples[j])))
+                for j in range(len(old_samples))]
+
+            w = prior_prob / sum(mw)
+            new_weights.append(w)
+        return new_weights
+
+    def set_threshold_schedule(self, schedule):
```
MichaelClerx, line 180: Would it make sense to combine these two methods? Could a schedule be more complex, e.g. 100 iterations with threshold t0, then 150 with threshold t1, etc.? In that case it might make more sense for a schedule to be something like
```
sampler.set_threshold_schedule([0, 1.23], [100, 1.123], [250, 1.0123])
```

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+        """ See :meth:`ABCSampler.ask()`. """
+        if self._ready_for_tell:
+            raise RuntimeError('ask called before tell.')
+        if self._t == 0:
+            self._xs = self._log_prior.sample(n_samples)
+        else:
+            self._xs = []
+
+            while len(self._xs) < n_samples:
+                theta_s_s = None  # to appease the linter
```
MichaelClerx, line 101:
```suggestion
                theta_s_s = None
```
The linter is right. Don't just use a variable you haven't defined yet :D

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+        return 'ABC-SMC'
+
+    def ask(self, n_samples):
+        """ See :meth:`ABCSampler.ask()`. """
+        if self._ready_for_tell:
+            raise RuntimeError('ask called before tell.')
+        if self._t == 0:
+            self._xs = self._log_prior.sample(n_samples)
+        else:
+            self._xs = []
```
MichaelClerx, line 98: Since we're doing this a lot it makes sense to preallocate and fill in _xs here instead of appending to it.

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+                        p=self._weights[self._t - 1])
+                    theta_s = self._samples[self._t - 1][indices]
+                    # perturb using K_t
+                    theta_s_s = np.add(theta_s,
+                                       self._perturbation_kernel.sample(1)[0])
+                    # check if theta_s_s is possible under the prior
+                    # sample again if not
+                self._xs.append(theta_s_s)
+        self._ready_for_tell = True
+        return self._xs
```
MichaelClerx, line 115: This allows users of the abc method to modify the values in self._xs, potentially corrupting everything. Return a numpy array instead and set it to read only (see e.g. the mcmc samplers or optimisers)

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+        self._e_schedule = [1]
+
+        # Default value for current threshold
+        self._threshold = 1
+
+        # Size of intermediate distributions
+        self._nr_samples = 100
+
+        # Set up for first iteration
+        self._samples = [[]]
```
MichaelClerx, line 68: We need to think about this. The controller holds the samples too, so we're doubling our memory usage!

Should we update the ask() method so that the user (e.g. the controller) has to provide the samples list?

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+    .. math::
+        \begin{align}
+        & \theta^* \sim p_{t-1}(\theta) \textrm{, i.e. sample parameters from
+        previous intermediate distribution} \\
+        & \theta^{**} \sim K(\theta|\theta^{*}), \textrm{i.e. perturb }
+        \theta^{*} \textrm{ to     obtain to new point } x \sim
+        p(x|\theta^{**})\textrm{, i.e. sample data from sampling
+        distribution} \\
+        & \textrm{if } s(x) < \textrm{threshold}_(t), \theta^* \textrm{
+        added to list of samples[t]}
```
MichaelClerx, line 27:
```suggestion
        & \theta^* \sim p_{t-1}(\theta) \textrm{, sample parameters from
        previous intermediate distribution} \\
        & \theta^{**} \sim K(\theta|\theta^{*}), \textrm{perturb}
        \theta^{*} \textrm{to obtain to new point} x \sim \\
        p(x|\theta^{**}), \textrm{sample data from sampling
        distribution} \\
        & \textrm{if } s(x) < \textrm{threshold}_(t), \theta^* \textrm{
        then add to list of samples[t]}
```

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+#
+import pints
+import numpy as np
+
+
+class ABCSMC(pints.ABCSampler):
+    r"""
+    Implements the ABC-SMC algorithm as describd in [1].
+
+    In each iteration of the algorithm, the following steps occur:
```
MichaelClerx, line 16: Before doing this, please explain in words how it works. This way, you can get across things like the "previous intermediate distribution" (what is that? it just appears in step 1. you haven't told the reader they should have one yet)

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Parameters
+    ----------
+    nr_samples
+        The number of samples requested for intermediate distributions.
+    error_schedule
+        The schedule of error threshold distance for all distributions.
+    """
+
+    def __init__(self, log_prior, perturbation_kernel=None):
```
MichaelClerx, line 54: Why is the kernel a constructor arg (while the schedule is not)?
Should this be changeable after construction? (like the schedule again)

MichaelClerx (changes requested):
Looked in a bit more detail now. Some stuff we need to think & talk about!

Most importantly: whos job is it to store all the samples (this class or the controller)? and how do we avoid using unnecessary memory

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,198 @@
...
+        if self._t == 0:
+            self._xs = self._log_prior.sample(n_samples)
+        else:
+            self._xs = []
+
+            while len(self._xs) < n_samples:
+                theta_s_s = None  # to appease the linter
+                while (theta_s_s is None or
+                       self._log_prior(theta_s_s) == -np.inf):
+                    indices = np.random.choice(
```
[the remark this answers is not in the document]
phumtutum: I don't think there is a way because of the `self._log_prior(theta_s_s) == -np.inf` condition which may require multiple samples.

pints/_abc/_abc_smc.py
phumtutum: I don't think this would be necessary. The idea of having intermediate sizes that are different from the final number of samples was to save time by computing the intermediate distributions faster (as fewer samples are required). For example, if we have 10 generations, requesting 1000 samples with intermediate sizes of 200, this would finish much faster than if 1000 were used for the intermediate generation sizes (the latter is the way the SMC/PMC papers would do it). This idea with using smaller intermediate sizes  was not in any paper and it was something worth trying in my dissertation.

I think going for something like `sampler.set_threshold_schedule([0, 1.23], [100, 1.123], [250, 1.0123])` would not be of much use as I cannot think of any case in which this level of control over exact distributions is helpful. If we find any use for it maybe we can add it in a different PR?

pints/tests/test_abc_smc.py
MichaelClerx: This (and a few others) are commitable suggestions, btw, you can use the commit button above (or "add to batch") to make the suggested changes directly :D

pints/_abc/_abc_smc.py
MichaelClerx: OK. Just make sure you explain the whole "intermediate distribution" thing (and what values t takes etc.) in detail in the main docstring. Also clearly indicate any changes you made from the algos in the papers we cite.

pints/_abc/_abc_smc.py
phumtutum: Indeed holding the samples in the `ABCSMC` class as well is a waste. However, after some thinking I believe that changing the ask and tell functions is not the best solution because
- It would change the ask and tell interface for all ABC algorithms (and some algorithms don't need access to the samples) as well as the ABC controller (as it would need to pass on the last generation of samples in the ask and tell calls).
- After finishing a generation of samples, the `ABCSMC` (see the `_advance_time` and `_calculate_weights` functions) also needs access to the current generation of samples, not only the last one.

I think it might be better to simply store in the `ABCSMC` class the current and last generation so that they can be used. In this case there will be some memory duplication, but if we have 10 generations, only the last 2 generations will be stored. Let me know if this seems reasonable.

MichaelClerx: Ok! And your idea of how the schedules should work means that the last 2 generates would be small, e.g. 2000 samples each instead of 10000?

MichaelClerx: I suppose if memory is an issue the controller can be told to stop logging in memory and only log on disk, so then this overhead becomes more bearable

MichaelClerx: Cool! Is there a performance benefit to pre-allocation self._samples? We know it's going to be a matrix of `2 x nr_samples x n_parameters` so could create a numpy array of the correct size and then fill that in, instead of appending to liists?

phumtutum:
I have changed some things in accordance to the comments:
- `nr_samples` and `error_schedule` can be set in the constructor
- `perturbation_kernel` can be set through its setter. This is particularly useful as there previously was no way of setting a separate perturbation kernel without adding it through features.
- "Trying t=1, threshold=1.5" is now printed for t=0 as well and it prints "Starting t=0, with threshold=3.0" instead of "trying ...".
- No longer saving all samples in `ABCSMC`, only the last 2 generations. In particular, a two-dimensional list `self._samples` now contains on `self._samples[self._t % 2]` the current set of samples and on `self._samples[(self._t - 1) % 2]` the previous set of samples. This is used for speed of execution as copying the current samples into the previous samples takes time and can bring the 41 seconds it takes (on my machine) to run the ABC SMC ipynb sampler, to ~70 seconds.
- Added ask and tell to ipynb.

@MichaelClerx please have a look when you have time and let me know if there is anything else that should be changed.

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,231 @@
...
+    nr_samples
+        The number of samples requested for intermediate distributions.
+    error_schedule
+        The schedule of error threshold distance for all distributions.
+
+    References
+    ----------
+    .. [1] "Toni, Tina, et al. Approximate Bayesian computation scheme
+            for parameter inference and model selection in dynamical systems.
+            Journal of the Royal Society Interface, 6.31: 187-202, 2009.
```
MichaelClerx, line 63:
```suggestion
    .. [1] Toni, Welch, Strelkowa, Ipsem, Stumpf. Approximate Bayesian 
           computation scheme for parameter inference and model selection in
           dynamical systems. Journal of the Royal Society Interface,
           6.31: 187-202, 2009.
```

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,231 @@
...
+            https://doi.org/10.1098/rsif.2008.0172
+
+    .. [2] "Beaumont, Mark A., et al. Adaptive approximate Bayesian
+            computation. Biometrika, 96.4: 983-990, 2009."
```
MichaelClerx, line 67:
```suggestion
    .. [2] Beaumont, Cornuet, Marin, Robert. Adaptive approximate Bayesian
           computation. Biometrika, 96.4: 983-990, 2009.
```

pints/_abc/_abc_smc.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class ABCSMC(pints.ABCSampler):
+    r"""
+    Implements the ABC-SMC algorithm as described in [1]. This
+    algorithm is also referred to as ABC Population Monte Carlo
+    (ABC PMC) [2].
+
+    In this algorithm we try to obtain ``nr_samples`` samples at
+    once. In order to transition between each set of ``nr_samples``,
```
MichaelClerx, line 19: This can easily be misunderstood as "This algorithm is just sample(nr_samples)". 

You also don't need to introduce `nr_samples` in the very first line. First tell me that there are going to be several rounds of ABC, each with a different proposal distribution (the SMC bit). Only once that's clear go into sizes of intermediate distributions etc.

MichaelClerx: You could even start with an example?

> In ABC methods, typically, a distance metric comparing the observed data and the simulated is used. Here, we use the root mean square error (RMSE), and we accept a parameter value if the for the first intermediate distribution, for the second one and for the last. 

This bit from the notebook is much clearer than the docstring!

MichaelClerx (changes requested):
Few small presentation things left

> Another alternative is to use the ask and tell interface directly. We provide an example which plots the samples for three ranges: 0-200, 200-400 and 400-1000.

Please replace by something like "We can use the ask-and-tell interface to visualise the algorithm's behaviour. In this example we run ABC-SMC with 3 successive generations (iterations 0-200, 200-400 and 400-1000) and plot each generation separately."

Then make sure the x-axes (and y-axes if you plot normalised?) of the 3 plots are the same, so that we can see the progression from generation to generation.

Then add some lines explaining what we are seeing. The 2nd generation seems to be worse than the first? Not what you'd expect based on fig 4c in the Toni paper. Am I seeing this wrong? (Then the plots need improving). Should we choose a better schedule? Or is there something else going on?

phumtutum:
I think there is a bit of a misunderstanding of how the algorithm is implemented. Intermediate generations are NOT added to the final samples, they just help lead to the creation of the distribution for the final samples. This is also what is used in the original Toni paper. If we look in Section 3.1.1 for example, 

> Next, we apply the ABC SMC approach. [...] The number of particles in each population is N=1000. To ensure the gradual transition between populations, we take T=5 populations with [...]

This implies that we have 5 generations with 1000 samples each. If we look in Figure 2, all of the histograms add up to 1000 samples. If intermediate generations samples would take part in the final samples, these histograms should add up to 5000.

image


This can be observed on histograms from Figure 3 (adding up to 100 and there are N=100 samples per generation), Figure 5 (adding up to 200 and there are N=200 samples per generation), etc.

@MichaelClerx, considering this, seeing the difference between the 0-200, the 200-400 and 400-1000 samples is not that relevant as they are part of the same generation (the final one). Differences between sections are probably just noise as the samples follow the exact same procedure. I will normalise the x-axis and clean the ipynb.

phumtutum:
I think there is a bit of a misunderstanding of how ABC SMC is implemented. The intermediate generations of the ABC SMC algorithm are not added to the final samples. This can be seen in the Toni paper. For example, in section 3.1.1:

> Next, we apply the ABC SMC approach. [...] The number of particles in each population is N=1000. To ensure the gradual transition between populations, we take T=5 populations [...]

This implies there are 1000 samples per generation, with 5 generations. However, when we observe Figure 2, we can see that the corresponding histograms add up to 1000 samples:

image

If the samples of intermediate generations were to be part of the final samples, these would add up to 5000, which is not the case. A similar issue can be seen with the histograms from Figure 3 (add up to 100 with 100 samples per generation), Figure 5 (add up to 200 with 200 samples per generation), etc.

@MichaelClerx considering this, I think there is not much point in analysing sections 0-200, 200-400 and 400-1000 as they are part of the same generation. They are derived through the same procedure and the only difference between them is probably just noise. I will change the relevant text for the ipynb. Do let me know if you think the difference between all the 3 sections is still a problem.

MichaelClerx:
@phumtutum I'm not suggesting you add all the generations together.
Figure 2 clearly shows that each successive population has a narrower distribution. (see also Fig 4c). So I'm expecting the same from our method.

> considering this, I think there is not much point in analysing sections 0-200, 200-400 and 400-1000 as they are part of the same generation. They are derived through the same procedure and the only difference between them is probably just noise.

This doesn't make sense to me. Are you saying your t is not the T from the Toni paper? If so then what is it and why are you only doing one population?

phumtutum:
@MichaelClerx so what I am saying is that, if we have 3 generations (T=3), the algorithm will never return the samples obtained by generations 1 and 2, it will only return the samples from generation 3. Therefore, printing something like Figure 4c from the Toni paper would not be possible. The 0-200, 200-400 and 400-1000 sections are all from the same generation: the last (third) one. Because of this, the 0-200, 200-400 and 400-1000 samples don't get narrower (as they are not part of different populations/generations).

So my t is the same as the one from the Toni paper, I am simply not returning the intermediate generations, I am only returning the last one. I am doing this as I thought that adding the intermediate distributions can add significant noise and the only relevant samples are the ones from the final generation. Should I return all of them? For example [abcpy](https://github.com/eth-cscs/abcpy/blob/master/abcpy/inferences.py#L770) uses a flag `full_output` (which is turned off by default) which, if turned on, returns the samples from the intermediate distributions as well.

MichaelClerx:
I agree that the final distribution is not all samples, but only the final population.

But if you're using ask and tell you can obtain a set of samples for each population, like you do in the notebook. If you plot those, and if everything is set up correctly, you should get something like Figure 2. But in the notebook example I don't see any convergence.

(Now that I'm thinking about it, we probably _do_ want to return all samples and make the user manually throw out the earlier populations. Because otherwise they can't check that the method is converging.
Also, it mirrors the MCMC case where warm-up is manually discarded after the user has confirmed convergence.)

15 Apr 2022 13:44:18 phumtutum ***@***.***>:

phumtutum:
Ok then, I will make it also return the samples of intermediate samples as well.

phumtutum:
@MichaelClerx I added access to intermediate samples as well. Here is the picture for generation 1:
generation_1

Generation 2:
output

Generation 3:
output

As you can see, the samples are getting narrower. The difference is not as big as in the Toni paper because we are using a normal distribution with mean I and covariance 0.001 * I as a perturbation kernel, which is quite narrow already, but the difference is still significant between generations.

ben18785:
Thanks both. @MichaelClerx -- yes, I think it's probably a good thing that, as you say, we allow the user to access the intermediate samples: this allows them to check all is well and, more generally, to understand how the method works.

MichaelClerx:
Thanks @phumtutum . That looks really good! Can you update the notebook to remove the ask-tell bit, and just show users how to plot the three histograms shown above to check that their method is working?

MichaelClerx (approved):
Thanks!

PR #1444, opened 2022-03-29
Issue 1443 multiple logpdfs

### LINKED ISSUE #1443: Optionally copy logpdfs before running mcmc
Currently, the MCMC controller calls `__call__` or `evaluateS1` on the same `log_pdf` object for all mcmc chains.

This causes problems for more complex LogPDF objects which have attributes which may be updated based on the parameters sent to the call method (in my case, the LogPDF object holds a grid of ODE solver time points which is updated adaptively as the chain moves through parameter space).

It's easy enough to fix by making one copy.deepcopy of the log_pdf for each chain, and adding an appropriate pints.Evaluator so each chain is just calling its own log_pdf object. @MichaelClerx let me know if this makes any sense as an optional argument to the MCMC controller !

rcw5890:
Closes #1443 
Here's my current progress on this change. It should have everything except the case of parallel evaluation.


pints/_evaluation.py
```diff
@@ -413,6 +413,43 @@ def clear(q):
...
+    """
+    def __init__(self, functions, args=None):
+        super(MultiSequentialEvaluator, self).__init__(functions[0], args)
+
+        # Check functions
+        for function in functions:
+            if not callable(function):
+                raise ValueError('The given function must be callable.')
+        self._functions = functions
+        self._num_functions = len(functions)
```
MichaelClerx, line 441: I think we were moving to using _n_thingies as the count of thingies?

pints/_evaluation.py
```diff
@@ -413,6 +413,43 @@ def clear(q):
...
+
+    def _evaluate(self, positions):
+        if len(positions) != self._num_functions:
+            raise ValueError('Number of positions does not equal number of '
+                             'functions.')
+        scores = [0] * len(positions)
```
MichaelClerx, line 447: I suppose this could even be a
```
return [f(x, *self._args) for f, x in zip(self._functions, positions)]
```
?

MichaelClerx: (I should probably have done that for the sequential one :D)

pints/_mcmc/__init__.py
```diff
@@ -301,24 +303,54 @@ def __init__(
...
-        # Check function
-        if not isinstance(log_pdf, pints.LogPDF):
-            raise ValueError('Given function must extend pints.LogPDF')
+        if isinstance(log_pdf, pints.LogPDF):
+            self._multi_logpdf = False
+
+        else:
+            self._multi_logpdf = True
+            try:
+                log_pdfs_iter = iter(log_pdf)  # noqa
```
MichaelClerx, line 312: It's very neat to check for both the iter and len parts of the sequence interface, but perhaps just len() is enough? (Then you can combine both checks: try: if len() != raise; except TypeError)

pints/_mcmc/__init__.py
```diff
@@ -301,24 +303,54 @@ def __init__(
...
+            except TypeError:
+                raise ValueError('`log_pdf` must either extend pints.LogPDF, '
```
MichaelClerx, line 314: I know it used to be a valueerror, but this should have been a typeerror

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

https://docs.python.org/3/library/exceptions.html#TypeError

pints/_mcmc/__init__.py
```diff
@@ -301,24 +303,54 @@ def __init__(
...
+                                 'or be a list of objects which extend '
+                                 'pints.LogPDF')
+
+            if len(log_pdf) != chains:
+                raise ValueError('`log_pdf` must either extend pints.LogPDF, '
```
MichaelClerx, line 319: This is a valueerror. As are the remaining ones. (The contents of the right type arg are of the wrong type = a value error, imo)

pints/_mcmc/__init__.py
```diff
@@ -274,7 +274,9 @@ class MCMCController(object):
     ----------
     log_pdf : pints.LogPDF
         A :class:`LogPDF` function that evaluates points in the parameter
-        space.
+        space, or a list of :class:`LogPDF` of the same length as `chains`. If
+        multiple LogPDFs are provided, each chain will call only its
+        corresponding LogPDF.
```
MichaelClerx, line 279: Needs a note either here or in set_parallel to say we can't have parallel execution of multiple logpdfs

pints/tests/test_mcmc_controller.py
```diff
@@ -361,6 +374,100 @@ def test_multi(self):
...
+            xs,
+            transformation=pints.LogTransformation(n_parameters),
+            sigma0=[1, 0.1, 0.01])
+        mcmc.set_parallel(True)
+        mcmc.set_max_iterations(n_iterations)
+        mcmc.set_log_to_screen(False)
+        with self.assertRaises(ValueError):
+            chains = mcmc.run()
+
+        # Test that both logpdfs are called
```
MichaelClerx, line 451: Cooooooooool

MichaelClerx (changes requested):
Looking good! Couple of small things

pints/_evaluation.py
rcw5890: Agreed, that is slightly neater

pints/_mcmc/__init__.py
rcw5890: Ok, makes sense. I've gone through the errors here and tried to be more careful with Value vs Type

pints/tests/test_mcmc_controller.py
rcw5890: Yeah, we've been trying to use some of this stuff on the Sabs projects this year so I've been learning it.

MichaelClerx (approved):
I haven't checked it in super deep detail, but it looks good to me :D

pints/_evaluation.py
```diff
@@ -413,6 +413,41 @@ def clear(q):
         return errors
 
 
+class MultiSequentialEvaluator(Evaluator):
+    """
+    Evaluates a list of functions (or callable objects) for a list of input
+    values of the same length, and returns a list containing the calculated
+    function evaluations.
```
chonlei, line 420: Although this is clear, it can be useful to give an example below and explain when we want to use this instead of the other two evaluators.

chonlei: I know this might not be used by most of the users, but even when I first looked at it, I had to look at the code to remind myself why we wanted this.

chonlei (commented):
Looks good @rcw5890, thanks! Just a comment.

rcw5890:
> Looks good @rcw5890, thanks! Just a comment.


Ooops, I just merged it! I do agree with your point though, so I'll stick a quick change in another branch

PR #1447, opened 2022-04-19
Parallelisation fix for ABC Rejection

phumtutum:
ABC Rejection is not working properly when parallelisation is enabled. In particular, if we have 8 cores for example, we try 8 sample values at once. The algorithm accepts all of them if at least one of them is below the error threshold. This is incorrect, only the ones below the error threshold should be accepted.

### DIFF
--- a/pints/_abc/_abc_rejection.py
+++ b/pints/_abc/_abc_rejection.py
@@ -70,7 +70,7 @@ class RejectionABC(pints.ABCSampler):
         if np.sum(accepted) == 0:
             return None
         else:
-            return [self._xs.tolist() for c, x in
+            return [x.tolist() for c, x in
                     enumerate(accepted) if x.all()]
 
     def threshold(self):


pints/_abc/_abc_rejection.py
```diff
@@ -70,7 +70,7 @@ def tell(self, fx):
         if np.sum(accepted) == 0:
             return None
         else:
-            return [self._xs.tolist() for c, x in
+            return [x.tolist() for c, x in
```
ben18785, line 73: What is the "c" doing here? It looks like it isn't used within the list comprehension...?

ben18785 (changes requested):
I'm sure it does what is needed but I wasn't sure what the "c" was being used for...

pints/_abc/_abc_rejection.py
phumtutum: `c` was the index, so for `accepted = [0.1, 0.2]`, we would have c=0, x=0.1 and c=1, x=0.2. I think this is not needed, I will simply removed the usage of enumerate.

phumtutum:
@ben18785 I have made it return in the same way as the ABC SMC algorithm: https://github.com/pints-team/pints/blob/master/pints/_abc/_abc_smc.py#L166 .

ben18785:
I.e. that it returns the same thing as the non-parallel version of the algorithm..?

phumtutum:
Apologies for taking a while to respond. I tested the version and it was not working properly.
`[x.tolist() for c, x in
                    enumerate(accepted) if x]`
was working properly but 
`[self._xs[c].tolist() for c, x in
                    enumerate(accepted) if x]`
 was not, even though it was working in ABC SMC. This is because, the `accepted` array was not the same as in ABC SMC. I have fixed this and tested it at https://github.com/pints-team/pints/blob/parallel-test-branch/rejection-abc.ipynb . I don't think that adding the parallel part to our ipynb would be a great idea since it is slower than the sequential one (because the degradation model simulations are fast enough that the distribution of work across cores simply adds too much time) and would not provide a good example.

pints/_abc/_abc_rejection.py
```diff
@@ -66,12 +66,12 @@ def tell(self, fx):
         self._ready_for_tell = False
 
         fx = pints.vector(fx)
-        accepted = self._xs[fx < self._threshold]
-        if np.sum(accepted) == 0:
+        accepted = fx < self._threshold
+        if np.any(accepted) == 0:
```
MichaelClerx, line 70:
```suggestion
        if not np.any(accepted):
```

phumtutum:
@ben18785 I have added the test. Sorry for taking a while, I focused on writing the thesis and revising for exams.

PR #1449, opened 2022-08-03
Test failures after updates in github actions

MichaelClerx:
- Flake8 has started complaining about `del(x)`, and rightfully so, perhaps: https://docs.python.org/3/reference/simple_stmts.html#the-del-statement
- Same for other statements (while, not, assert)
- Some slight numerical changes in tests
- language can no longer be `None` in the docs conf
- One bug?


MichaelClerx:
Now same for `assert()`

MichaelClerx:
Remaining failure:

```
======================================================================
FAIL: test_start_with_twenty (test_toy_markov_jump_model.TestMarkovJumpModel)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pints/pints/pints/tests/test_toy_markov_jump_model.py", line 36, in test_start_with_twenty
    self.assertEqual(values[-1], 0)
AssertionError: array([20.]) != 0
```

MichaelClerx:
@phumtutum do you know what's happening here?

MichaelClerx:
@phumtutum any idea? Or any of the other @pints-team/devs ?

MichaelClerx:
There seems to have been a change (a bug?) in scipy.interpolate.interp1d that was causing a bug in `MarkovJumpModel.interpolate_mol_counts` with the newer scipy version.

However, the bug was due to using a feature introduced in scipy 0.17, while pints minimum scipy requirement is 0.14. So rewrote the code not to use that feature and now it's fine.

PR #1455, opened 2022-08-14
Gave ScalingTransformation private properties and added equation to docstring

### LINKED ISSUE #1454: ScalingTransformation has visible properties, could do with better docs

MichaelClerx:
Closes #1454.

### DIFF
--- a/pints/_transformation.py
+++ b/pints/_transformation.py
@@ -835,15 +835,21 @@ class RectangularBoundariesTransformation(Transformation):
 class ScalingTransformation(Transformation):
     """
     Scaling transformation scales the input parameters by multiplying with an
-    array ``scalings`` element-wisely. And its Jacobian matrix is a diagonal
-    matrix with the values of ``1 / scalings`` on the diagonal.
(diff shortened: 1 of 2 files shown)


pints/_transformation.py
```diff
@@ -835,23 +835,29 @@ def _softplus(self, q):
 class ScalingTransformation(Transformation):
     """
     Scaling transformation scales the input parameters by multiplying with an
-    array ``scalings`` element-wisely. And its Jacobian matrix is a diagonal
-    matrix with the values of ``1 / scalings`` on the diagonal.
+    array ``scalings``.
+
+    The transformation from model to search parameters is performed as::
+
+        x_search = x_model * scalings
```
chonlei, line 842: Why not sticking to the usual `p, q` convention throughout?

chonlei (commented):
Happy with the changes except the docstring bit which I think I've used `p = x_model` and `q = x_search` throughout in the transformation docstring. I think it'd better to string to that convention.

chonlei (approved):
Thanks!

PR #1456, opened 2022-08-15
Added translation to scalingtransformation and added unitcube transformation

### LINKED ISSUE #1453: UnitCubeTransformation or ScalingTranslationTransformation
I think some methods like having parameters that live on the unit cube, so we should maybe add a UnitCubeTransformation or a more general scaling & translation transformation?

We've already got a `ScalingTransformation`, which could maybe be adapted to also have an `offset` constructor argument?

There's also a `RectangularBoundariesTransformation` that maps from rectangular boundaries onto R^n. Perhaps this name should be changed as there's other transformations you might want to do based on rectangular boundaries?

Thoughts @chonlei ?

MichaelClerx:
Closes #1453.


pints/_transformation.py
```diff
@@ -1134,3 +1147,32 @@ def sample(self, n):
...
+
+    Extends :class:`ScalingTransformation`.
+    """
+    def __init__(self, lower, upper):
+
+        # Check input
+        self._lower = pints.vector(lower)
+        self._upper = pints.vector(upper)
+        self._n_parameters = len(lower)
+        del lower, upper
```
chonlei, line 1169: Do we really have to del these? We could have del scaling and translation in `ScalingTransformation` too.

chonlei (approved):
Looks good to me, thanks Michael. Just that `del` comment see what you think, but either way is fine.

pints/_transformation.py
MichaelClerx: It's completely unnecessary. But I sometimes like to delete a variable after so that I can't accidentally use it later (e.g. mistakenly refer to `lower` (which the user can still change) instead of `self._lower`

MichaelClerx:
Thanks @chonlei !

PR #1458, opened 2022-08-16
Made gradient descent set its learning rate as min(sigma0)

MichaelClerx:
Updated its tests to ignore boundaries. Updated its docstring to make clear that it does this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project will be documented in this file.
 - [#1378](https://github.com/pints-team/pints/pull/1378) Added a class `pints.LogNormalLogLikelihood`.
 
 ### Changed
-- [#XXXX](https://github.com/pints-team/pints/pull/XXXX) The `GradientDescent` optimiser now sets its default learning rate as `min(sigma0)` (it can still be changed afterwards with `set_learning_rate()`).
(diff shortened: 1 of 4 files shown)


MichaelClerx:
Again now a new one @DavAug ! But did notice some things while copying its tests

DavAug (approved):
Thanks @MichaelClerx I am happy with the changes. I am just wondering whether it makes sense to link the learning rate to sigma0. Don't we use sigma0 to randomise the initial seeding of the optimiser which is not really related to the learning rate? So, it might be cleaner to simply separate sigma0 and the learning rate, eta, entirely and introduce eta as an input parameter to the method with a default value?

MichaelClerx:
Thanks @DavAug 
We can't really have default values other than x0 and sigma0, because the controller is responsible for creating the optimisers most of the time. The users can already set the learning rate via the controller (`.optimiser().set_learning_rate()`) & i think that's the best we can do

Regarding the link to `sigma0`. It's used by stochastic optimisers as a "standard deviation", which tells you about the "scale" of the parameters i.e. the sensitivity of the error/logpdf w.r.t. those parameters (at x0). In this (deterministic) method the learning rate is related to that scale, so it made sense to me to set a first guess based on sigma?

Issue #1105, opened 2020-04-07
Add some more simple gradient descent methods


MichaelClerx:
Part of #684 
E.g.

- [x] ~gradient descent with [momentum](https://en.wikipedia.org/wiki/Stochastic_gradient_descent#Momentum)~ #1460
- [x] ~adaptive learning rate gradient descent, e.g. [AdaGrad](https://en.wikipedia.org/wiki/Stochastic_gradient_descent#AdaGrad)~ #1468 

(Note both the links are for stochastic gradient descent, but we'd be implementing a simpler, non-stochastic one)

Won't be brilliant, but can be very informative, so largely an education thing this (or just fun)

MichaelClerx:
We have Adam & iRprop now, which meet the bill.
Also tested AdaGrad (#1468) and AdaDelta (#1470) but they didn't look promising

PR #1459, opened 2022-08-16
iRprop- local optimiser.

MichaelClerx:
See #1105

Local gradient using optimiser with adaptive step size.
Seems to work pretty well

https://doi.org/10.1016/S0925-2312(01)00700-7

The easiest to read psuedo code for this algorithm is in a figure in the paper above.
Note that the `< 0` / `> 0` is deliberate: the step size is not updated for `= 0`.


pints/_optimisers/_irpropmin.py
```diff
@@ -0,0 +1,199 @@
...
+        1. It sets the update at this iteration to zero (using
+           ``sign(df_j[i]) * step_size_j[i] = 0 * step_size_j[i]``).
+        2. It ensures that the next iteration is performed (since
+           ``df_j[i + 1] * df_j[i] = 0`` so neither if statement holds).
+
+    In this implementation, the ``step_size`` is initialised as ``sigma_0``,
+    the increase (0.5) & decrease factors (1.2) are fixed, and a minimum step
+    size of ``1e-3 * min(sigma0)`` is enforced.
+
+    This is an unbounded methods: Any ``boundaries`` will be ignored.
```
DavAug, line 45: typo: methods -> method

pints/_optimisers/_irpropmin.py
```diff
@@ -0,0 +1,199 @@
...
+        # Adapt step sizes
+        self._step_size[dprod > 0] *= self._eta_max
+        self._step_size[dprod < 0] *= self._eta_min
+
+        # Bound step sizes
+        if self._step_min is not None:
+            self._step_size = np.maximum(self._step_size, self._step_min)
+        #if self._step_max is not None:
+        #    self._step_size = np.minimum(self._step_size, self._step_max)
+
```
DavAug, line 168: Do we want to keep this, even if we introduce no upper bound?

DavAug (approved):
Looks very good to me, @MichaelClerx 
I agree with not exposing the hyperparameters for now, given that according to the literature the algorithm appears to be quite robust to changes in them.

I just have minor comments. Happy for this to be merged ☺️

MichaelClerx:
Thanks @DavAug !

PR #1460, opened 2022-08-16
Adding Adam optimiser

MichaelClerx:
See #1105 

Another adaptive local optimiser

https://doi.org/10.48550/arXiv.1412.6980

The equations aren't obvious, but you can check them against Algorithm 1 in the paper linked above.
The magic numbers in this one _could_ be hyperparameters, but I'm not sure anyone will want to tweak them...


pints/_optimisers/_adam.py
```diff
@@ -0,0 +1,163 @@
...
+        # Powers of decay rates
+        self._b1t = None
+        self._b2t = None
+
+        # Exponential decay rates for the moment estimates
+        self._b1 = 0.9    # 0 < b1 <= 1
+        self._b2 = 0.999  # 0 < b2 <= 1
+
+        # Step size
+        self._alpha = np.min(self._sigma0)
```
DavAug, line 66: Is there a reason that we initialise this method with `sigma0 = 0.1` and not with the recommended alpha = 0.001 from Algorithm 1 in the reference?

pints/_optimisers/_adam.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # "Update biased first moment estimate"
+        self._m = self._b1 * self._m + (1 - self._b1) * dfx
+
+        # "Update biased secon raw moment estimate"
+        self._v = self._b2 * self._v + (1 - self._b2) * dfx**2
+
+        # "Compute bias-corrected first moment estimate"
+        m = self._m / (1 - self._b1t)
+        self._b1t *= self._b1
```
DavAug, line 141: Can we add a comment here that we update b1 and b2 at the end as opposed to the start of the loop because of the initialisation loop in line 117? 

I think this implementation is equivalent to the pseudo code in the reference that updates t in the beginning, but it might not be immediately obvious.

pints/tests/test_opt_adam.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+#
+# Tests the API of the iRprop- optimiser.
+#
```
DavAug, line 4: Naming ADAM not iRprop-

pints/tests/test_opt_adam.py
```diff
@@ -0,0 +1,118 @@
...
+import pints.toy
+
+from shared import StreamCapture
+
+
+debug = False
+method = pints.Adam
+
+
+class TestIRPropMin(unittest.TestCase):
```
DavAug, line 22: Same here, IRPropMin -> Adam

DavAug (approved):
Thanks @MichaelClerx , I particularly like the notebook and the impact of the step size on the result. 

I just have some minor comments. I am happy for this to go in ☺️

Slightly off topic: I wonder whether the performance of Adam can be improved for our problems by adding noise to the gradient, since the algorithm was originally developed for stochastic objective functions and that might avoid getting stuck in tight locations of parameter space.

pints/_optimisers/_adam.py
MichaelClerx: It's the default value for all optimisers. But I suppose I can change it here!

pints/_optimisers/_adam.py
MichaelClerx: I wrote that in a slightly silly way. Have updated now and skipped the unneeded first iteration

MichaelClerx:
Thanks @DavAug !

> Slightly off topic: I wonder whether the performance of Adam can be improved for our problems by adding noise to the gradient, since the algorithm was originally developed for stochastic objective functions and that might avoid getting stuck in tight locations of parameter space.

Good idea. I tried it, and it does help if you get the noise just right. Not 100% sure though. Wouldn't recommend it :D

PR #1462, opened 2022-08-17
Added max evaluations stopping criterion to opt controller.

### LINKED ISSUE #937: Add `max_evaluations` stopping criteria
To OptimisationController

When comparing optimisers, or trying to write some hybrid/combined method, the concept of an "evaluation budget" is quite common, and can be easily implemented if we set a `max_evaluations` stopping criterion

MichaelClerx:
Closes #937.


pints/_optimisers/__init__.py
```diff
@@ -455,6 +455,10 @@ def __init__(
         self._unchanged_threshold = 1          # smallest significant f change
         self.set_max_unchanged_iterations()
 
+        # Maximum evaluations
+        self._max_evaluations = None
+        self.set_max_evaluations()
```
DavAug, line 460: Can remove line 459? As far as I can see, line 460 sets already the number of max evaluations to None.

DavAug (approved):
Thanks @MichaelClerx Looks really good and it's definitely super useful to be able to limit the number of evaluations. 

I am happy for this to go in 😊 I just have a minor comment.

pints/_optimisers/__init__.py
MichaelClerx, line 460:
```suggestion
```

PR #1465, opened 2022-08-25
Extended IRPropMin with boundary support and methods to set min and max step size

MichaelClerx:
⬆️  that

The boundary support is non-standard and can cause you to get stuck if you have a problematic problem (i.e. gradient next to the boundaries pointing outside). But very useful if you need to avoid evaluating outside some range that causes errors...


MichaelClerx:
@DavAug maybe don't review just yet: I think i'll need to change the strategy for step-size shrinking with general boundaries

MichaelClerx:
OK ready now @DavAug 

- Added max step size
- Added get/set for max & min step size, although default min step size still set with sigma0
- Added boundary checks: for rectangular boundaries this is straightforward. For non-rectangular it's a bit weird: I first reduce all step sizes until the boundaries are OK. Then I try setting them all back to their original values again (one by one). The idea here is that if I need to reduce step sizes in dimensions 3, 7, and 8, then that will take me forever to find out if I start by trying to reduce just 1 step size, then all combinations of 2 step sizes, then all of 3. So this looks stupid but should work

MichaelClerx:
(The strategy is biased against parameters/dimensions with a lower indice: these will get reduced first (even if reducing another parameter would also have worked). I thought a bit about randomisation but that makes this a stochastic method and I don't want to go there. The bias shouldn't matter, I hope: as long as we get to the right solution in the end)

pints/_optimisers/_irpropmin.py
```diff
@@ -127,12 +166,35 @@ def needs_sensitivities(self):
...
+        """
+        self.set_min_step_size(x[0])
+        self.set_max_step_size(x[1])
+
+    def set_max_step_size(self, step_size):
+        """
+        Sets the maximum step size (use ``None`` to let step sizes grow
+        indefinitely).
+        """
+        self._step_max = None if step_size is None else float(step_size)
```
DavAug, line 189: Can we throw an error when step_max < step_min? Does the method break when step_max = step_min?

pints/_optimisers/_irpropmin.py
```diff
@@ -127,12 +166,35 @@ def needs_sensitivities(self):
...
+
+    def set_min_step_size(self, step_size):
+        """
+        Sets the minimum step size (use ``None`` to let step sizes shrink
+        indefinitely).
+        """
+        self._step_min = None if step_size is None else float(step_size)
```
DavAug, line 196: Same here. Checking step_max > step_min is probably a good idea, and that step_min > 0

pints/_optimisers/_irpropmin.py
```diff
@@ -178,7 +238,50 @@ def tell(self, reply):
         self._current_df = dfx
 
         # Take step in direction indicated by current gradient
-        self._proposed = self._current - self._step_size * np.sign(dfx)
+        p = self._current - self._step_size * np.sign(dfx)
+
+        # Allow boundaries to reduce step size
+        if self._lower is not None:
+            # Rectangular boundaries: reduce individual step sizes until OK
+            out = np.logical_or(p < self._lower, p >= self._upper)
```
DavAug, line 246: Is there a reason for including the upper boundary and not the lower boundary?

pints/_optimisers/_irpropmin.py
```diff
@@ -89,6 +109,15 @@ def __init__(self, x0, sigma0=0.1, boundaries=None):
         # Current step sizes
         self._step_size = np.array(self._sigma0)
 
+        # Rectangular boundaries
+        self._lower = self._upper = None
+        if isinstance(self._boundaries, pints.RectangularBoundaries):
+            self._lower = self._boundaries.lower()
+            self._upper = self._boundaries.upper()
```
DavAug, line 116: I am wondering whether it's more clear to move this to line 244. The self._lower and self._upper are not used in any other methods are they? And to me, it's not immediately clear that line 244 is only being invoked for rectangular boundaries.

pints/_optimisers/_irpropmin.py
```diff
@@ -178,7 +238,50 @@ def tell(self, reply):
         self._current_df = dfx
 
         # Take step in direction indicated by current gradient
-        self._proposed = self._current - self._step_size * np.sign(dfx)
+        p = self._current - self._step_size * np.sign(dfx)
+
+        # Allow boundaries to reduce step size
+        if self._lower is not None:
```
DavAug, line 244: See comment above about making more clear that this is only used for rectangular boundaries

DavAug (commented):
Thanks @MichaelClerx this looks very good! I have not reviewed everything yet, but I thought I leave some comments already. I'll have another look tomorrow 😊

pints/_optimisers/_irpropmin.py
MichaelClerx: 😳  I didn't add it out of laziness.

Do you think we should validate late?
I wouldn't want this situation:

```
# Set min step size to 1, max to 10
opt.set_min_step_size(1)

ERROR! MAX STEP SIZE IS 0.1
```

so I was thinking the check would happen on the first call to `ask()` ?

pints/_optimisers/_irpropmin.py
MichaelClerx: Yeah, the RectangularBoundaries class uses a half-open interval: https://github.com/pints-team/pints/blob/master/pints/_boundaries.py#L91-L97

This code is doing the same check manually, so should use the same convention

pints/_optimisers/_irpropmin.py
MichaelClerx: 🤔 
I want to limit the number of function calls & inspection though, so no `isinstance` or `lower()` in `tell()`.
Not a deal-breaker though. Happy to use the isinstance check inside tell if you prefer

(I'm being efficient here out of habit, not really out of design)

MichaelClerx:
Thanks!

pints/_optimisers/_irpropmin.py
DavAug: Hmm I don't have a strong opinion on this 😊 I personally think it's nice when attempting to set a hyperparameter that breaks the method immediately raises an error, or maybe at least a UserWarning? But if the error thrown in `ask()` is clear enough about which hyperparameter breaks the optimiser, I am also happy with that 😊

pints/_optimisers/_irpropmin.py
DavAug: I am happy either way 😊 The way it is right now is probably easy enough to understand, but I would be surprised if the boundary method calls have a significant impact on the run time.

DavAug (approved):
Looks good @MichaelClerx ! I also think that the step size adjustment for non-rectangular boundaries feels a little clumsy and, as you said, there is definitely bias to move towards smaller index dimensions. But at the same time, I think this will only start to matter once we introduce more complicated boundaries than currently available in pints.

MichaelClerx:
Thanks @DavAug ! I've added the checks and merged it in

PR #1469, opened 2022-08-31
Tweaks to Adam

MichaelClerx:
Also played with boundaries, but that just caused the method to run out of momentum & get stuck so abandoned the idea.

Made a PR anyway because in the meantime I
- added pseudo code
- added two values to its logging output
- improved testing & fixed a bug in the irpropmin test


MichaelClerx:
Not at all urgent, but would still be nice if you could have a look @DavAug

MichaelClerx:
Anyone feel like reviewing this @pints-team/devs ? It's mostly just documentation improvements! 5 minute job

pints/_optimisers/_adam.py
```diff
@@ -12,16 +12,36 @@
...
 
-    This implementation uses a fixed step size, set as `` min(sigma0)``. Note
-    that the adaptivity in this method comes from the changing moments, not
-    the step size.
+    Pseudo-code is given below. Here the value of the j-th parameter at
+    iteration i is given as ``p_j[i]`` and the corresponding derivative is
+    denoted ``g_j[i]``::
+
+        b1[i] = b1[i - 1] * beta1
+        b2[i] = b2[i - 1] * beta2
```
DavAug, line 27: I am a little bit unsure how b1 and b2 are initialised from this documentation. Is b1[i] = beta1**i, so b1[0]=1? Or is something more fancy happening?

DavAug: Ok yes, from looking at the code, I can see that is what is happening. 

Could we maybe simplify the documentation then and just write b1[i] = beta^i, which avoids defining initial conditions separately?

DavAug (approved):
Thanks @MichaelClerx . I think this looks great! I just have a minor comment on the docs

pints/_optimisers/_adam.py
MichaelClerx: Thanks @DavAug . Yes the pseudo code was a bit confusing and wrong (or maybe better to say that I was confused and wrong). Have updated it now (but with beta^(1+i) to account for my i=0 equalling the mathsy t=1)

PR #1474, opened 2022-11-08
Make "contribution" section a bit less daunting

MichaelClerx:
Does this make it sound less off-putting?


MichaelClerx:
Could add in some note to say if you're "watching" you might want to check "participants and mentions" instead of "all activity" ?

ben18785 (approved):
Thanks @MichaelClerx -- I like this idea.

PR #1484, opened 2023-07-25
I1483 gaussian integrated log uniform log likelihood

### LINKED ISSUE #1483: Create GaussianIntegratedLogUniformLogLikelihood
Similar to https://pints.readthedocs.io/en/stable/log_likelihoods.html#pints.GaussianIntegratedUniformLogLikelihood, but assume a prior that is uniform in log-transformed space instead of the original parameter space, i.e. `p(\sigma) \propto 1 / \sigma`, following Eq. (7)-(9) in https://doi.org/10.1093/gji/ggaa168.

chonlei:
Closes #1483

### DIFF
--- a/pints/_log_likelihoods.py
+++ b/pints/_log_likelihoods.py
@@ -551,11 +551,24 @@ class GaussianIntegratedUniformLogLikelihood(pints.ProblemLogLikelihood):
     Calculates a log-likelihood assuming independent Gaussian-distributed noise
     at each time point where :math:`\sigma\sim U(a,b)` has been integrated out
     of the joint posterior of :math:`p(\theta,\sigma|X)` in the same way as in
-    :class:`pints.GaussianIntegratedLogUniformLogLikelihood`.
+    :class:`pints.GaussianIntegratedLogUniformLogLikelihood`,
+
+    .. math::
+        \begin{align} p(\theta|X) &= \int_{0}^{\infty} p(\theta, \sigma|X)
+        \mathrm{d}\sigma\\
+        &\propto \int_{0}^{\infty} p(X|\theta, \sigma) p(\theta, \sigma)
+        \mathrm{d}\sigma,\end{align}
 
     Note that this is exactly the same statistical model as
     :class:`pints.GaussianLogLikelihood` with a uniform prior on
     :math:`\sigma`.
+
+    Similar to :class:`pints.GaussianIntegratedLogUniformLogLikelihood`, a
+    possible advantage of this log-likelihood compared with using a
+    :class:`pints.GaussianLogLikelihood`, is that it has one fewer parameters
+    (:math:`sigma`) which may speed up convergence to the posterior
+    distribution, especially for multi-output problems which will have
+    ``n_outputs`` fewer parameter dimensions.
 
     The log-likelihood is given in terms of the sum of squared errors:
 
--- a/pints/_log_likelihoods.py
+++ b/pints/_log_likelihoods.py
@@ -532,9 +532,9 @@ class GaussianIntegratedLogUniformLogLikelihood(pints.ProblemLogLikelihood):
 
         # Pre-calculate
         n = self._nt
-        self._n_over_2 = n / 2.0
+        self._n_over_2 = n / 2
         self._log_gamma = scipy.special.gammaln(self._n_over_2)
-        self._constant_1 = - (n - 2.0) / 2.0 * np.log(2.0)
+        self._constant_1 = -(n - 2) / 2 * np.log(2)
         self._constant = self._constant_1 + self._log_gamma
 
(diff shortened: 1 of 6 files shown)


pints/_log_likelihoods.py
```diff
@@ -465,6 +465,90 @@ def evaluateS1(self, parameters):
         return L, dL
 
 
+class GaussianIntegratedLogUniformLogLikelihood(pints.ProblemLogLikelihood):
```
MichaelClerx, line 468: Add a note (quite early on) saying how this is different from the GaussianIntegratedUniformLogLikelihood

pints/_log_likelihoods.py
```diff
@@ -465,6 +465,90 @@ def evaluateS1(self, parameters):
...
+    Calculates a log-likelihood assuming independent Gaussian-distributed noise
+    at each time point where :math:`p(\sigma)\propto 1/\sigma` has been
+    integrated out of the joint posterior of :math:`p(\theta,\sigma|X)`,
+
+    .. math::
+        \begin{align} p(\theta|X) &= \int_{0}^{\infty} p(\theta, \sigma|X)
+        \mathrm{d}\sigma\\
+        &\propto \int_{0}^{\infty} p(X|\theta, \sigma) p(\theta, \sigma)
+        \mathrm{d}\sigma,\end{align}
+
```
MichaelClerx, line 479: This is exact same as next class

pints/_log_likelihoods.py
```diff
@@ -465,6 +465,90 @@ def evaluateS1(self, parameters):
...
+    :class:`pints.GaussianLogLikelihood` with a uniform prior on
+    :math:`\sigma` in the logarithmic transformed space.
+    This is also known as Jeffrey's prior, as the standard deviation is a scale
+    parameter, as discussed here: https://doi.org/10.1093/gji/ggaa168.
+
+    A possible advantage of this log-likelihood compared with using a
+    :class:`pints.GaussianLogLikelihood`, is that it has one fewer parameters
+    (:math:`sigma`) which may speed up convergence to the posterior
+    distribution, especially for multi-output problems which will have
+    ``n_outputs`` fewer parameter dimensions.
```
MichaelClerx, line 490: This is all very similar to the other class. Please can you rewrite so that the simplest class has this text, and then the next level up explains the differences with the previous class?

pints/_log_likelihoods.py
```diff
@@ -465,6 +465,90 @@ def evaluateS1(self, parameters):
...
+        self._no = problem.n_outputs()
+
+        # Add parameters to problem
+        self._n_parameters = problem.n_parameters()
+
+        # Pre-calculate
+        n = self._nt
+        self._n_over_2 = n / 2.0
+        self._log_gamma = scipy.special.gammaln(self._n_over_2)
+        self._constant_1 = - (n - 2.0) / 2.0 * np.log(2.0)
```
MichaelClerx, line 537:
```suggestion
        self._n_over_2 = n / 2
        self._log_gamma = scipy.special.gammaln(self._n_over_2)
        self._constant_1 = -(n - 2) / 2 * np.log(2)
```

MichaelClerx: On the last line, do you want to put the np.log(2) at the start maybe? Because the way it's written now makes my typo alarm bells go off 😅

pints/_log_likelihoods.py
```diff
@@ -465,6 +465,90 @@ def evaluateS1(self, parameters):
...
+    def __call__(self, x):
+        error = self._values - self._problem.evaluate(x)
+        sse = np.sum(error**2, axis=0)
+
+        # Calculate
+        sse = pints.vector(sse)
+        return np.sum(
+            self._constant -
+            self._n_over_2 * np.log(sse)
+        )
```
MichaelClerx, line 549:
```suggestion
        return np.sum(self._constant - self._n_over_2 * np.log(sse))
```

MichaelClerx (approved):
Some small suggestions and a request to rewrite the docstring. But overal looks good.

pints/_log_likelihoods.py
MichaelClerx: OR, since the code for this class seems to be simpler, do it the other way around

pints/_log_likelihoods.py
MichaelClerx: Well, just do something that makes it clear early on that there are 2 very similar classes, and that this isn't a mistake :D

pints/_log_likelihoods.py
MichaelClerx: (or not. see above)

pints/_log_likelihoods.py
chonlei: Done.

Issue #1487, opened 2023-07-25
Docs no longer build


MichaelClerx:
Seems to be failing because threadpoolctl version used does not support Python 3.7. Bit weird as our own 3.7 tests work fine.
Can change python version, but not with current set up.

Will need to add a new settings file before september anyway:
  https://blog.readthedocs.com/migrate-configuration-v2/

PR #1488, opened 2023-07-25
Added .readthedocs.yaml

MichaelClerx:
Addresses #1487 

~Not sure how to test~


MichaelClerx:
Works if we force Python 3.9, but not 3.11 because then the particular sphinx that readthedocs installs doesn't work 🤷🏻‍♂️ 

https://readthedocs.org/projects/pints/builds/21399780/

chonlei (approved):
Thanks, Michael!

PR #1495, opened 2023-09-25
Updated matplotlib.patches. Ellipse syntax to match 3.x version (but still backwards compatible)

MichaelClerx:
Fixes notebooks test.


DavAug (approved):
Thank you @MichaelClerx Very cool notebook! I actually haven't had a look at it before

PR #1500, opened 2023-10-19
Added censored log-likelihood and some tests for it

k-shep:


k-shep:
@DavAug and @rccreswell 

I’ve had a go at the censored log-likelihood, which we discussed last year. I appreciate that there are quite a lot of tests, so I thought I’d spell out the cases I’m testing (for a variety of objects):
1.	none of the data are censored;
2.	some data are lower/upper censored and some aren’t censored;
3.	some of the data are lower censored, some are upper censored and some are not censored;
4.	all of the data are lower/upper censored.

I’m testing the likelihood and S1 are correct for these cases and the following object types (for the values):
1.	list;
2.	one dimensional array;
3.	two dimensional array single (i.e. a 1D array in a 2D array format);
4.	two dimensional array multi (i.e. a ‘proper’ 2D array).

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,323 @@ def evaluateS1(self, x):
...
+                if len(b) != self._no:
+                    raise ValueError(
+                        'Upper limit must be a ' +
+                        ' scalar or a vector of length n_outputs.')
+
+        if (a is not None) and (b is not None):
+            diff = b - a
+            if np.any(diff <= 0):
+                raise ValueError('Upper limit ' +
+                                 'must exceed lower limit.')
```
DavAug, line 984: With the change above, you could remove the none checks

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,323 @@ def evaluateS1(self, x):
...
+        # Define the condition for whether a point is not censored
+        if (self._a is not None) and (self._b is not None):
+            self._condition = (self._a < self._values
+                               ) & (self._values < self._b)
+        elif (self._a is not None) and (self._b is None):
+            self._condition = self._a < self._values
+        elif (self._a is None) and (self._b is not None):
+            self._condition = self._values < self._b
+        else:
+            raise ValueError("Not a censored likelihood")
```
DavAug, line 997: With the change above you can simply this logic and don't actually need to raise the ValueError anymore:

self.condition = (self._values >= self._a) & (self._values <= self._b)

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,323 @@ def evaluateS1(self, x):
...
+            for i in range(np.shape(self._values)[1]):
+                for j in range(np.shape(self._values)[0]):
+                    if a is not None:
+                        if self._values[j, i] < self._a[i]:
+                            raise ValueError(
+                                "A value is below the lower limit.")
+                    if b is not None:
+                        if self._values[j, i] > self._b[i]:
+                            raise ValueError(
+                                "A value is above the upper limit.")
```
DavAug, line 1019: Content comment:

Perhaps, I misunderstood the use of the likelihood. But isn't the purpose of the sensored likelihood that measurement values below the threshold exist but the measurement values can not be trusted because the measuring device has reached its resolution limit or equivalent? 

I think you could do without this check.

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,323 @@ def evaluateS1(self, x):
...
+
+        # Print data and values that are censored
+        if verbose:
+            print("The data are {}. \n The lower censored values"
+                  " are {}. \n The upper censored values"
+                  " are {}.".format(self._values,
+                                    np.extract(self._values == self._a,
+                                               self._values),
+                                    np.extract(self._values == self._b,
+                                               self._values)))
```
DavAug, line 1029: Cosmetic comment:

In my mind, this seems like something that should not be in the __init__ but, perhaps, in a separate method in case you are interested in knowing how the data was sensored. But given that you define the thresholds, it also appears more like a debugging functionality (perhaps, even more reason to move it into a method or remove it completely).

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,323 @@ def evaluateS1(self, x):
...
+
+        # Number of points that aren't censored (for each observation
+        # in the multioutput case)
+        self._n_not_censored = np.sum(self._condition, axis=0)
+
+        # Pre-calculate parts
+        self._logn = 0.5 * self._n_not_censored * np.log(2 * np.pi)
```
DavAug, line 1036: Accuracy comment:

This is quite cheap to calculate and you actually increase the floating point error by pre-calculating it. So, perhaps, you could just move that back into the __call__ and evaluate method.

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,323 @@ def evaluateS1(self, x):
...
+            scipy.stats.norm.cdf(x=self._values,
+                                 loc=self._problem.evaluate(x[:-self._no]),
+                                 scale=sigma)),
+            where=self._values == self._a)
+        upper_censored_sum = np.sum(
+            np.log(1 - scipy.stats.norm.
+                   cdf(x=self._values,
+                       loc=self._problem.evaluate(x[:-self._no]),
+                       scale=sigma)),
+            where=self._values == self._b)
```
DavAug, line 1058: Same here as for the lower threshold

DavAug (commented):
Hi @k-shep ,

Thank you for the great work! The likelihood looks very good 😊

I have some minor comments related to cosmetics, performance, and feedback as well as one more fundamental comment on the justification for the likelihood calculation. I have to admit, I am not super close to this topic atm, but it appears to me that sensored values may get higher likelihood contributions than neighbouring unsensored values. I would just be curious to learn whether this is what we want and why that is a good thing 😊

I hope you have a nice evening and best wishes,
David

DavAug:
Ok, just to follow up on my comment before: I had a quick look at the PK modelling reference and your implementation seems to be consistent with method M3 (except, perhaps, a factor of 2, but I am assuming this comes from the difference in the Phi and CDF definitions). 

I am still not quite convinced that this is the best method to handle censored data, because it will effectively drive the model to go as far into the censored region as possible to maximise the likelihood. So, for a model with sufficient flexibility, you might get strange results where the model shoots off to minus infinity whenever there is a censored measurement below the censoring threshold (or to infinity for censored measurements above the censoring threshold). That doesn't really seem to be what we want. 

Their M4 suggestion would do a little bit better by bounding the range of censored measurements from below for the lower censored measurements (and above for the upper censored measurements) but then you can already anticipate that the Gaussian likelihood will just be maximised when the model positions itself right in the middle between the lower bound and the censoring threshold whenever a censored measurement is present. 

I think these shortcomings are very exciting and leave quite a bit of room for some methodological work 😊

DavAug:
Also one more methodological comment: 

Given that a censored measurement contributes with the CDF to the likelihood, as opposed to uncensored measurements contributing with the PDF, the likelihood will be larger the more censored measurements are present (by construction because the CDF is the integral over the PDF). 

For the maximum likelihood approach in the PK modelling reference this matters maybe less (although I still don't think it's ideal), for Bayesian inference this will have the consequence that posterior distributions will be ich narrower than they should. This feels very counterintuitive, as posteriors should probably be wider for uncensored measurements. Just something to think about. 😊

k-shep:
Thanks @DavAug  for the review!

I think I’ve made all the changes you suggested, and the code is much simpler now.

Just as a general comment, I had previously tried to copy the code or coding style from elsewhere in the file, so I think some of your suggestions might also be relevant for that code.

pints/_log_likelihoods.py
k-shep: I’ve removed this check. I’ve seen values at the limit of detection recorded as equal to it, but I think you’re correct that in general this may not be the case.

k-shep:
> Ok, just to follow up on my comment before: I had a quick look at the PK modelling reference and your implementation seems to be consistent with method M3 (except, perhaps, a factor of 2, but I am assuming this comes from the difference in the Phi and CDF definitions).
> 
> I am still not quite convinced that this is the best method to handle censored data, because it will effectively drive the model to go as far into the censored region as possible to maximise the likelihood. So, for a model with sufficient flexibility, you might get strange results where the model shoots off to minus infinity whenever there is a censored measurement below the censoring threshold (or to infinity for censored measurements above the censoring threshold). That doesn't really seem to be what we want.
> 
> Their M4 suggestion would do a little bit better by bounding the range of censored measurements from below for the lower censored measurements (and above for the upper censored measurements) but then you can already anticipate that the Gaussian likelihood will just be maximised when the model positions itself right in the middle between the lower bound and the censoring threshold whenever a censored measurement is present.
> 
> I think these shortcomings are very exciting and leave quite a bit of room for some methodological work 😊

Yes – my industrial supervisor recommended using the M3 method, so this is why I started implementing this censored log-likelihood. I think in the reference, they’re multiplying the whole likelihood (both censored and uncensored terms) by -2, although I’m not sure why.

Thank you for your thoughts on this – I think it is potentially an interesting thing to investigate and, if I have time, I’m hoping I might be able to do this later in my PhD. In the papers that I came across when I was looking at this before, I think people dealt with the censored data in different ways – I think I came across the use of the M3 method and also removing the censored data, and I believe when I tried to compare these for simulated data they gave different results in terms of parameter inference (via finding the MLE). There didn’t seem to be a consensus for the best way to deal with censored data, at least in the field I was looking at – I haven’t, however, looked too much into the general theory of this.

From looking at the experiments I tried before, it looks like your thoughts on censoring ‘narrowing’ the likelihood (and posterior) are also correct. The likelihood surface does seem to look flatter when values are not censored, compared to when they are.

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,302 @@ def evaluateS1(self, x):
...
+
+    Parameters
+    ----------
+    problem
+        A :class:`SingleOutputProblem` or :class:`MultiOutputProblem`.
+    lower
+        The lower limit for censoring.
+    upper
+        The upper limit for censoring.
+    verbose
```
DavAug, line 939: Can you remove the verbose input? I don't think it's used anymore.

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,302 @@ def evaluateS1(self, x):
...
+        # Convert the limit to an object of the correct type
+        if limit is not None:
+            if np.isscalar(limit):
+                limit = np.ones(self._no) * float(limit)
+            else:
+                limit = pints.vector(limit)
+
+        return limit
+
+    def print_censored_values(self):
```
DavAug, line 1008: Cosmetic comment:

Methods are typically ordered alphabetically with underscores coming first. So this method should come after the __call__ method.

DavAug: And the evaluateS1 method.

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,302 @@ def evaluateS1(self, x):
...
+        # the most expensive step in inference, especially for ODE models
+        output = self._problem.evaluate(theta)
+
+        squared_error = np.sum((self._values - output)**2,
+                               axis=0, where=self._condition)
+
+        # Calculate part of the likelihood corresponding to the censored data
+        lower_censored_sum = np.sum(np.log(
+            scipy.stats.norm.cdf(x=self._a, loc=output, scale=sigma)),
+            where=self._values <= self._a)
```
DavAug, line 1035: Cosmetic / performance comment:

I just realised that this mask also does never change, once initiated. Is there a reason why you only predefined self._condition but not the masks for lower and upper? I would suggest to define all masks in the same place: either everything in the __init__; or everything in the __call__. 

For large data, this may also be a cost worth front loading, which would, perhaps, favour the __init__ option.

pints/_log_likelihoods.py
```diff
@@ -861,6 +862,302 @@ def evaluateS1(self, x):
...
+        upper_censored_sum = np.sum(
+            np.log(1 - scipy.stats.norm.
+                   cdf(x=self._b, loc=output, scale=sigma)),
+            where=self._values >= self._b)
```
DavAug, line 1039: Same here.

DavAug (approved):
Hi @k-shep 

Really good job, also with the extensive testing! 

I have some minor additional cosmetic comments after a second pass, but I don't think there needs to be another review. If you can have a look at some of those, perhaps especially the ones in the doc string, I am happy for this to go in 😊

Thank you for the hard work!

k-shep:
Thanks @DavAug for reviewing this again. I’ve made the changes you’ve suggested. I have changed the documentation, although I’ve written it differently to how you suggested. Would you be OK just checking that you agree what I’ve written makes sense now?

DavAug:
> Thanks @DavAug for reviewing this again. I’ve made the changes you’ve suggested. I have changed the documentation, although I’ve written it differently to how you suggested. Would you be OK just checking that you agree what I’ve written makes sense now?

Hi @k-shep 

Of course, no problem! Yes the docstring makes sense and reads very well! Good job 😊

DavAug:
> Thanks @DavAug :) - should I now just click the merge button?

Yes 😊

k-shep:
Hi @DavAug,  I just noticed that I didn't put this log-likelihood class in the correct position for it to be in alphabetical order in that file. Is the best way for me to correct this by opening a new issue and pull request, and making this change?

DavAug:
Hi,

Good catch! Yes, create an issue for it and a PR when you have changed the order 😊

Thank you! 

Best,
David

k-shep:
Thanks @DavAug - I've created the PR for this here: https://github.com/pints-team/pints/pull/1526

Issue #682, opened 2019-02-09
Decide what to do with discrete parameters in PINTS


ben18785:
To begin with we should probably state that PINTS does not allow discrete parameters.

Later on, we should probably implement some discrete samplers. Not difficult to do so, but will require a bit a bit of fiddling. Should encourage users to marginalise out their discrete variables when possible.

ben18785:
List of potential discrete samplers:

- [Continuous relaxation HMC](Additionally, any way we could facilitate marginalising out of discrete parameters would allow users to estimate their models in PINTS as they stand).

MichaelClerx:
Just decided in meeting that current solution is:

- [ ] To begin with we should probably state that PINTS does not allow discrete parameters.

ben18785:
Yep, in the README -- I can't think where it would go in the docs (other than in lots of places!)

PR #1506, opened 2023-11-07
Add note to readme on discrete parameters

MichaelClerx:
See #682 

Could also go in definition of error and likelihood ~?~ !


MichaelClerx:
@ben18785 what do you think? LogPDF and ErrorMeasure are maybe not the first place you'd look, but at least they sit nicely atop the hierarchy of all things-with-parameters

I-Bouros (approved):
I think this is ideal for this moment.

PR #1508, opened 2023-11-21
Allowing xtol and ftol stopping criteria in OptimisationController

### LINKED ISSUE #1502: Add "xtol" and "ftol

MichaelClerx:
Closes #1502 

- [x] ftol is already possible, using `set_max_unchanged_iterations(1, ftol)`
- [x] xtol can now similarly be implemented with `set_max_unmoved_iterations(1, xtol)`


MichaelClerx:
@martinjrobins @chonlei @mirams can you have a look and see if this makes sense, before I add tests?

CHANGELOG.md
```diff
@@ -5,8 +5,9 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- [#1508](https://github.com/pints-team/pints/pull/1508) Added a method `OptimisationController.set_max_unmoved_iterations` that allows methods to stop after 1 or more iterations with no significant movement in parameter space.
```
mirams, line 8: "unmoved/unchanged" is a bit vague and not immediately obvious to me in terms of whether it refers to xtol or ftol? `unmoved parameters`, `unchanged objective`?

MichaelClerx: I agree! But can't think of anything better that doesn't involve deprecating `set_max_unchanged_iterations` which is probably one of the most used methods of the controller class

mirams: Hmmmm, given all the different names that exist already, inventing another set that are pints specific doesn't seem great anyway - and even more so if they are vague!

PR #1523, opened 2024-03-25
Rt Notebook example - corrected

I-Bouros:


MichaelClerx:
Can you give a bit of background about this example? What's the reason for including it (from a PINTS point of view: which features does it highlight)?

In its current form I don't think it belong in the "toy" section of examples, as those all showcase a toy model included in the PINTS module.

If you did want to make this into a PINTS toy problem, we would have to think about how (or whether) to include the CSV data, as we would have to then package this data as part of the `pints` module

I-Bouros:
> Can you give a bit of background about this example? What's the reason for including it (from a PINTS point of view: which features does it highlight)?
> 
> In its current form I don't think it belong in the "toy" section of examples, as those all showcase a toy model included in the PINTS module.
> 
> If you did want to make this into a PINTS toy problem, we would have to think about how (or whether) to include the CSV data, as we would have to then package this data as part of the `pints` module

It's meant to be an additional example for the paper, and to showcase how someone can try and define their own logPDF objects rather than resorting to singleOutput problems. I did struggle to find a good place to include it, to be honest.

MichaelClerx:
I suppose it could always go under "Miscellaneous", or it could go in a separate repo? (`pints-team/rt-estimation-renewal-equation`?), or in a repo with things related to the paper?

MichaelClerx (approved):
Looks good to me!

MichaelClerx:
@ben18785 want to comment on the science of this? Or shall we just merge it in?

MichaelClerx:
OK, in it goes!

Issue #1530, opened 2024-06-17
Numpy 2 compatibility


MichaelClerx:
Major numpy version is now available through pip, will likely require some changes

MichaelClerx:
https://numpy.org/devdocs/release/2.0.0-notes.html

MichaelClerx:
```
======================================================================
ERROR: test_bounded (test_opt_cmaes.TestCMAES.test_bounded)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/michael/dev/pints/pints/tests/test_opt_cmaes.py", line 55, in test_bounded
    found_parameters, found_solution = opt.run()
                                       ^^^^^^^^^
  File "/home/michael/dev/pints/pints/_optimisers/__init__.py", line 648, in run
    xs = self._optimiser.ask()
         ^^^^^^^^^^^^^^^^^^^^^
  File "/home/michael/dev/pints/pints/_optimisers/_cmaes.py", line 52, in ask
    self._initialise()
  File "/home/michael/dev/pints/pints/_optimisers/_cmaes.py", line 135, in _initialise
    self._es = cma.CMAEvolutionStrategy(self._x0, self._sigma0, options)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/michael/dev/myokit/venv/lib/python3.12/site-packages/cma/evolution_strategy.py", line 1503, in __init__
    self.boundary_handler = self.boundary_handler(opts['bounds'])
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/michael/dev/myokit/venv/lib/python3.12/site-packages/cma/constraints_handler.py", line 274, in __init__
    self.bounds_tf = BoxConstraintsLinQuadTransformation(self.to_dim_times_two(bounds))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/michael/dev/myokit/venv/lib/python3.12/site-packages/cma/transformations.py", line 263, in __init__
    super(BoxConstraintsLinQuadTransformation, self).__init__(bounds)
  File "/home/michael/dev/myokit/venv/lib/python3.12/site-packages/cma/transformations.py", line 124, in __init__
    self.initialize()
[...]

PR #1531, opened 2024-06-17
Updates for numpy 2 compatibility

MichaelClerx:
1. Not many changes in PINTS luckily, just replacing a `np.array(x, copy=False)` with `np.asarray(x)`
2. Remaining errors are in `pycma`. These are already fixed in development but we'll need to wait for a new release.

### DIFF
--- a/pints/_optimisers/_nelder_mead.py
+++ b/pints/_optimisers/_nelder_mead.py
@@ -186,7 +186,7 @@ class NelderMead(pints.Optimiser):
                     self._xs[1 + i][i] *= x_grow
 
             # Ask for initial points
-            return np.array(self._xs, copy=True)
+            return np.copy(self._xs)
 
         # Shrink operation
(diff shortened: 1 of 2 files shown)


MichaelClerx:
Fixes #1530

MichaelClerx:
The cma issue is also what's breaking the notebook tests right now

nikohansen:
I just released the latest `pycma` which is compatible with numpy 2.0.

MichaelClerx:
Thanks @nikohansen !

MichaelClerx:
In the meantime, a scipy update remove `cumtrapz` which had been deprecated, so fixed that too. It means the test that used this function now requires scipy 1.12, released Jan 2024

chonlei (commented):
Thanks, Michael!

chonlei (approved):

PR #1693, opened 2024-12-18
Added a note about sigma0 to documentation.

RomanSyunyaev:
Added a note to the documentation which will potentially save  a lot of hedache for users.

I am not quite sure about the sphere -- it looks like the CMA-ES code is doing something more complicated, and takes into account the boundaries as well.  But apparently CMA-ES object is initialised with a single value for the variance (line 128 in the _cmaes.py).

### DIFF
--- a/pints/_optimisers/_cmaes.py
+++ b/pints/_optimisers/_cmaes.py
@@ -18,6 +18,16 @@ class CMAES(pints.PopulationBasedOptimiser):
     CMA-ES stands for Covariance Matrix Adaptation Evolution Strategy, and is
     designed for non-linear derivative-free optimization problems.
 
+    CMAES object is initialised with x0 (initial position), sigma0 (list of
+    initial variances), and boundaries (list of boundaries). For the
+    compatibility with the other optimization methods sigma0 can be specified
+    either as a scalar value (one standard deviation for all coordinates)
+    or as an array with one entry per dimension. However inital covariance
+    matrix corresponds n-dimensional sphere with the min(sigma0) radius. Note,
+    that if sigma0 is not provided it is initialised with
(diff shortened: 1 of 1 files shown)


pints/_optimisers/_cmaes.py
```diff
@@ -18,6 +18,16 @@ class CMAES(pints.PopulationBasedOptimiser):
...
+    CMAES object is initialised with x0 (initial position), sigma0 (list of
+    initial variances), and boundaries (list of boundaries). For the
+    compatibility with the other optimization methods sigma0 can be specified
+    either as a scalar value (one standard deviation for all coordinates)
+    or as an array with one entry per dimension. However inital covariance
+    matrix corresponds n-dimensional sphere with the min(sigma0) radius. Note,
+    that if sigma0 is not provided it is initialised with
+    (1 / 6) * boundaries.range(), that means that e.g. narrow boundaries will
+    result in a slow convergence.
+
```
MichaelClerx, line 30:
```suggestion
    CMA-ES is initialised with an initial position ``x0``, an initial
    covariance ``sigma0``, and optionally a set of :class:`pints.Boundaries`.

    The initial covariance ``sigma0`` should be a scalar (one standard
    deviation for all parameters), but for compatibility with other optimisers
    an array is also accepted: in this case the smallest value in ``sigma0``
    will be used. If no ``sigma0`` is given a guess will be made based on the
    boundaries (if provided) or ``x0``, see :meth:`Optimiser` for details. The
    method's authors suggest choosing ``sigma0`` such that the optimum is
    expected to be within ``3 * sigma0`` from the initial position ``x0``.
```

MichaelClerx:
Thanks @RomanSyunyaev . Picking this up in separate PR

Issue #1691, opened 2024-12-09
`SingleChainMCMC` class errors with `set_log_pdf_storage(True)`


MarcBerliner:
Hello,

I am using pints for MCMC, and I ran into an issue storing the log PDFs. This error seems to only occur with child classes of the `SingleChainMCMC` class, such as `NoUTurnMCMC`.

MWE:
```python
import pints
import pints.toy

# Create log pdf
log_pdf = pints.toy.GaussianLogPDF([5, 5], [[4, 1], [1, 3]])

# Create mcmc
x0 = [[2, 2]]
method = pints.NoUTurnMCMC

opt = pints.MCMCController(
    log_pdf,
    chains=1,
    x0=x0,
    method=method,
)

# opt.set_log_pdf_storage(False) # succeeds
opt.set_log_pdf_storage(True) # fails

opt.run()
```
which produces the stacktrace:
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
TypeError: float() argument must be a string or a real number, not 'tuple'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
Cell In[14], [line 20](vscode-notebook-cell:?execution_count=14&line=20)
     [11](vscode-notebook-cell:?execution_count=14&line=11) opt = pints.MCMCController(
     [12](vscode-notebook-cell:?execution_count=14&line=12)     log_pdf,
     [13](vscode-notebook-cell:?execution_count=14&line=13)     chains=1,
     [14](vscode-notebook-cell:?execution_count=14&line=14)     x0=x0,
     [15](vscode-notebook-cell:?execution_count=14&line=15)     method=method,
     [16](vscode-notebook-cell:?execution_count=14&line=16) )
     [18](vscode-notebook-cell:?execution_count=14&line=18) opt.set_log_pdf_storage(True)
---> [20](vscode-notebook-cell:?execution_count=14&line=20) opt.run()

File ~/Documents/GitHub/pints/pints/_mcmc/__init__.py:757, in MCMCController.run(self)
    [754](https://file+.vscode-resource.vscode-cdn.net/Users/marcberliner/Documents/GitHub/pints/examples/sampling/~/Documents/GitHub/pints/pints/_mcmc/__init__.py:754) if store_evaluations:
[...]

PR #1700, opened 2024-12-18
LogPDF storage issue with multi-chain methods

MichaelClerx:
Tests and fix for #1691 

Touched a few more files than necessary, because I updated some docstrings while looking for a multi-chain method using sensitivities. There wasn't one, so I added a "fake" one for testing.


MichaelClerx:
Very easy fix @martinjrobins !
Testing a bit more laborious but still mostly copy pasted from other tests so shouldn't be hard to review

pints/tests/test_mcmc_controller.py
```diff
@@ -706,7 +726,9 @@ def test_log_pdf_storage_in_memory_multi(self):
         x0 = np.array(self.real_parameters) * 1.05
         x1 = np.array(self.real_parameters) * 1.15
         x2 = np.array(self.real_parameters) * 0.95
-        xs = [x0, x1, x2]
+        x3 = np.array(self.real_parameters) * 0.95
```
MichaelClerx, line 729: This bit just removes a warning because diff ev wants you to use many chains

martinjrobins (approved):
looks great, thanks @MichaelClerx :)

PR #1742, opened 2026-03-06
Expand platforms that unit tests run on

### LINKED ISSUE #1741: Rationalise unit test workflows
I don't like that we have two workflows for unit tests:

1. https://github.com/pints-team/pints/blob/main/.github/workflows/unit-test-os-coverage.yml
2. https://github.com/pints-team/pints/blob/main/.github/workflows/unit-test-python-coverage.yml

(I do realise that this was originally my design choice!)

Let's rationalise into a single workflow, ensuring we have coverage on all modern platforms and python versions, including expanding our testing to arm64 architectures.

fcooper8472:
Previously, unit tests have only been run on x86_64 platforms, but these have included all current (and two old) python CPython version.

I propose testing:

- All current python versions, each on two different operating systems
- The most recent deceased python version
- The next pre-release version (to check future compatibility)

Along with each of Ubuntu, macOS and Windows, both x86_64 and arm64 architectures, just to check there are no nasty surprises for users.

The current syntax I'm proposing is a little ugly. But it's not justifiable to do the full cartesian product of all pythons and operating systems.


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+            extras: ""
+          - py: "3.12"
+            os: ubuntu-24.04
+            extras: ".[stan]"
+          - py: "3.12"
+            os: windows-2025
+            extras: ""
+          - py: "3.13"
+            os: ubuntu-24.04-arm
+            extras: ".[stan]"
```
MichaelClerx: Can we restrict the stan stuff to one version?
It only runs on 3.10, 3.11, and 3.12, I think

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    # Skip job if it's a draft PR
+    if: github.event.pull_request.draft == false
+
+    strategy:
+      matrix:
+        include:
+          - py: "3.9"
+            os: ubuntu-22.04
+            extras: ""
```
MichaelClerx: Possible to have it default to "" instead of respecifying every time?

MichaelClerx:
Just out of curiousity: Would it be less verbose to run a full matrix but put some if() condition in that rejects certain (i,j) combos? Or is that even more typing?

.github/workflows/unit-tests.yml
fcooper8472: Yes, we can do this. Good thinking.

.github/workflows/unit-tests.yml
fcooper8472: Yes, I'll restrict the .[stan] builds.

fcooper8472:
> Just out of curiousity: Would it be less verbose to run a full matrix but put some if() condition in that rejects certain (i,j) combos? Or is that even more typing?

I think there end up being a large number to exclude.... I'll get a proof of concept working and then let's see if it can be streamlined at all.

MichaelClerx:
Pystan just fixed this. Should now run on .12, .13, and .14
https://github.com/stan-dev/pystan/pull/416/changes

